PR #40, opened 2016-02-25
Fix for 0.5

andreasnoack:
Use `Compat.LinAlg.checksquare`.


dmbates:
Thanks for this fix, Andreas.  Right now the dependency on https://github.com/dcjones/Showoff.jl is hanging things up for 0.5, as you know.

PR #41, opened 2016-03-01
Use StatsBase 0.8.0's AIC, BIC and df

nalimilan:
These functions now have generic definitions in StatsBase.
Without this, a conflict arises when running the tests, so we can't tag a new release of StatsBase.

I didn't touch `npar`, but since it's exactly the same function as `df`, we should probably keep only one of them. I find `npar` slightly misleading since people won't usually think of sigma as a parameter, but maybe that's just me.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ DataFrames 0.6
 Distributions 0.8
 NLopt 0.2
 Showoff
+StatsBase 0.8.0
--- a/src/MixedModels.jl
+++ b/src/MixedModels.jl
@@ -10,8 +10,6 @@ export LinearMixedModel,
        MixedModel,
        VarCorr,
 
-       AIC,        # Akaike's Information Criterion
-       BIC,        # Schwatz's Bayesian Information Criterion
        bootstrap,  # Create bootstrap replications of a model
        fixef,      # extract the fixed-effects parameter estimates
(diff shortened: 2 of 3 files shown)


nalimilan:
Thanks! Could you tag a new release and open a PR to register it, so that we can merge it at the same time as StatsBase?

dmbates:
@nalimilan Astonishingly, I am caught in another git mess that I don't know how to resolve.  (That's sarcasm - I don't think I have ever tagged and published a release without getting caught in a git mess.) I'm not even sure what the problem is as there does seem to be a valid release of the `Showoff` package.  There are problems with `Showoff` but only on Julia version `0.5.0-dev`, I believe.

I have tagged a release but now I get this
```julia
julia> Pkg.publish()
INFO: Validating METADATA
ERROR: packages with unsatisfiable requirements found:
    MixedModels v0.4.5 – no valid versions exist for package Showoff

 in error at error.jl:21
 in check_metadata at pkg/entry.jl:635
 in publish at pkg/entry.jl:372
 in anonymous at pkg/dir.jl:31
 in cd at file.jl:22
 in cd at pkg/dir.jl:31
 in publish at pkg.jl:61

```

dmbates:
I think I have it worked out now.  See https://github.com/JuliaLang/METADATA.jl/pull/4695

PR #46, opened 2016-05-10
Diagwts

dmbates:
Many fixes, especially to `glmm`

- use `Diagonal{T}` for weights.  In 0.5 the `scale!` method for vector/vector or vector/matrix is replaced by multiplication by a `Diagonal`
- `wt` and `offset` become named arguments
- considerable rewriting of calls to eliminate unnecessary memory allocations
- switch `unscaledre!` to use the random effects on the `B` scale (part cleaning up memory allocations)
- prepare for using threads in `updateμ!` methods (waiting on new release of `Compat` package)
- add datasets for testing
- add tests for glmm


nalimilan:
What do you think about renaming `wt` to `weights`? That would be more explicit, and more consistent with `offset`, which is spelled in full. We could make the same change to GLM.jl too.

PR #55, opened 2016-08-30
Binomial

dmbates:
Be more careful when detecting the Binomial distribution in log-likelihood calculation for GLMMs


dmbates:
Travis failure was actually a timeout on Homebrew initialization on OSX,

PR #57, opened 2016-10-12
Categorical

dmbates:
Too many changes in this commit but it is something to work with.


src/remat.jl
```diff
@@ -66,19 +66,25 @@ end
 
 Base.eltype(R::ReMat) = eltype(R.z)
 
-function Base.copy!{S,R}(d::PooledDataVector{S,R}, s::PooledDataVector{S,R})
-    copy!(d.pool, s.pool)
+function Base.copy!(d::CategoricalVector, s::CategoricalVector)
```
nalimilan: Please define this function in CategoricalArrays.jl instead.

dmbates:
Failures on AppVeyor are related to DataArrays on nightlies branch.

src/remat.jl
nalimilan: Now in 0.1.0 (https://github.com/JuliaLang/METADATA.jl/pull/6792).

dmbates: @nalimilan Thanks for the reminder.  I'll delete my version from the master branch of MixedModels.

It turns out that I may not actually call that method.  I need to do some more checking.

PR #58, opened 2016-10-13
Switch to using GLM.GlmResp, not vectors

dmbates:


dmbates:
Decided not to wait on Travis OSX tests.  They seem very slow today.

PR #75, opened 2017-03-28
Fix NLopt relative solution tolerance, add max feval NLopt criterion

kaskarn:
The relative X tolerance criterion was previously set to optsum.ftol_rel
(this may have been a typo). This change changes it to optsum.xtol_rel

In addition, this sets the NLopt maxeval stopping criterion to
optsum.feval ( -1 by default)


dmbates:
Thank you for catching these issues.

PR #76, opened 2017-03-31
replace model frame member with fixef coefnames

kleinschmidt:
The only thing the ModelFrame member of the LinearMixedModel type was used for was as an argument to `coefnames`.  I want to fix a mixed model without an actual model frame, so this PR replaces the ModelFrame member with a `Vector{String}` of fixed effect coef names which are then used directly.


dmbates:
Sorry for the delay.  This PR slipped off my radar as I am mostly working on another branch.

PR #78, opened 2017-04-13
Switch to the lower Cholesky representation

dmbates:


appveyor.yml
```diff
@@ -1,14 +1,13 @@
...
+  - JULIAVERSION: "julianightlies/bin/winnt/x86/julia-latest-win32.exe"
+  - JULIAVERSION: "julianightlies/bin/winnt/x64/julia-latest-win64.exe"
   - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
 
 branches:
   only:
-    - master
-    - /release-.*/
+    - LowerCholesky
```
tkelman: was this supposed to be temporary?

dmbates: Yes.  Thanks for catching that.

Can I make this change in release v0.8.0?  If so, can you outline how I would do it?

tkelman: there are instructions on the attobot readme, but too late now that avik merged it. since this is CI only it's not necessary to retag for this

PR #80, opened 2017-05-08
Fix small typo

pkofod:


dmbates:
Thanks for catching that.

PR #87, opened 2017-06-05
fix method to convert FactorReTerm matrix to sparse matrix

### LINKED ISSUE #86: mismatch in size for FactorReTerm
The `size` of a `FactorReTerm` doesn't agree with the `size` of its sparsified version:

Using the `corr` `FactorReTerm` from the tests:

```julia
julia> size(corr)
(180,36)

julia> size(sparse(corr))
(360,18)
```

Is that intended behavior? If so, I'm not sure I understand what the purpose of the `sparse(::FactorReTerm)` is; I'd assumed it was to recover the non-condensed form of the random effects matrix.

kleinschmidt:
This uses the correct row/column sparse matrix indices for vector ReTerms. (Fixes #86)


dmbates:
Thanks for fixing this.

PR #103, opened 2017-11-27
Convert to using DataFrames v0.11, CSV, CategoricalArrays and StatsModels

dmbates:


src/modelterms.jl
```diff
@@ -74,26 +74,26 @@ end
...
+* `f`: the grouping factor as a `CategoricalVector`
 * `z`: the raw random-effects model matrix as a vector.
 * `wtz`: a weighted copy of `z`
 * `fnm`: the name of the grouping factor as a `Symbol`
 * `cnm`: the column name as a string
 * `Λ`: the relative covariance multiplier
 """
 mutable struct ScalarFactorReTerm{T,V,R} <: AbstractFactorReTerm{T}
-    f::PooledDataVector{V,R}
+    f::CategoricalVector{V,R}
```
nalimilan: Note that this is an abstract type since there are more parameters. If making this struct concrete matters, the whole type of `f` should be made a type parameter.

dmbates: Thanks for the heads-up.

Let me describe why I want to know these parameters and perhaps you can tell me if I need a concrete type to achieve this.  Linear algebra operations regarding these "terms" usually involve indexing operations that depend on `f.refs`.  I found using `@code_warntype` that it was very important for the compiler to specialize on the element type of `f.refs`.  For `ScalarFactorReTerm{T,V,R}`, `V` is unimportant but `T` and `R` are very important.

With that in mind, should I extend the list of parameters in `CategoricalVector`?

nalimilan: I think so, unless you pass `r.refs` to a separate function, which is going to be specialized anyway. But note that you don't need to list all parameters, you can just do `f::V` with `V<:AbstractCategoricalVector` in the type signature, and drop `R`.

Issue #107, opened 2017-12-04
dof is not defined for MixedModels.GeneralizedLinearMixedModel{Float64}


andyreagan:
Looks like it's not implemented:

> 
> Stacktrace:
>  [1] dof(::MixedModels.GeneralizedLinearMixedModel{Float64}) at /home/jovyan/.julia/v0.6/StatsBase/src/statmodels.jl:73
>  [2] aic(::MixedModels.GeneralizedLinearMixedModel{Float64}) at /home/jovyan/.julia/v0.6/StatsBase/src/statmodels.jl:106
>  [3] include_string(::String, ::String) at ./loading.jl:522

The linear model code is here, I'm not sure how to modify for the generalized model though: https://github.com/dmbates/MixedModels.jl/blob/7d152ad8ac69338fd19c41f8fcf5a82a201dd0e8/src/pls.jl#L260

andyreagan:
I think it should be the same, right?

andyreagan:
Fixed by #110

PR #108, opened 2017-12-04
Fixes dmbates/MixedModels.jl#107

andyreagan:


dmbates:
Thanks for catching that omission.  Your solution is not quite right because `dof(m.LMM)` counts a scale parameter, which is not present for common families like `Bernoulli`, `Binomial` and `Poisson`.  I will have to go back to the `GLM` package to see how the counting is done there.

andyreagan:
Okay!

Another one to grab, which I expect is correct, is:

    StatsBase.nobs(m::GeneralizedLinearMixedModel) = nobs(m.LMM)

Issue #106, opened 2017-12-03
UndefVarError: coefnames not defined


IgorDouven:
I'd like to explore the Mixed Models package, but I keep getting the same error message. Here is an example:

using DataFrames, MixedModels

df = DataFrame(Var1 = rand(100), Var2 = rand(100), Id = repeat(1:10, inner=10))

mm1 = fit!(lmm(@formula(Var1 ~ Var2 + (1 + Var2 | Id)), df))

The error I get is:

WARNING: both StatsBase and DataFrames export "coefnames"; uses of it in module MixedModels must be qualified
ERROR: UndefVarError: coefnames not defined
Stacktrace:
 [1] #lmm#16(::Array{Any,1}, ::Dict{Any,Any}, ::Distributions.Normal{Float64}, ::Function, ::DataFrames.Formula, ::DataFrames.DataFrame) at /Users/igor/.julia/v0.6/MixedModels/src/pls.jl:112
 [2] lmm(::DataFrames.Formula, ::DataFrames.DataFrame) at /Users/igor/.julia/v0.6/MixedModels/src/pls.jl:83

Here is the versioninfo()

Julia Version 0.6.1
Commit 0d7248e (2017-10-24 22:15 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, haswell)

And Pkg.status("MixedModels") gives

MixedModels    0.13.0   master

IgorDouven:
I was able to fix the problem by replacing two occurrences of 'coefnames' in pls.jl by 'DataFrames.coefnames'.

andyreagan:
The issue is that StatsModels exported coefnames in 0.19.2

If the dependency for MixedModels is pinned at 0.19.0 (rather than 0.19) in the REQUIRE file this should fix this issue.
I can make this pull request but it might just be easier for @dmbates.

andyreagan:
In the meantime, @IgorDouven you can fix this by just pinning stasbase:
    Pkg.pin("StatsBase",v"0.19.0")

dmbates:
[...]

PR #109, opened 2017-12-04
Fixes dmbates/MixedModels.jl#106

andyreagan:


dmbates:
Again, thanks for the PR.  Your change doesn't really fix the problem because the original and the modified requirement are synonymous. The first version number is the lower bound and a lower bound of 0.19 is the same as a lower bound of 0.19.0.  To disallow StatsBase 0.19.2 and later you would need to put an upper bound of 0.19.1 but I would prefer not to do that.

I do not mean to discourage you from submitting pull requests.  I am just trying to explain why I will end up doing things differently.

andyreagan:
Yeah, that makes sense! I misunderstood here:
https://docs.julialang.org/en/stable/manual/packages/#Requirements-Specification-1

I think that the conflict between statsmodels and statsbase needs to be fixed between those two groups, rather than us pinning a version that works.
Maybe I'm confused, though.

dmbates:
@andyreagan Yes.  The solution is to release a new version of `StatsModels` that does not define `coefnames` and depends on `StatsBase 0.19.2`

andyreagan:
Yup, seeing your comment over on StatsBase's latest release as well!

Issue #93, opened 2017-08-07
Add predict() function


am800:
Dear Douglas Bates,

With my request to adding a predict(object, newdata) to your package, I am referring to the following post: https://stackoverflow.com/questions/45457286/predict-using-julia-lang-mixedmodels-jl

Adding a predict() like function would greatly add to the functionality of your package.

Thanks!

dmbates:
Fixed in v0.16.0 via #112

PR #111, opened 2017-12-07
Fixes dmbates/MixedModels.jl#93

andyreagan:


andyreagan:
I'm not sure this is terribly efficient - it iterates through and forms the full matrix for each random effect group variable.

dmbates:
@andyreagan Thanks for the PR.

I hadn't thought about this issue for some time.  The solution existed but under a different name.  It is called `fitted` rather than `predict` for a `LinearMixedModel`.

For a `GeneralizedLinearMixedModel`, `m`, the linear predictor is actually stored and updated as `m.η`.

Do you want to change the PR or shall I just insert these method definitions?

dmbates:
I should have been more explicit.  For a `GeneralizedLinearModel` just replace `StatsBase.predict(obj.LMM)` in your `predict` method with `obj.η`.

andyreagan:
Okay, so we don't need the linear model code? I think calling it `predict`
might be helpful.

Got it re: eta. I'll change that.

I am trying to make it work to make predictions on a new data frame
(testing data as opposed to training), but I the issue I have is when there
are more/less levels of a group, the construction of a new trms object with
the LMM constructor has different size.
I might have to extract that code its own function so it can accept fixed
levels.

I'll update when I have it working!
On Dec 7, 2017 4:31 PM, "Douglas Bates" <notifications@github.com> wrote:

I should have been more explicit. For a GeneralizedLinearModel just replace
StatsBase.predict(obj.LMM) in your predict method with obj.η.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/dmbates/MixedModels.jl/pull/111#issuecomment-350100333>,
or mute the thread

dmbates:
Is it okay with you if I just go ahead and make the changes to create `predict` methods without the second argument for the time being.  I have another set of changes that I would like to release today and I thought I might add the `predict` methods that correspond to `fitted` while doing that.

Trying to write a `predict` method for an existing model and new data gets complicated because you need to check the covariate names and the levels of factors.  With fixed-effects terms the levels should be fixed, as the name indicates.  With random effects the levels might change when you get a new data set so you need to be a bit more explicit about how you want to interpret the random effects in the prediction.

andyreagan:
Yeah go ahead!

The levels issue is the one I am working on now.

PR #113, opened 2017-12-12
Expanding levels respecting DF categorical, predicting with new DF

andyreagan:
new DF can have a subset of the levels used to train the model, so long as the DF has the levels set appropriately.

### DIFF
--- a/src/modelterms.jl
+++ b/src/modelterms.jl
@@ -198,14 +198,14 @@ end
 function Base.sparse(A::VectorFactorReTerm{T,V,R,S}) where {T,V,R,S}
     n = size(A, 1)
     colind = Matrix{Int32}(S, n)
-    ord_lookup = A.f.pool.invindex
+    rr = A.f.refs
     @inbounds for j in 1:n
-        offset = (ord_lookup[A.f[j]] - 1) * S
+        offset = (rr[j] - 1) * S
         for i in 1:S
             colind[i, j] = offset + i
         end
     end
-    sparse(Vector{Int32}(repeat(1:n, inner=S)), vec(colind), vec(A.z), n, length(ord_lookup)*S)
+    sparse(Vector{Int32}(repeat(1:n, inner=S)), vec(colind), vec(A.z))
 end
 
 """
--- a/src/modelterms.jl
+++ b/src/modelterms.jl
@@ -198,15 +198,14 @@ end
 function Base.sparse(A::VectorFactorReTerm{T,V,R,S}) where {T,V,R,S}
     n = size(A, 1)
     colind = Matrix{Int32}(S, n)
-    lvls = levels(A.f)
-    ord = CategoricalArrays.order(A.f.pool)
+    ord_lookup = A.f.pool.invindex
     @inbounds for j in 1:n
-        offset = (ord[A.f[j]] - 1) * S
+        offset = (ord_lookup[A.f[j]] - 1) * S
         for i in 1:S
             colind[i, j] = offset + i
         end
     end
-    sparse(Vector{Int32}(repeat(1:n, inner=S)), vec(colind), vec(A.z), n, length(lvls)*S)
+    sparse(Vector{Int32}(repeat(1:n, inner=S)), vec(colind), vec(A.z), n, length(ord_lookup)*S)
 end
 
 """
--- a/src/modelterms.jl
+++ b/src/modelterms.jl
@@ -199,13 +199,14 @@ function Base.sparse(A::VectorFactorReTerm{T,V,R,S}) where {T,V,R,S}
     n = size(A, 1)
     colind = Matrix{Int32}(S, n)
     lvls = levels(A.f)
+    ord = CategoricalArrays.order(A.f.pool)
     @inbounds for j in 1:n
-        offset = (findfirst(lvls,A.f[j]) - 1) * S
+        offset = (ord[A.f[j]] - 1) * S
         for i in 1:S
             colind[i, j] = offset + i
         end
     end
-    sparse(Vector{Int32}(repeat(1:n, inner=S)), vec(colind), vec(A.z))
+    sparse(Vector{Int32}(repeat(1:n, inner=S)), vec(colind), vec(A.z), n, length(lvls)*S)
 end
 
 """
--- a/src/modelterms.jl
+++ b/src/modelterms.jl
@@ -198,9 +198,9 @@ end
 function Base.sparse(A::VectorFactorReTerm{T,V,R,S}) where {T,V,R,S}
     n = size(A, 1)
     colind = Matrix{Int32}(S, n)
-    rr = A.f.refs
+    lvls = levels(A.f)
     @inbounds for j in 1:n
-        offset = (rr[j] - 1) * S
+        offset = (findfirst(lvls,A.f[j]) - 1) * S
         for i in 1:S
             colind[i, j] = offset + i
         end
--- a/src/pls.jl
+++ b/src/pls.jl
@@ -58,8 +58,7 @@ function LinearMixedModel(f, trms, wts)
     LinearMixedModel(f, trms, sqrtwts, A, LowerTriangular(L), optsum)
 end
 
-model_response(mf::ModelFrame, d::Distribution=Normal()) =
-    model_response(mf.df[mf.terms.eterms[1]], d)
+model_response(mf::ModelFrame, d::Distribution=Normal()) = model_response(mf.df[mf.terms.eterms[1]], d)
 
 model_response(v::AbstractVector, d::Distribution) = Vector{partype(d)}(v)
 
@@ -216,7 +215,7 @@ function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}, trms::AbstractTerm
     v
 end
 
-function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T = fitted!(v, m, m.trms)
+fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where {T} = fitted!(v, m, m.trms)
 
 StatsBase.fitted(m::LinearMixedModel{T}) where {T} = fitted!(Vector{T}(nobs(m)), m)
 
--- a/src/pls.jl
+++ b/src/pls.jl
@@ -80,7 +80,7 @@ function get_tdict(terms)
     tdict
 end
 
-function get_trms(tdict::Dict,df,fr::AbstractDataFrame,T::Type,n::Int)
+function get_trms(tdict::Dict, df, fr::AbstractDataFrame, T::Type, n::Int)
     trms = AbstractTerm{T}[]
     for (grp, lhs) in tdict
         gr = compress(categorical(getindex(df, grp)))
@@ -128,7 +128,7 @@ function lmm(f::Formula, fr::AbstractDataFrame;
     T = eltype(X)
     y = model_response(mf, rdist)
     tdict = get_tdict(mf.terms.terms)
-    trms = get_trms(tdict,mf.df,fr,T,n)
+    trms = get_trms(tdict, mf.df, fr, T, n)
     sort!(trms, by = nrandomeff, rev = true)
     push!(trms, MatrixTerm(X, coefnames(mf)))
     push!(trms, MatrixTerm(y))
@@ -246,11 +246,15 @@ function StatsBase.predict(m::LinearMixedModel{T}, test::AbstractDataFrame; #, t
     trms = get_trms(tdict, mf.df, test, T,n)
 
     # now, sort them to have the same order as in the training data
-    sort!(trms, by = nrandomeff, rev = true)
+    # they were sorted this way in the training:
+    # sort!(trms, by = nrandomeff, rev = true)
+    # now we'll just use that as a lookup
+    fnms = [m.trms[i].fnm for i=1:(length(m.trms)-2)]
+    sort_idx = [findfirst(fnms,x) for x=[y.fnm for y in trms]]
+    trms = trms[sort_idx]
+    # need the fixed effects data:
     push!(trms, MatrixTerm(X, coefnames(mf)))
-    # for the indexing on fitted
-    # push!(trms, AbstractArray{T})
-    # push!(trms, zeros(T, 0))
+    # for the indexing on fitted:
     push!(trms, MatrixTerm(zeros(T, 0)))
     
     fitted!(Vector{T}(size(test, 1)), m, trms)
--- a/src/pls.jl
+++ b/src/pls.jl
@@ -206,9 +206,8 @@ function StatsBase.fit!(m::LinearMixedModel{T}, verbose::Bool=false) where T
     m
 end
 
-function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T
+function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}, trms::AbstractTerm{T}) where T
     ## FIXME: Create and use `effects(m) -> β, b` w/o calculating β twice
-    trms = m.trms
     A_mul_B!(vec(v), trms[end - 1], fixef(m))
     b = ranef(m)
     for j in eachindex(b)
@@ -217,19 +216,18 @@ function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T
     v
 end
 
+function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T = fitted!(v, m, m.trms)
+
 StatsBase.fitted(m::LinearMixedModel{T}) where {T} = fitted!(Vector{T}(nobs(m)), m)
 
 StatsBase.predict(m::LinearMixedModel) = fitted(m)
 
-function StatsBase.predict(obj::LinearMixedModel,newDF)
+function StatsBase.predict(m::LinearMixedModel{T}, newDF::AbstractDataFrame) where T
     # just a hack here, we'll generate a new model to have it do the dataframe manipulation
-    new_obj = lmm(obj.formula,newDF)
-    # println(full(new_obj.trms[1]))
-    ranef_total = 0
-    for grp=1:length(obj.trms)-2
-        ranef_total += full(new_obj.trms[grp])*collect(Base.Iterators.flatten(ranef(obj)[grp]))
-    end
-    new_obj.trms[end-1].x*fixef(obj) + ranef_total
+    # note, you do need at least 2 of the levels from the training data
+    # could be done in one line...
+    new_m = lmm(m.formula,newDF)
+    fitted!(Vector{T}(size(newDF,1)), m, new_m.trms)
 end
 
 StatsBase.residuals(m::LinearMixedModel) = model_response(m) .- fitted(m)
--- a/src/pls.jl
+++ b/src/pls.jl
@@ -69,6 +69,49 @@ function model_response(v::CategoricalVector, d::Bernoulli)
     nlevs < 2 ? zeros(v, partype(d)) : partype(d)[cv == levs[2] for cv in v]
 end
 
+function get_tdict(terms)
+    tdict = Dict{Symbol, Vector{Any}}()
+    for t in filter(x -> Meta.isexpr(x, :call) && x.args[1] == :|, terms)
+        fnm = t.args[3]
+        isa(fnm, Symbol) || throw(ArgumentError("rhs of $t must be a symbol"))
+        tdict[fnm] = haskey(tdict, fnm) ? push!(tdict[fnm], t.args[2]) : [t.args[2]]
+    end
+    isempty(tdict) && throw(ArgumentError("No random-effects terms found in $f"))
+    tdict
+end
+
+function get_trms(tdict::Dict,df,fr::AbstractDataFrame,T::Type,n::Int)
+    trms = AbstractTerm{T}[]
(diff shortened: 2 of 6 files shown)


andyreagan:
I found an issue, this only works with levels that are a subset, but contain the full range of the training levels. Working on a fix to that.

src/pls.jl
```diff
@@ -221,6 +221,17 @@ StatsBase.fitted(m::LinearMixedModel{T}) where {T} = fitted!(Vector{T}(nobs(m)),
...
 StatsBase.predict(m::LinearMixedModel) = fitted(m)
 
+function StatsBase.predict(obj::LinearMixedModel,newDF)
+    # just a hack here, we'll generate a new model to have it do the dataframe manipulation
+    new_obj = lmm(obj.formula,newDF)
+    # println(full(new_obj.trms[1]))
+    ranef_total = 0
+    for grp=1:length(obj.trms)-2
+        ranef_total += full(new_obj.trms[grp])*collect(Base.Iterators.flatten(ranef(obj)[grp]))
+    end
```
dmbates: You really don't want to do `full(new_obj.trms[grp])`, which could easily exhaust the available memory for some models with tens of millions of observations and hundreds of thousands of levels in a grouping factor.  Look at the existing `fitted` function and the use of `unscaledre!` for an alternative.

I am okay with writing a version of this if you prefer.  I make these suggestions assuming that you are interested in learning the low-level details.

The point of all of the use of sparse and blocked matrices and the various types of AbstractTerm is to avoid creating the full matrix representation.

andyreagan:
Yeah that makes sense, I'll update so that it's not building the full
matrices.

--
Andy Reagan
Department of Mathematics and Statistics
The University of Vermont
16 Colchester Ave.
Burlington, VT 05405
andy@andyreagan.com

src/modelterms.jl
```diff
@@ -198,9 +198,9 @@ end
 function Base.sparse(A::VectorFactorReTerm{T,V,R,S}) where {T,V,R,S}
     n = size(A, 1)
     colind = Matrix{Int32}(S, n)
-    rr = A.f.refs
+    lvls = levels(A.f)
     @inbounds for j in 1:n
-        offset = (rr[j] - 1) * S
+        offset = (findfirst(lvls,A.f[j]) - 1) * S
```
nalimilan: I don't understand that code, but if `n` is large it would be more efficient to do something like `ord = CategoricalArrays.order(A.f.pool); ord[A.f[j]]`.

andyreagan:
Does that accomplish a faster lookup than findfirst(array,value)? I'm just
using the levels array to look up the index of that level.

--
Andy Reagan
Department of Mathematics and Statistics
The University of Vermont
16 Colchester Ave.
Burlington, VT 05405
andy@andyreagan.com

src/modelterms.jl
andyreagan: Does that accomplish a faster lookup than findfirst(array,value)? I'm just
using the levels array to look up the index of that level.

nalimilan: Indexing an array is (much) faster than a linear search, especially if levels are e.g. strings. Whether the difference matters or not depends on `n`.

andyreagan: Okay I just didn't know that is what `CategoricalArrays.order` builds.

`n` here can be in the thousands, and this is called once for each row in the data. So, yes, worthwhile. Thanks!

dmbates:
@andyreagan It is still not clear to me why you want to build even the sparse matrix.  The unscaledre! function updates the response vector in a loop, thereby avoiding the creation and storage of potentially large objects that are only used once.  One of the wonderful aspects of programming in Julia as opposed to R, Matlab/Octave, Python, etc., is that you do not need to "fear the loop".

andyreagan:
This is the sparse matrix that I was referring to:

https://github.com/dmbates/MixedModels.jl/blob/9d70f8f516599a6174638d5bd4a009716c77167e/src/modelterms.jl#L208

Which needs to be built to make predictions, no?
Trying to reuse as much as possible.

The latest commit should be acceptable.

andyreagan:
There are still a few tests that didn't pass, I'll look into these more.

src/modelterms.jl
```diff
@@ -198,14 +198,14 @@ end
 function Base.sparse(A::VectorFactorReTerm{T,V,R,S}) where {T,V,R,S}
     n = size(A, 1)
     colind = Matrix{Int32}(S, n)
-    rr = A.f.refs
+    ord_lookup = A.f.pool.invindex
     @inbounds for j in 1:n
-        offset = (rr[j] - 1) * S
+        offset = (ord_lookup[A.f[j]] - 1) * S
```
nalimilan: I don't understand why you do this. It doesn't make a lot of sense to extract the value and then lookup its reference code inside the pool. If I'm not mistaken that's strictly equivalent to `A.f.refs[j]`. Doesn't the solution based on `order` work? Or maybe you need the inverse of `order`?

andyreagan: Good question. The solution based on order does not work, it broke the tests because it was looking up the value. Have a look at the object created by the pool method of a `CategoricalArray`, and what `order()` on that gives you back (which is a list of the refs). We needed a dictionary to look that up from the values of `A.f`.

From what I can tell, the difference between this lookup using the `ref` is that this respects the full set of levels (whereas refs will only use the levels seen in the data). By going this new way, you allow a user to specify levels on the data programmatically.

nalimilan: No, I don't think that's the problem here. You're iterating over the entries in `A.f` in this loop, so only values present in the data are considered. And `order` returns the ordering of all `levels`, not only of those present in the data. Can you explain exactly what the problem with the `order`-based approach?

dmbates:
@andyreagan Regarding the sparse matrix, you don't need to create it and there is a non-trivial overhead in doing so.  Multiplication of a `ScalarFactorReTerm` or a `VectorFactorReTerm` by an array of random effects can be performed in place by `unscaledre!`.  Such a term can be represented as a sparse matrix but it is a special type of sparse matrix in that each row has the same number of non-zeros.  In the case of a `ScalarFactorReTerm` it is even simpler because each row has exactly one nonzero.  What you would need to do is to incorporate the lookup of the index of the level of the factor in the new data frame into the indexing of `b` in the `unscaledre!` methods.

Also, there is no need to look up each value of the new factor in the index table of the old factor.  Just map the levels of the new one to the levels of the old one, as @nalimilan suggests.

Admittedly the name `unscaledre!` is not the most intuitive for this operation.  I wrote it as part of the bootstrap simulation code then realized that it could be used in creating the fitted values.  Essentially the simulation process is just creating fitted values for simulated random effects, then adding the per-observation noise.

src/modelterms.jl
andyreagan: What exactly do you mean by the `order`-based approach? I tried the code you posted before, and it didn't work, so I fixed it with this.

andyreagan: This is the code you posted before, which is what I assume you mean

    ord = CategoricalArrays.order(A.f.pool); ord[A.f[j]]

But that simply doesn't work.

andyreagan:
I'm not calling `sparse` directly in any of my code, just wrapping the `fitted!` method for the `LinearMixedModel`, which looks like it uses that `unscaledre!` function already (https://github.com/dmbates/MixedModels.jl/pull/113/files#diff-a74475b07cb5d49fc3222742b04ae9e7L215)

src/modelterms.jl
nalimilan: The exact equivalent of what you're currently doing is just `A.ref[j]`. So if that's really what you need, just use that, but that's also what the old code did.

The problem is, I don't think that's what's is intended here. The approach described above returns levels in their internal storage order, not in the user-visible and user-defined order defined by `levels`. `order` allows passing from the internal codes to the user-visible codes, if that's what is intended.

nalimilan: Oh, I just noted a possible explanation for the failure of the example you give in your last comment. It should be:
```julia
ord = CategoricalArrays.order(A.f.pool); ord[A.f.refs[j]]
```

andyreagan: Oh yeah that fixes yours, and makes more sense.

dmbates:
If you are not using `sparse` why do you change it?  Neither @nalimilan nor I understand the reason for the changes.  Could you explain, please?

dmbates:
Were you planning on adding tests of the new `predict` methods?

andyreagan:
I just arrived at that conclusion on my own, I didn't need to change the sparse code.

Yeah I can test them, using the sleep study data?

dmbates:
Sleep study data sounds fine.  It can exercise the `VectorFactorReTerm` case.  You may want to insert the example in the appropriate stanza of `test/pls.jl`.

andyreagan:
Yup, on it.

andyreagan:
@nalimilan I'm having trouble making two categorical arrays have the same index. I can set the levels to be the same with:

    levels!(a,levels(b)

but that doesn't make the indexes the same, e.g., `a.pool.index == b.pool.index`.
I also tried making sure that they are both ordered.

andyreagan:
I wrote a test but I'm not sure why it's not working. Hmm.

nalimilan:
There's currently no way to force two arrays to have the same indexes, except by filling them from scratch in the same order. But why do you need this? It should be an implementation detail without any consequence as long as you use `order`.

andyreagan:
Okay that's maybe not the important issue.

The problem I have now is that the new `predict(model, newDF)` method works when `newDF` is the same as the original dataframe for the model (the predictions are all the same), but when you subset the original dataframe into, say, the first 1000 rows, the predictions on those 1000 rows are not the same as the first 1000 rows of the predictions for the full dataset.

In other words, this line works:
https://github.com/dmbates/MixedModels.jl/pull/113/commits/a6a0a99e2c5c1b31a638fcea022c98c727d6f8c3#diff-bbd0226dcad28723d2ed194703ab6447R134
But this one does not:
https://github.com/dmbates/MixedModels.jl/pull/113/commits/a6a0a99e2c5c1b31a638fcea022c98c727d6f8c3#diff-bbd0226dcad28723d2ed194703ab6447R145

nalimilan:
That's just a guess, but maybe the code working with `CategoricalArray` needs to use `order`? Currently it works directly with the `refs` field, which is suspicious since that's supposed to be an implementation detail. At least, IIRC StatsModels uses the ordering of `levels`/`order`, not that of `refs`, to build the model matrix.

andyreagan:
@nalimilan so it would be this code: 

https://github.com/dmbates/MixedModels.jl/blob/6d99ac613fd84364fcd8f79a47fd611413e0ffa9/src/simulate.jl#L186

which is ultimately called by `fitted`, that could be giving us the weird results?

nalimilan:
Yes, possibly. I really don't know, but it would be worth trying to change all places where `refs` is used to use `order` as I described above.

andyreagan:
And I also take it that you're meaning that the `ref` of a CategoricalArray aren't meant to be used as part of a public api, but just a backend detail.

I guess the question then is how to use `order`? I tried, but you said my code was ultimately the same as using the `ref`.

nalimilan:
> I guess the question then is how to use order? I tried (in the wrong place), but you said my code was ultimately the same as using the ref.

No, that's not what I said at all. It's going through `invindex` which is equivalent.

andyreagan:
Okay, so I just don't know how to use `order` properly, could you explain? I have read the documentation and looked at the source code on `CategoricalArrays.jl`

nalimilan:
I gave an example in one of my recent comments.

andyreagan:
I looked back and added that to my previous. Can you point me to any documentation of what is returned by `CategoricalArrays.order`?

nalimilan:
It just returns a vector with the order in which each code in `refs` should be sorted, i.e. what is its order of appearance in `levels`.

dmbates:
I think the problem here is that you are not mapping the levels of the categorical vectors in the data frame for prediction to those in the data frame used to fit the model.  The random effects coefficients, and also the fixed-effects coefficients that refer to contrasts from categorical vectors, are indexed by those levels.  If you bring in a new data frame then you need to take the levels of any categorical vectors in the new frame and express them relative to the old levels.  That is you need to create `refs` for the new vector relative to the old levels.

You can't just take coefficients from a model fit to one set of categorical vectors and apply them by position to a new set of categorical vectors.  You have to map through the levels of the categorical vectors.

@nalimilan Can this be achieved with `levels!`?  In other words, can the `newlevels` argument be longer than the original levels of the CategoricalArray?

nalimilan:
> @nalimilan Can this be achieved with levels!? In other words, can the newlevels argument be longer than the original levels of the CategoricalArray?

Yes, you can certainly add levels which are not present in the data. But note that, as I stressed above, you shouldn't work with `refs` directly, as they are not affected at all by calls to `levels`, you need to go through `order` too.

dmbates:
@nalimilan Thanks for the warning about `order`.  I don't think I will change the fitting code at this point because fitting is always done with respect to a fixed set of levels and a fixed order.  It is only when introducing new categorical vectors that there is a need to make sure that the values of the new vector map correctly to the values of the old vector.

nalimilan:
I think it makes sense to use `order` all the time, because that follows the order defined by the user. Without it, the order of levels can be completely unexpected for the user. It shouldn't require a lot of changes I think.

andyreagan:
So @nalimilan you're saying we can drop in the following instead of using `a.refs[j]`:

    ord = CategoricalArrays.order(a.pool)
    for j=...
        # a.refs[j]
        ord[a[j]]
    end

@dmbates the problem is not that the levels aren't set or aren't the same. We have a problem when when just subsetting the dataframe used to train. I'll update the tests to show that the levels are the same while it still breaks.

nalimilan:
Yes. I don't know whether it will fix the problem, but it sounds better in general, unless you really don't care about the order of levels.

dmbates:
If `CategoricalArrays.order` and the double indexing is to be used in one place then I suspect that every use of the `refs` will need to be changed, and I want to have a lot better understanding of what goes on inside of `CategoricalArrays` before making such a change.  I haven't had time to read through the sources or all of the comments by @nalimilan in this thread and I don't see any mention of this in the `CategoricalArrays` documentation.

@nalimilan Where should I start reading about this?

nalimilan:
Unfortunately that's not really documented since that was supposed to be internal. We should improve the docs regarding this.

But there's no much to say: `CategoricalArrays.order(a.pool)[a.refs[i]]` will give you the position of the entry `i` in `levels`. The `refs` only give you the position in the `index`, which does not reflect the user-defined ordering.

dmbates:
Okay, thanks @nalimilan 

It looks like this is an issue that should be addressed because I am making assumptions about `refs` and `levels` that can be invalid.  One case I can think of is the labels for the results of `ranef` when `named = true`.

The general issue of a `predict` method incorporating new values is not high on my priority list because it is one of those issues in mixed-effects models that people think should be easy but in fact is extremely difficult to get right.  You have fixed-effects coefficients and random effects, which are also coefficients, so it is easy to think of them as the same sort of thing but they aren't.  Fixed-effects coefficients are parameters whereas random effects are unobserved random variables.  When you try to reason about them in terms of the stochastic and deterministic parts of the model it gets really tricky, unless you approach it from a Bayesian perspective where everything is a random variable.  But you wouldn't use this software for a Bayesian analysis anyway.

andyreagan:
Right- the "random effects" for each group are the modes of the posterior for each random variable (as I've read in your documentation).
Is there an estimate of the variance of the random effect variables that is created? I.e., if they are normal, then the mean/mode and variance should specify them. I know that the theta are what is actually fit, but I am not sure how the theta inform the random effects for each group. Maybe that is derived and I missed it, but that's one point I'm still confused on.

In a bayesian sense you would also get the variance for the fixed effects parameters.

And the tests are now passing, for the simple methods that I wrote.

dmbates:
@andyreagan For a linear mixed-effects model the values returned by `ranef` are the conditional means of the random effects, given the observed data (and fixing the values of the parameters at their MLE's).  `condVar` returns the conditional variances of the random effects but, at present, only works for the simplest of cases with a single `ScalarFactorReTerm`.

I'll take a look at the tests but I strongly suspect that the code will give the wrong answers in the case of a grouping factor in the new data frame not having the levels in the same order as in the old data frame, even if the `CategoricalVector` in question is not ordered.

src/pls.jl
```diff
@@ -217,10 +225,35 @@ function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T
...
 StatsBase.fitted(m::LinearMixedModel{T}) where {T} = fitted!(Vector{T}(nobs(m)), m)
 
 StatsBase.predict(m::LinearMixedModel) = fitted(m)
 
+function StatsBase.predict(m::LinearMixedModel{T}, test::AbstractDataFrame; #, train::AbstractDataFrame;
+	 contrasts=Dict()) where T
+    mf = ModelFrame(m.formula, test, contrasts=contrasts)
+    X = ModelMatrix(mf).m
+    n = size(X, 1)
+    T = eltype(X)
```
dmbates: `T` is already defined in the function signature.  Do you really want to override that here?

src/pls.jl
```diff
@@ -217,10 +225,35 @@ function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T
...
+
+    # # can the tdict be used from the original data? (without building this modelframe again)
+    # mf_train = ModelFrame(m.formula, train, contrasts=contrasts)    
+    # tdict = get_tdict(mf_train.terms.terms)
+    tdict = get_tdict(mf.terms.terms)
+
+    # make a new trms object
+    trms = get_trms(tdict,mf.df,test,T,n)
+    
+    sort!(trms, by = nrandomeff, rev = true)
```
dmbates: I don't think you want to sort the `trms` according to `nrandomeff` in the new data frame.

You probably want to align the random effects terms according to the grouping factor which is stored as the `fnm` member of the struct.

src/pls.jl
```diff
@@ -217,10 +225,35 @@ function fitted!(v::AbstractArray{T}, m::LinearMixedModel{T}) where T
...
+    push!(trms, MatrixTerm(X, coefnames(mf)))
+    # for the indexing on fitted
+    push!(trms, MatrixTerm(X, coefnames(mf)))
```
dmbates: Why `X` twice?  I realize you don't necessarily have a `y` column in the data frame but you just need a column of zeros of type `T`.  Even if you do decide to push `X` onto the vector twice you can define and evaluate it once then push it twice.

andyreagan: Yeah I had put `[]` but it wasn't happy with that. I think I'd just need `Array{T}`.

src/pls.jl
andyreagan: Aha! This could be the source of one of the issues I am experiences. I'll take a look at this.

src/pls.jl
andyreagan: Oh, no.

src/pls.jl
andyreagan: Okay I've answered my part 1 question: yes it uses the data.

andyreagan: I'm not sure why they need to be sorted in the first place (I suspect there is a reason, although it's also fine if you don't want to explain it to me).

nalimilan: Maybe they were sorted explicitly because the old code used the `refs` field of `CategoricalArray`, which isn't sorted? If so, using `order` would make it unnecessary to sort explicitly (which breaks any user-defined order).

andyreagan: Hm, I don't think so. This is sorting each different random effect, not for within an effect.

I fixed the code that I wrote to respect the original sort, anyway.

dmbates:
@andyreagan I appreciate your energy and enthusiasm but I don't plan on incorporating this PR.  The process of creating this PR has been too undisciplined.  Instead of deciding what `predict` should return, including careful consideration of how to match levels of categorical variables in the original DataFrame to which the model was fit and in the DataFrame passed as an argument, the history of this PR is a sequence of quick rewrites every time an issue comes up.  This "whack-a-mole" approach leads to software that is difficult to maintain.  Software engineering works best with careful consideration of the outcomes and a disciplined approach to building and testing intermediate steps.  Even now, after what I find to be substantial changes in the underlying code, there are only a couple of perfunctory tests of the method and only using the same dataframe as was used to fit the model.  All the difficult cases are not being tested.

If the only cases being tested are the only important cases then this PR is not needed.  The one-argument form of the `predict` method is, in effect, the only one being tested and that already exists.

When I find that the order of file inclusions in `src/MixedModels.jl` and the `lmm` function itself are being changed for reasons I don't understand and that do not appear to be documented, I can't accept such changes.  I ask you to understand that I have spend 5 years working on this package and over 20 years working on the software packages to fit such models.  I need to be able to understand, and agree with, decisions that are made on the basic structure of the package and the computational methods it incorporates.

andyreagan:
@dmbates thank you for the recognition of enthusiasm, and of course I don't take it personally that you want large changes to be done in a principled fashion.

Just as a note, `src/MixedModels.jl` didn't change, and I'd be happy to explain why I pulled out parts of `lmm` and put them into functions (simply so that code be re-used elsewhere without being copied).

andyreagan:
You mentioned that having this type of functionality is not of particular interest (or, at least, is more nuanced than one may initially think), so I'll just ask to not waste time for any of us: would you like me to more fully document the goals of the PR, and clean up the commit history (i.e., start a new PR from scratch), such that it is acceptable to integrate?

PR #138, opened 2018-09-24
Getproperty

dmbates:
- Use `getproperty` to query characteristics of the model and `setproperty` to install a new value of 'theta'.


nalimilan:
I think it's also useful to define `propertynames` to enable autocompletion.

dmbates:
@nalimilan Thanks for the reminder.  I will do that.

PR #146, opened 2018-11-16
use MatrixTerm/is_matrix_term to separate fixed and ranef terms

kleinschmidt:
I've moved all the logic for separating "matrix terms" (here fixed effects) and "non-matrix terms" (like random effects) into statsmodels in the latest commits.  Those also introduced a `MatrixTerm` that explicitly marks its children terms for concatenation at the `model_cols` stage (and guarantees that it produces a matrix) which also simplifies the code here somewhat


dmbates:
Thanks, Dave.

Issue #145, opened 2018-11-11
[Feature Request] Support for rank deficient


Nosferican:
Allow for rank deficient models to be handled and estimated.

Nosferican:
I can try working the PR for this. How would it best handle it from a design approach in the package?

PR #150, opened 2018-12-26
Fix rank-deficient models

Nosferican:
Issue #145. Should be a bug fix from updating code to 0.7/1.0 and part of a patch release.


dmbates:
Thanks.

nalimilan:
Could you add a test?

Nosferican:
I thought there was a test covering it, but after verifying it was missing. Good call. Here is #154.

PR #151, opened 2018-12-26
Update metadata

Nosferican:
License: Update year
README:
- drop PkgEval since it has been deprecated for 0.7/1.0
- Drop reference for `lmm` and `glmm` as documentation favors using the `fit` with the type
- Drop the installation quickstart since package is registered


dmbates:
Thanks again.

PR #164, opened 2019-02-19
Allow kwargs through fit

Nosferican:
Allows same arguments constructors take to be passed through `fit`.


dmbates:
Sorry, I didn't see #164 and #165 until today.

PR #165, opened 2019-02-19
Fixes bug on dof_residual for Linear

Nosferican:
Fixes bug on `dof_residual` for `LinearMixedModel`

### DIFF
--- a/src/pls.jl
+++ b/src/pls.jl
@@ -277,7 +277,7 @@ StatsBase.dof(m::LinearMixedModel) = size(m)[2] + sum(nθ, m.trms) + 1
 
 function StatsBase.dof_residual(m::LinearMixedModel)
     (n, p, q, k) = size(m)
-    n - REML * p
+    n - m.optsum.REML * p
 end
 
 StatsBase.loglikelihood(m::LinearMixedModel) = -objective(m)/2
--- a/test/pls.jl
+++ b/test/pls.jl
@@ -64,6 +64,7 @@ end
 
     fit!(fm1, REML=true)
     @test isapprox(objective(fm1), 319.65427684225216, atol=0.0001)
+    @test dof_residual(fm1) ≥ 0
     print(IOBuffer(), fm1)
 end
 


Nosferican:
Actually, I am not sure that formula is correct.

dmbates:
It is not clear what "degrees of freedom for residuals" means for a mixed-effects model.  It is well-defined as the dimension of the orthogonal to the column-span of the model matrix in the case of a linear model with fixed-effects only.  However, once you have random-effects in a model, degrees of freedom as dimensions no longer have meaning because the hat matrix is not a projection.

Nosferican:
I guess since the most common use would be for inference on the fixed effects estimates, maybe define it in terms of that if possible? That would allow to define `confint` and such, no?

dmbates:
It is much more complicated than you are making it out to be.  There could be arguments made for any number between `n - p` and `n - (p + q)` and what are you going to do with such a number even if it could be calculated?  You want to use the Kenward-Roger correction instead, don't you?

These issues all come down to thinking that the desire to generalize everything about linear models to linear mixed models should make it possible to do so.  It doesn't.  Isn't it just easier to accept that not everything related to linear models is immediately applicable to linear mixed models?

Nosferican:
I believe the KR correction only applies to REML models, no? I think for MLE models, sometimes these models will be re-fitted by REML to get the estimates. At least for the application I am working with, the standard/regulatory requirement is to obtain the confidence interval of one of the fixed effect estimates. I believe for join-significance the current standard is an ANOVA type III, but I believe an LR test should work if not mistaken (assuming a sensible degree of freedom, maybe a Wald test too?)

Maybe have the KR estimate be the default for `dof_residual`, assuming it is valid for inference for MLE models?

dmbates:
We seem to have skipped over the part about there is no code in this package for the KR "correction" and I don't know of any sensible way of evaluating it.  The only formulas that I have seen are of the sort created by people who don't actually do any computing so the formulas involve the inverse of matrices of size n by n.  That not only becomes infeasible as n gets large, it also becomes irrelevant.  If you look at, say, what happens to critical values of a t-distribution as the number of degrees of freedom becomes large, it is best to interpolate in the inverse of the number of degrees of freedom.  Many, many years ago when we used statistical tables for evaluating these things the typical tables would use 1:30, 40, 60, 120, Infinity for the degrees of freedom because 30, 40, 60, 120, Infinity are evenly spaced on the inverse scale.  The difference between 9286.23486 and 10000 df is negligible.

It was 20 years ago that Saikat DebRoy and I showed that you can use a generalized least squares or a penalized least squares approach to evaluating fixed-effects estimates and that penalized least squares is much more efficient.  Among other things using PLS allows you to work with the profiled log-likelihood and makes it feasible to fit models that have crossed grouping factors.

If you feel that it is urgent to have KR degrees of freedom then I would appreciate your giving me explicit instructions (or create a PR) on how to evaluate them using the information about the model provided by this package.

Nosferican:
I guess the SAS software [documentation](http://documentation.sas.com/?docsetId=statug&docsetTarget=statug_glimmix_details36.htm&docsetVersion=15.1&locale=en) offers: Between-Within, Containment, and Satterthwaite in addition to Kenward-Roger. Stata's [documentation](https://www.stata.com/manuals/memixed.pdf#memixed) provides repeated, anova, satterthwaite, and kroger. I suspect repeated/anova and between-within/containment seem to be the same, but differ in the name by software. Currently, requesting the `dof_residual` just errors with `REML is not found`. I defer to you if in the meantime it is worth to implement any of the first two which seem to be appropriate for simple mixed effects models or more complex with balanced data. After having those, then we can add Satterthwaite or Kenward-Roger later on. If you think, it is best to wait for the more robust class of estimators, we could remove the `dof_residual` to default to the standard error `dof_residual is not defined for T`.

PR #177, opened 2019-06-14
Use Project.toml and some fixes

andreasnoack:
It looks like a few things from GLM aren't exported anymore.

I also made minimal changes to a few tolerances.


dmbates:
Perhaps I am missing something obvious but did you remember to add the `Project.toml` file?

dmbates:
Would it do to just copy `Project.toml` from the "terms2.0" branch?

andreasnoack:
Indeed I'd forgotten to check it in. Just did. I was surprised not to find a WIP PR for the updates but didn't consider looking through the branches. Anyway, I think it should work now since tests are passing locally.

PR #183, opened 2019-07-31
Allow for `fit` to be used

Nosferican:
Allows for `StatsBase.fit` method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
   - linux
   - osx
 julia:
-  - 1.1
+  - 1.2
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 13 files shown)


dmbates:
@Nosferican Were you intending also to add a `fit!` method for `LinearMixedModel`?

Nosferican:
I was more focused on doing the `fit(GeneralizedLinearMixedModel, ...)` method.
Right now one doesn't need to do the `fit(LinearMixedModel, ...)` since by omitting the link it is smart and creates the correct implied model if not mistaken. I think
```
fit!(model::LinearMixedModel, ....)
```
isn't necessary for end-users, but it is implemented since the `GeneralizedLinearMixedModel` has the `LinearMixedModel` as a field. I think I saw that. Do let me know your thoughts and I can check the PR for it after rebasing it.

Nosferican:
This should be ready for review.

Nosferican:
The 32-bit infrastructure issue on Windows seems unrelated. Probably worth investigating after this PR.

PR #184, opened 2019-07-31
Update to Project.toml

Nosferican:
- Some extras were not required
- Moved to `test/Project.toml`
- Semver compat
- Export some missing StatsBase API


Nosferican:
@dmbates I think this is ready for review. After this PR, I can rebase and #183 should pass then.

dmbates:
@Nosferican First, thank you very much for making these two PRs.  I really appreciate your doing this.

It appears that the Travis failure was a timeout problem that occurred with Julia v"1.1.1" but not v"1.2.0-rc2".  Is that the way you see it too?

The Appveyor problem also looks like it might be v"1.1.1" vs v"1.2.0-rc2" related but has the added confounding of Windows 32-bit versus Windows 64-bit.  Again, is that the way you see it?

To tell the truth I wouldn't be grief-stricken if the package could not be installed on Windows 32-bit.

dmbates:
@Nosferican This looks good to me.  I do have a comment on #183 though.

Nosferican:
My pleasure.

If not mistaken, Julia `v"1.1.0"` still uses the Pkg.jl `extras`/`targets` while Julia `v"1.2"` on uses the `test/Project.toml`/`docs/Project.toml` (has `extras`/`targets` as deprecation period). For end users, this change has no impact. It only affects the developer side and C.I.

It shouldn't affect anything on the Windows 32-bit. If any testing CI infrastructure changed other than upping the version from `v"1.1"` to `v1.2` it should be reverted. Otherwise, it should be good.

Issue #149, opened 2018-12-13
Issue with interactions of random effects


vargovick:
I'm currently having issues fitting a model with interactions in the random effects. This model does work in R with lme4. The model I'm trying to fit is:
`mdl = GeneralizedLinearMixedModel(@formula(response ~ 1 + (1|x/y)), dat, Bernoulli());`

It works without the y term, i.e.:
`mdl = GeneralizedLinearMixedModel(@formula(response ~ 1 + (1|x)), dat, Bernoulli());`
works just fine.

The error I get is:
`MethodError: no method matching getindex(::DataFrames.DataFrame, ::Expr)
Closest candidates are:
  getindex(::DataFrames.DataFrame, !Matched::Colon) at C:\Users\Danny.Vargovick\.julia\packages\DataFrames\KUWId\src\dataframe\dataframe.jl:261
  getindex(::DataFrames.DataFrame, !Matched::Colon, !Matched::Colon) at C:\Users\Danny.Vargovick\.julia\packages\DataFrames\KUWId\src\dataframe\dataframe.jl:321
  getindex(::DataFrames.DataFrame, !Matched::Union{Integer, Symbol}) at C:\Users\Danny.Vargovick\.julia\packages\DataFrames\KUWId\src\dataframe\dataframe.jl:249
  ...
Stacktrace:
 [1] (::getfield(StatsModels, Symbol("##18#19")){DataFrames.DataFrame})(::Expr) at C:\Users\Danny.Vargovick\.julia\packages\StatsModels\AYB2E\src\modelframe.jl:145
 [2] iterate at .\generator.jl:47 [inlined]
 [3] collect_to!(::Array{Array{Int64,1},1}, ::Base.Generator{Array{Any,1},getfield(StatsModels, Symbol("##18#19")){DataFrames.DataFrame}}, ::Int64, ::Int64) at .\array.jl:656
 [4]`

I'm running Julia version 1.0.2. 

Any help is much appreciated!

kleinschmidt:
That slash syntax on the groups isn't currently supported (which is the cryptic error your seeing) and I'm actually not sure what exactly it does in lme4; fit random effects of `x` nested in `y`?

Maybe @dmbates can suggest an alternative way of specifying the same model.  Perhaps with separate random effects by `x` and `y`?

vargovick:
Thanks, @kleinschmidt. 

[...]

PR #197, opened 2019-09-25
[WIP] Support for interactions and explicit nesting in blocking variables

palday:
Addresses #149 .

Added functionality:

- [ ] Support interactions in blocking variables `(1|batch&cask)` (equivalent to R `(1|batch:cask)`).
- [x] Support nesting syntax  `(1|batch/cask)` via expansion to outer and nested blocks `(1|batch) + (1|batch&cask)`

I'm not sure that adding the nesting syntax is truly desirable (I find it horribly confusing and it emphasizes an outdated computational perspective), but it's relatively low hanging fruit once we have the interactions in blocking variables. It's also convenient to support a wider range of the formulas `lme4` supports for some of the work with RCall and JuliaCall.

Alternatively, we could just not support this syntax and add some documentation discussing how to prepare your data to use existing syntax.


palday:
Allowing a single term to expand into multiple terms during `apply_schema` would help here and also provide a relatively straightforward way to address JuliaStats/StatsModels.jl#119 .

palday:
@dmbates I think this is good for merge if you're okay with it.

PR #201, opened 2019-09-27
Force non-zero step size for general optimization after boundary fit in fast=true

### LINKED ISSUE #94: Gamma models convergence checks and run time
I am trying to fit a Gamma model to a reaction time data (there is a recent paper that argues for not transforming RTs and using GLMM with Gamma or Inverse Gaussian --> http://journal.frontiersin.org/article/10.3389/fpsyg.2015.01171/full )

I have a few question about running GLMMs in Julia, but before asking them I will go over a reproducible example using a known dataset (my dataset is large and takes too long to run). I am new to GLMMs and Julia, so I might have done something wrong. Let me know if that's the case.

Using Baayen's data from langugaeR package in R, I ran 4 models, 3 out of which do not make sense for this data. There are number of variables, but I will focus on the native language (English and other) and word class (Animal and Plant).  Also, I've converted RTs to raw RTs, as in the dataset they are log-ed.

    using DataArrays, DataFrames , MixedModels, GLM, RDatasets, RData
    #Baayen's data from langugaeR package in R saved as .csv from R
    lexdec=readtable("lexdec.csv", makefactors=true)

	lexdec_1 = fit!(glmm(@formula(rt_raw ~ 1 + Class*NativeLanguage + (1+Class*NativeLanguage|Subject) + (1|Word)), lexdec, Gamma(), IdentityLink()))
	   ERROR: MethodError: no method matching getindex(::DataFrames.DataFrame, ::Expr)

	lexdec_2 = fit!(glmm(@formula(rt_raw ~ 1 + Class*NativeLanguage + (1+Class+NativeLanguage|Subject) + (1|Word)), lexdec, Gamma(), IdentityLink()))

	lexdec_3 = fit!(glmm(@formula(rt_raw ~ 1 + Class*NativeLanguage + (1+NativeLanguage|Subject) + (1|Word)), lexdec, Gamma(), IdentityLink()))

	lexdec_4 = fit!(glmm(@formula(rt_raw ~ 1 + Class*NativeLanguage + (1+Class|Subject) + (1|Word)), lexdec, Gamma(), IdentityLink()))

Lexdec_1 did not run. I suppose the interaction term is not supported in the random effects. Lexdec_2, 3, 4 ran without a problem. However, Lexdec_2 and 3 do not make sense for this data as random slopes for  native language are inappropriate here (each person has only one language background, so there's no data to estimate how that person would behave if they had the other language background). Only Lexdec_4 is a valid model here. Nevertheless, all models except lexdec_1 ran just fine. I also ran the models using lmm in Julia with identical results on convergence.

For comparison I ran Gamma models in R using lme4 and glmmTMB. glmer from lme4 sucsesfully ran all 4 models (including the interaction model). glmmTMB gave convergence warnings on Lexdec_1, 2, and 4, while no warning on Lexdec_3. 

I observed different results when running  lmm models in R using lme4 and glmmTMB. lmer from lme4 gave warnings on Lexdec_1 and 3, while glmmTMB gave warnings on Lexdec_1, 2, and 3 (was glmmTMB able to catch the inappropriate models?)

My questions:
1. Are there any convergence checks? I can run any model with any parameters without getting any convergence warnings.
2. Are interactions of random effects supported?
3. What can I do to decrees runtime? My dataset has ~20k observations and there are 2 within factors with 3 levels each and one between factor with 2 levels.
4. Are there any plans to implement support for Inverse Gaussian? I ran Inverse Gaussian models using lme4 and glmmTMB with identical results, so I guess there is not much of difference from RT data.

I am running arch linux 4.12.3-1-ARCH with openblas-lapack installed from AUR.

```
julia> versioninfo()
Julia Version 0.6.0
Commit 903644385b* (2017-06-19 13:05 UTC)
Platform Info:
 OS: Linux (x86_64-pc-linux-gnu)
 CPU: Intel(R) Core(TM) i7-4710HQ CPU @ 2.50GHz
 WORD_SIZE: 64
 BLAS: libopenblas (HASWELL)
 LAPACK: liblapack
 LIBM: libm
 LLVM: libLLVM-4.0.0 (ORCJIT, haswell)

 julia> Pkg.status()
  - MixedModels 0.10.0
  - GLM 0.7.0
```

### LINKED ISSUE #95: Watch for boundary cases from fast=true in glmm fits
If one of the constrained elements of θ converges to zero in the `fast=true` stage when fitting a `glmm` then it should be given a nonzero starting value in the general optimization.

palday:
Resolves #94, Closes #95. 

For boundary fits in GLMM, the initial step for general optimization in `fast=false` includes a zero in (at least) one dimension. This forces that to be nonzero by replacing zeros with the smallest nonzero value.

Note: this will not address models where everything is zero (including fixed effects), but that's another can of worms.


palday:
Stung by floating point comparison. I'll fix that.

palday:
Optimization gives different results on different platforms, but neither is the 'correct' answer, see #206.

Issue #174, opened 2019-03-30
Inconsistency in Bootstrap docs


nalimilan:
The [manual page on bootstrap](http://dmbates.github.io/MixedModels.jl/latest/bootstrap/) seems to have a few small inconsistencies. First, it starts mentioning `bootstrap!`, but then presents `bootstrap`. Then it says "The results for each bootstrap replication are stored in the columns of the matrix passed in as the first argument.", but the example is `bootstrap(100_000, m1)`.

Besides that, the docs are great!

Issue #172, opened 2019-03-28
calculating bootstrap confidence intervals


jen265:
I tried to calculate boostrap confidence intervals for my linear mixed effects model (name: fm1) with MixedModels.jl using the syntax below (for simplicity, I dropped 8 predictors). Unfortunately, printing "bootstrap(10000, fm1)" displays the fixed effects in a random order, which makes it impossible to calculate the confidence interval for a specific fixed effect. Is it possible that printing "bootstrap(10000, fm1)" displays the fixed effects in the order they appear in the mixed model?
Any help is very much appreciated.

modbstp = bootstrap(10000, fm1);
print(names(modbstp))
ci_intercept = percentile(modbstp[:β01], [2.5, 97.5])
ci_agr = percentile(modbstp[:β02], [2.5, 97.5])
ci_cns = percentile(modbstp[:β03], [2.5, 97.5])

palday:
Bootstrap will change a bit when #202 lands. You can try out the new functionality by installing that branch:
```julia
# get here by typing the right square bracket at the julia prompt
# get out of here by backspacing on an empty line
pkg> add MixedModels#bootstraptuples 
```
And if you want to go back to the main release:
```julia
pkg> free MixedModels
```

PR #202, opened 2019-09-27
Bootstraptuples

dmbates:
Change the type of object returned by `parametricbootstrap` to be a `Tables.ColumnTable` with fixed names for the objective, the residual standard deviation, the coefficient estimates and the theta parameters.


palday:
Closes #172 . Could resolve #174. 

(Have the docs been updated for the new bootstrap?)

src/MixedModels.jl
```diff
@@ -82,6 +83,7 @@ export @formula,
        refit!,
        residuals,
        response,
+       shortestCovInt,
```
nalimilan: I'm quite late to the party, but this name doesn't follow the Julia conventions (lowercase).

PR #204, opened 2019-09-30
Handle implicit intercept and allow full dummy coding in ranefs

### LINKED ISSUE #195: (Suppression of) intercept term in lhs of random effects
I just saw that with a experimental factor on the left hand side of random-effects term, suppressing the intercept does not give the desired model.

I can fit a model with an experimental factor, `f`, in the random effects specification for the grouping factor, `g`, and get the expected result.
```julia
julia> using DataFrames, MixedModels

julia> dat = DataFrame(y = ones(18),
            g = string.(repeat('a':'f', inner=3)),
            f = string.(repeat('A':'C', outer=6)));

julia> mod = LinearMixedModel(@formula(y ~ 1 + f + (1+f|g)), dat);

julia> refit!(simulate!(mod, β = [1.0, -0.2, 0.3], σ = 0.05,
                        θ = [1.0, -0.1, 0.3, 0.2, 0.1, 0.3]))
 y ~ 1 + f + (1 + f | g)
   logLik   -2 logLik     AIC        BIC    
  24.297916 -48.595831 -28.595831 -19.692114

Variance components:
              Column      Variance     Std.Dev.    Corr.
 g        (Intercept)  0.00713718070 0.084481836
          f: B         0.00060249946 0.024545864  0.54
          f: C         0.00282675031 0.053167192  0.35 -0.48
 Residual              0.00067137974 0.025910997
 Number of obs: 18; levels of grouping factors: 6

  Fixed-effects parameters:
────────────────────────────────────────────────────
              Estimate  Std.Error   z value  P(>|z|)
────────────────────────────────────────────────────
(Intercept)   0.994279  0.0360753   27.5612   <1e-99
f: B         -0.221398  0.0180058  -12.2959   <1e-34
f: C          0.351254  0.0263613   13.3246   <1e-39
────────────────────────────────────────────────────

julia> dat.y = mod.y;

julia> mod0 = fit(MixedModel, @formula(y ~ 0 + f + (0+f|g)), dat)
 y ~ 0 + f + (0 + f | g)
   logLik   -2 logLik     AIC        BIC    
  17.362429 -34.724857 -20.724857 -14.492255

Variance components:
           Column   Variance   Std.Dev.    Corr.
 g        f: B    0.006736238 0.08207459
          f: C    0.009447073 0.09719606  1.00
 Residual         0.005346109 0.07311709
 Number of obs: 18; levels of grouping factors: 6

  Fixed-effects parameters:
───────────────────────────────────────────
      Estimate  Std.Error  z value  P(>|z|)
───────────────────────────────────────────
f: A  0.994279  0.0298499  33.3092   <1e-99
f: B  0.77288   0.0448745  17.2231   <1e-65
f: C  1.34553   0.0496541  27.0981   <1e-99
───────────────────────────────────────────
```
Notice that the vector-valued random effects in the second model, using a term of the form `(0+f|g)`, are two-dimensional, not three-dimensional like β.

Something in `src/randomeffectsterm.jl` needs to be modified in the way the modelcols for the lhs is evaluated.  I may need help from @kleinschmidt on this one.

kleinschmidt:
This fixes #195 by adding an intercept term on the LHS of random effects term
when one is not present, and the intercept is not explicitly omitted with `0`.
It also adds the `fulldummy` syntax to manually promote a categorical term to
`FullDummyCoding`, which makes it possible to specify each of the following LHS:

1. `(f | g)` - implicit intercept, default contrasts (full rank)
2. `(1 + f | g)` - manual intercept, default contrasts (equivalent to 1)
3. `(0 + f | g)` - no intercept, full dummy coding (also full rank)
4. `(1 + fulldummy(f) | g)` - intercept with full dummy coding (overparametrized)

1-3 are the same as how fixed effects are handled, but 4 is special and possibly useful for investigating the effect of regularization etc. (cc. @kliegl)


palday:
Was failing on Julia 1 on Linux due to a timeout downloading `Coverage.jl` (this seems to be a common problem with Julia 1.0; I really won't miss supporting it...), but I've restarted that CI.

src/randomeffectsterm.jl
```diff
@@ -53,3 +59,24 @@ function StatsModels.modelcols(t::RandomEffectsTerm, d::NamedTuple)
...
+                        "coding (only CategoricalTerms)"))
+
+function fulldummy(t::CategoricalTerm)
+    new_contrasts = StatsModels.ContrastsMatrix(StatsModels.FullDummyCoding(),
+                                                t.contrasts.levels)
+    t = CategoricalTerm(t.sym, new_contrasts)
+end
+
+fulldummy(x) =
+    throw(ArgumentError("fulldummy can't be used outside of a mixedmodel formula"))
```
palday: Should this be `MixedModel` to refer to the package (and superclass) or `mixed-model` to refer to the concept?

src/randomeffectsterm.jl
```diff
@@ -23,9 +23,15 @@ function StatsModels.termvars(t::RandomEffectsTerm)
...
 
-function StatsModels.apply_schema(t::FunctionTerm{typeof(|)}, schema::StatsModels.FullRank,
-        Mod::Type{<:MixedModel})
-    lhs, rhs = apply_schema.(t.args_parsed, Ref(schema), Mod)
+function StatsModels.apply_schema(t::FunctionTerm{typeof(|)},
+                                  schema::StatsModels.FullRank,
+                                  Mod::Type{<:MixedModel})
+    schema = StatsModels.FullRank(schema.schema)
+    lhs, rhs = t.args_parsed
+    if !StatsModels.hasintercept(lhs) && !StatsModels.omitsintercept(lhs)
```
palday: I understand the logic of this line (neither explicit "commission" nor explicit omission), but the surface appearance is "if both mutually exclusive conditions hold" .... (no change necessary)

src/randomeffectsterm.jl
```diff
@@ -23,9 +23,15 @@ function StatsModels.termvars(t::RandomEffectsTerm)
...
+        lhs = InterceptTerm{true}() + lhs
```
palday: This is also much nicer than the originally proposed fix because the LHS stays on the left (even if the LHS of a `|` is essentially the RHS or a `~`.

test/ranefterm.jl
```diff
@@ -0,0 +1,56 @@
+using MixedModels, StatsModels
```
palday: Should the tests "RandomEffectsTerm" and  "Categorical Blocking Variable" be moved from `FactorReTerm.jl` to here?

Issue #203, opened 2019-09-30
missing rankUpdate! method for repeated blocking factor


palday:
Splitting the LHS of a single blocking factor across multiple RE groups breaks things.

```julia
julia> fit!(LinearMixedModel(@formula(Reaction ~ 1 + Days + (1|Subject) + (0+Days|Subject)), rcopy(R"sleepstudy")))
ERROR: MethodError: no method matching rankUpdate!(::LinearAlgebra.Diagonal{Float64,Array{Float64,1}}, ::LinearAlgebra.Diagonal{Float64,SparseArrays.SparseVector{Float64,Int32}}, ::Float64)
```
I think the idea was that everything with a shared blocking should be aggregated and that the zero-correlation structure should be achieved with `zerocorr!`, so these two specifications should yield identical blocking structures:

```julia
julia> describeblocks(LinearMixedModel(@formula(Reaction ~ 1 + Days + (1+Days|Subject)), rcopy(R"sleepstudy")))
1,1: UniformBlockDiagonal{Float64} (36, 36) UniformBlockDiagonal{Float64}
2,1: Array{Float64,2} (2, 36) Array{Float64,2}
2,2: Array{Float64,2} (2, 2) Array{Float64,2}
3,1: Array{Float64,2} (1, 36) Array{Float64,2}
3,2: Array{Float64,2} (1, 2) Array{Float64,2}
3,3: Array{Float64,2} (1, 1) Array{Float64,2}

julia> describeblocks(LinearMixedModel(@formula(Reaction ~ 1 + Days + (1|Subject) + (0+Days|Subject)), rcopy(R"sleepstudy")))
1,1: LinearAlgebra.Diagonal{Float64,Array{Float64,1}} (18, 18) LinearAlgebra.Diagonal{Float64,Array{Float64,1}}
2,1: LinearAlgebra.Diagonal{Float64,SparseArrays.SparseVector{Float64,Int32}} (18, 18) LinearAlgebra.Diagonal{Float64,SparseArrays.SparseVector{Float64,Int32}}
2,2: LinearAlgebra.Diagonal{Float64,Array{Float64,1}} (18, 18) LinearAlgebra.Diagonal{Float64,Array{Float64,1}}
3,1: Array{Float64,2} (2, 18) Array{Float64,2}
3,2: Array{Float64,2} (2, 18) Array{Float64,2}
3,3: Array{Float64,2} (2, 2) Array{Float64,2}
[...]

PR #207, opened 2019-10-01
add zerocorr within formula

kleinschmidt:
This provides special syntax to be used within a formula to zero out
correlations on random effects terms:

``` julia
LinearMixedModel(@formula(Y ~ 1 + U + zerocorr(1+U|G)), dat)
```

which is equivalent to 

``` julia
zerocorr!(LinearMixedModel(@formula(Y ~ 1 + U + (1+U|G)), dat))
```

### DIFF
--- a/docs/jmd/constructors.jmd
+++ b/docs/jmd/constructors.jmd
@@ -66,17 +66,6 @@ A model with random intercepts and random slopes for each subject, allowing for 
 fm2 = fit(MixedModel, @formula(Y ~ 1 + U + (1+U|G)), dat[:sleepstudy])
 ```
 
-A model with uncorrelated random effects for the intercept and slope by subject is fit as
-```{julia;term=true}
-fm3 = fit!(zerocorr!(LinearMixedModel(@formula(Y ~ 1 + U + (1+U|G)), dat[:sleepstudy])))
-```
-
-Note that the use of `zerocorr!` requires the model to be constructed, then altered to eliminate
-the correlation of the random effects, then fit with a call to the mutating function, `fit!`.
-```@docs
-zerocorr!
-```
-
 ### Models with multiple, scalar random-effects terms
 
 A model for the *Penicillin* data incorporates random effects for the plate, `G`, and for the sample, `H`.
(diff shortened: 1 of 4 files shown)


palday:
Related to #203: how do the two different formulations work when there are multiple RE terms with the same grouping variable?  Consider `...  (1+f|g) + (1+e|g)`.

`zerocorr!(m, :g)` zeroes all the rho's. `zerocorr` for both the `f` and `e` terms does the same. `zerocorr` for the `f` but not the `e` zeros those between levels of `f` and the intercept, and if we do the lme4 interpretation of the repeated grouping variables, then the rho'ss between `f` and `g` are also zeroed, but not the rho's between levels of `e` and the intercept. Right?

(More generally: how do we handle distinct terms with the same blocking variable? I think we need to make this explicit somewhere.)

kleinschmidt:
I thought it would be relatively straightforward to extend this to also further restrict the lambda (e.g., to use a scaled diagonal matrix, with all the variance components the same) but there's one more level of indirection necessary to do that, to map a single element of theta into multiple elements of lambda.

kleinschmidt:
@palday maybe a docs update is needed here, too.  At the moment, anything listed in a separate `(...|...)` block is going to be uncorrelated.  `zerocorr` just changes the correlation structure _within_ the block.

palday:
Right, `zerocorr` is term-local but `zercorr!` applies to all terms with the same blocking variable.

kleinschmidt:
Will rebase on #208 so that docs build from jmd

palday:
@kleinschmidt should we treat further vcov restrictions as something for the indeterminate future and a new PR?

palday:
Ah and the docs update is blocked until #209 lands.

kleinschmidt:
Yes I think let's get this merged and then think about whether to support more restricted vcov (like uniform diagonal etc.).  #209 is blocking the docs update because there's an error with one of the formulations right?

palday:
Right. Currently repeats of the same blocking variable across multiple RE terms breaks things. The tricky part isn't lumping them together but rather zeroing out the correlations implied by their independent listing.

kleinschmidt:
Okay I'll wait for #209 and then rebase this

palday:
@kleinschmidt  go for rebase

kleinschmidt:
Rebased and CI is green.

PR #208, opened 2019-10-01
run docs/weave.jl and docs/make.jl in docs stage

### LINKED ISSUE #174: Inconsistency in Bootstrap docs

kleinschmidt:
This should re-generate the markdown files from their jmd sources before
building the docs with documenter.

Fixes #174


kleinschmidt:
bootstrap.jmd fails as soon as it hits `parametricbootstrap` in line 44.  @dmbates any idea what's going on there?  I can't reproduce locally running 1.2 on linux so I'm stumped.

kleinschmidt:
After rebasing onto master it looks like the docs build without error so this is ready to merge I think.  It also incorporates some doc fixes from @dmbates so we should take a new version, but there has been some some other changes since the last version bump so maybe that should wait for a separate PR.

kleinschmidt:
Unclear...the md and svg files are now untracked.  Maybe good to be on the safe side to prevent people from adding them back in but they'll just get overwritten during the docs stage, and won't show up in clones so there's no temptation to try to edit them.

PR #210, opened 2019-10-02
Restore printing of variance components

dmbates:
- Switch to `NamedTuple` representation of standard deviations and correlations using the `m.σρs` extractor.
- `VarCorr` simply contains this `NamedTuple` and the dispersion parameter.
- Clean up the show method for `VarCorr`.
- Add tests for more coverage.


test/pls.jl
```diff
@@ -67,6 +82,7 @@ const LMM = LinearMixedModel
 
     fit!(fm1, REML=true)
     @test objective(fm1) ≈ 319.65427684225216 atol=0.0001
+    @test_throws ArgumentError loglikelihood(fm1)
```
palday: :+1:

PR #211, opened 2019-10-03
RCall support for lme4 and MixedModels

palday:
Functionality:
- [x] `lmerMod` to `LinearMixedModel`
- [ ] `glmerMod` to `GeneralizedLinearMixedModel`
- [x]  `LinearMixedModel` to `lmerMod`
- [ ]  `GeneralizedLinearMixedModel` to `glmerMod`

Tests:
- [x] `lmerMod` to `LinearMixedModel`
- [ ] `glmerMod` to `GeneralizedLinearMixedModel`
- [x]  `LinearMixedModel` to `lmerMod`
- [ ]  `GeneralizedLinearMixedModel` to `glmerMod`

CIs:
- [ ] Travis Linux
- [ ] Travis Mac
- [ ] Appveyor

Note that not all the optimization information is preserved and certain formulas (e.g. double-bar 
syntax in R) don't translate. Because of the slightly different internal representations, the data and theta vectors are copied over, but everything else is recomputed as necessary on the other side.


palday:
One major restriction is that you need to `@rput Tuple([MixedModel, DataFrame])` because lme4 has a slot for the original dataframe but MixedModels doesn't keep that around. (And because we're cheating and letting lme4 re-assemble the internal representation rather than converting MixedModels' internal rep, which is slightly different.)

palday:
This has been spun off into a [separate package, `JellyMe4`](https://github.com/palday/JellyMe4.jl). Two advantages to this approach:
1. Incremental increase in support for different types of models on its own release cycle separate from MixedModels.jl
2. `RCall`, etc. do not need to be added to `MixedModel`'s dependencies.

PR #212, opened 2019-10-03
Fix error from returning SVector for theta

kleinschmidt:
`parametricbootstrap` tries to write to elements of theta, but the theta that is
returned by getproperty is an SVector, so this errors

@palday I'll push a fix once the error is confirmed

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -323,9 +323,7 @@ end
 
 function Base.getproperty(m::LinearMixedModel, s::Symbol)
     if s == :θ || s == :theta
-        v = getθ(m)
-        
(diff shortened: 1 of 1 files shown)


kleinschmidt:
I thought that replacing property access with getθ would fix it but there's another place where the property is accessed and then passed to the GLMM constructor which only takes `Array`, not `SArray` or `AbstractArray` so i just dropped the SVector wrapper altogether (since it looks like the θ returned during bootstrapping are going to be wrapped in an SVector anyway for storage...)

palday:
@dmbates This seems like a by-product of some of the rebase-ry we were doing, so I'm merging it in. If that's not desired, we'll do a revert commit later.

PR #217, opened 2019-10-16
update BlockArrays support.

### LINKED ISSUE #215: Remove BlockArrays dependency
@dmbates, do you have any bandwidth for removing the BlockArrays dependency?

From,
> I just saw something else that will eventually need to be addressed.  The `BlockArrays` package changed after v0.7.0 to require the underlying arrays to be the same type.  This package uses heterogeneous `BlockArrays` as shown by the output of `describeblocks` and the version of `BlockArrays` must be capped at v0.7.0.
>
> Only a small portion of the capabilities of BlockArrays are used for `A` and `L` in a `LinearMixedModel`.  It is probably best to remove the dependency on `BlockArrays` and write custom code to organize and access the blocks.

_Originally posted by @dmbates in https://github.com/dmbates/MixedModels.jl/issues/185#issuecomment-520928646_

This omen is already manifested in a few projects. I believe from your comment and examining the code that it should be

https://github.com/dmbates/MixedModels.jl/blob/1745530694b13b27807c80d1dd10c8fb90edc807/src/linearmixedmodel.jl#L70-L93

Let me know if I could assign you this one or if I can try to take a stab at it with some guidance (i.e., I am not familiar with the BlockArrays package).

palday:
Uses the suggestion from #216 to use `AbstractMatrix` as block type that doesn't lose subtype information to address `BlockArrays > 0.7.0` compatibility.
```julia
julia> B = Matrix{AbstractMatrix{Float64}}(undef,2,1)

julia> B[1,1] = fill(1.0,2,2)

julia> B[2,1] = sparse(I,2,2)

julia> A = mortar(B)
2×1-blocked 4×2 BlockArray{Float64,2,Array{AbstractArray{Float64,2},2},BlockArrays.BlockSizes{2,Tuple{Array{Int64,1},Array{Int64,1}}}}:
 1.0  1.0
 1.0  1.0
 ────────
 1.0  0.0
 0.0  1.0

julia> A[Block(1,1)]
2×2 Array{Float64,2}:
 1.0  1.0
 1.0  1.0

julia> A[Block(2,1)]
2×2 SparseMatrixCSC{Float64,Int64} with 2 stored entries:
  [1, 1]  =  1.0
  [2, 2]  =  1.0
```

_Originally posted by @dlfivefifty in https://github.com/dmbates/MixedModels.jl/pull/216#issuecomment-542524324_

 Resolves #215. Depends on #216 (and will be rebased once that lands).

Note that I also removed support for `BlockArrays 0.3, 0.4` because a preliminary glance at the release notes suggests those versions don't support Julia 1.2. I've tested locally on `BlockArrays 0.5, 0.7, 0.8, 0.9, 0.10`, but I haven't done any performance comparisons.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
 
 [compat]
-BlockArrays = "0.3, 0.4, 0.5, 0.6, 0.7"
+BlockArrays = "0.5, 0.6, 0.7, 0.8, 0.9, 0.10"
 Distributions = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21"
 GLM = "1"
 NLopt = "0.5"
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -77,8 +77,8 @@ function LinearMixedModel(f::FormulaTerm, tbl::Tables.ColumnTable;
     terms = vcat(reterms, feterms)
     k = length(terms)
     sz = append!(size.(reterms, 2), rank.(feterms))
-    A = BlockArrays._BlockArray(AbstractMatrix{T}, sz, sz)
-    L = BlockArrays._BlockArray(AbstractMatrix{T}, sz, sz)
+    A = BlockArray(undef_blocks, AbstractMatrix{T}, sz, sz)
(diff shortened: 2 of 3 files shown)


palday:
@Nosferican Care to test whether you can install from this branch and still have Pumas / whatever incompatible packages installed?

Nosferican:
🎊🎊🎊 I had to open https://github.com/piever/ShiftedArrays.jl/pull/28, but with that merged everything goes back to playing well with each other.

PR #230, opened 2019-12-09
Added the compat bounds for missing entries

Nosferican:
Added the compat bounds for missing entries.


Nosferican:
Supersedes #224, #225, #226, #227, #228, #229.

PR #237, opened 2020-01-07
Replace multiplications, etc. with small matrices by explicit loops

dmbates:
Operations involving the potentially large matrix Λ involve stepping across blocks of columns (or rows) and applying a small triangular matrix,  λ.  By writing the inner operation as explicit loops instead of using views, a considerable amount of storage allocation can be avoided.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -72,6 +72,24 @@ function LinearAlgebra.ldiv!(
     B
 end
 
+"""
+    _rdivlowertriblk!(A::AbstractMatrix, Bd::Array, b, S)
+
+Divide in-place a block of `S` columns of `A` starting at `(b-1)*s+1` on the right by the lower-triangular `b`th face of Bd.
+"""
+function _rdivlowertriblk!(A::AbstractMatrix, Bd::Array, b, S)
+    coloffset = (b - 1) * S
+    @inbounds @simd for i = 1:size(A, 1)
+        for j = 1:S
+            Aij = A[i, j + coloffset]
+            for k = 1:j-1
+                Aij -= A[i, k + coloffset] * Bd[j, k, b]'
+            end
+            A[i, j + coloffset] = Aij / Bd[j, j, b]'
+        end
+    end
+end
+
 function LinearAlgebra.rdiv!(
     A::Matrix{T},
     adjB::Adjoint{T,<:LowerTriangular{T,UniformBlockDiagonal{T}}},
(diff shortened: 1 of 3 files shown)


src/remat.jl
```diff
@@ -386,13 +386,22 @@ end
...
+function rmulΛ!(A::M, B::ReMat{T,S}) where {M<:AbstractMatrix{T}} where {T,S}
     m, n = size(A)
     q, r = divrem(n, S)
-    iszero(r) || throw(DimensionMismatch("size(A, 2) = is not a multiple of block size"))
-    A3 = reshape(A, (m, S, q))
-    for k in 1:q
-        rmul!(view(A3, :, :, k), B.λ)
+    iszero(r) || throw(DimensionMismatch("size(A, 2) is not a multiple of block size"))
+    Bd = B.λ.data
+    @inbounds for k = 1:q
```
palday: I like the comment below on which multiplication these loops are replacing. Can you do the same here?

src/linalg.jl
```diff
@@ -76,12 +76,21 @@ function LinearAlgebra.rdiv!(
...
+    m, n = size(A)
     Bd = adjB.parent.data
-    m, n, k = size(Bd.data)
-    size(A, 2) == size(Bd, 1) && m == n || throw(DimensionMismatch(""))
-    inds = 1:m
-    for (i, f) in enumerate(Bd.facevec)
-        BLAS.trsm!('R', 'L', 'T', 'N', one(T), f, view(A, :, inds .+ m * (i - 1)))
+    r, s, blk = size(Bd.data)
+    n == size(Bd, 1) && r == s || throw(DimensionMismatch())
+    @inbounds for b = 1:blk
```
palday: I like the comment below on which multiplication these loops are replacing. Can you do the same here?

palday:
Do you have a good benchmark suite? I realize that precise timings are something that won't be valid across machines, but we should probably take some of the build test suite and see if it can be adapted as a benchmarking one. Add this to our wishlist of 'easy' things to do when I need a distraction....

dmbates:
> Do you have a good benchmark suite? I realize that precise timings are something that won't be valid across machines, but we should probably take some of the build test suite and see if it can be adapted as a benchmarking one. Add this to our wishlist of 'easy' things to do when I need a distraction....

There is a set of benchmarks for use with the `PkgBenchmark` package in the benchmark directory but I haven't run it for some time and it will probably need to be updated.  I will do that today.

PR #240, opened 2020-01-23
Switch bootstrap to return static vectors and a model copy

dmbates:
- modifications to `parametricbootstrap` to return a struct with a copy of the model for later analysis
- properties like `σs` and `σρs` to extract the random effects standard deviation and correlation estimates


palday:
Have you tried this with a model from a really tall dataset yet?

dmbates:
I haven't tried this with a tall data set.  It will, naturally, take longer because it does simulate the entire response vector. However, only the last column of blocks of `A` end up being updated (because the other's won't change).  It is certainly worthwhile doing some benchmarking to decide where the time is being spent but my guess is that the repeated optimization of the parameter estimates will overwhelm the time spent updating the structures for the simulated data sets.

Are you hinting at the possibility of simulating from the distribution of some "sufficient statistics" rather than simulating the entire response vector?  That might be possible.

palday:
I was actually thinking about the cost of `deepcopy` on a tall model / its model matrix. I like the idea of still having access to the model, but if you have 2e9 observations, then you have a large model matrix, even for simple random effects.

dmbates:
The reason for keeping a copy of the model from which the samples were generated is to allow for transforming (σ, θ) stored in the `bstr` field to standard deviations and correlations of random effects.  That's what the `byreterm` function is doing.  We may be able to get away with just keeping a copy of the `reterms` field of the model but even that is sort-of overkill because it is the small, simple fields, λ and `inds`, of the reterm that are used, not the large fields.

Why don't we keep a watch on this to see if it will be a problem and, if so, switch to storing the λ and `inds` for each `ReMat` only.

PR #241, opened 2020-01-28
Allow BlockArrays 0.11 by not calling nblocks

dmbates:
- the `BlockArrays` API changed in version 0.11 to remove the `nblocks` generic.
- will supersede #233


src/linearmixedmodel.jl
```diff
@@ -776,9 +778,9 @@ This is the crucial step in evaluating the objective, given a new parameter valu
 function updateL!(m::LinearMixedModel{T}) where {T}
     A = m.A
     L = m.L
-    k = BlockArrays.nblocks(A, 2)
+    k = length(m.allterms)
     for j = 1:k                         # copy lower triangle of A to L
-        for i = j:BlockArrays.nblocks(A, 1)
+        for i = j:length(m.allterms)
```
palday: In the original, this was done for two different dimensions, but since it's lower triangular, I guess this doesn't matter and actually makes things more consistent/obvious.

src/linearmixedmodel.jl
```diff
@@ -78,7 +79,7 @@ function LinearMixedModel(
     sort!(reterms, by = nranef, rev = true)
 
     # create A and L
-    terms = vcat(reterms, feterms)
+    terms = convert(Vector{Union{ReMat{T},FeMat{T}}}, vcat(reterms, feterms))
```
palday: I think this should be changed to `allterms` -- see next comment.

src/linearmixedmodel.jl
```diff
@@ -105,7 +106,7 @@ function LinearMixedModel(
     θ = foldl(vcat, getθ(c) for c in reterms)
     optsum = OptSummary(θ, lbd, :LN_BOBYQA, ftol_rel = T(1.0e-12), ftol_abs = T(1.0e-8))
     fill!(optsum.xtol_abs, 1.0e-10)
-    LinearMixedModel(form, reterms, feterms, sqrt.(convert(Vector{T}, wts)), A, L, optsum)
+    LinearMixedModel(form, reterms, feterms, terms, sqrt.(convert(Vector{T}, wts)), A, L, optsum)
```
palday: why does `terms` need to be added here? The field in the struct is `allterms`, which seems to replace the old `terms` field in functionality.

src/linearmixedmodel.jl
dmbates: Is it better to write that as `j:k`?  I don't see an advantage in evaluating `length(m.allterms)` every iteration.  The compiler is probably smart enough to hoist that out of the loop but there is no reason to burden it.

src/linearmixedmodel.jl
dmbates: The name `terms` was already in place and I just didn't bother to change it to `allterms` here.  `terms` is a local identifier and the name of the field is `allterms`.  I agree that it would make more sense to use `allterms` for both and will make that change.

src/linearmixedmodel.jl
palday: Probably -- the definition of `k` is only two lines ahead, so it's easy to see how `k` is defined and then it's obvious that both loops have the same upper bound.

dmbates:
In working on the profiling I feel that a slightly more general approach would be better.  What I intend to do is to store only the `allterms` field and make `reterms` and `feterms` properties rather than fields.

palday:
Would it make more sense to have that the other way around in terms of which are properties and which are fields?

dmbates:
If the travis tests work out I think this should be ready to go.

src/linearmixedmodel.jl
```diff
@@ -226,7 +204,7 @@ diagonal blocks from the conditional variance-covariance matrix,
     s² Λ(Λ'Z'ZΛ + I)⁻¹Λ'
 """
 function condVar(m::LinearMixedModel{T}) where {T}
-    retrms = m.reterms
+    retrms = filter(Base.Fix2(isa, ReMat), m.allterms)
```
dmbates: This change can be reverted.  The rhs of the result is just a (infintesimally) slower version of the line it replaces.

palday: I was actually just looking at that line ... I think the original is clearer if there's no real performance difference.

palday:
Having `allterms` as the field actually simplifies a bunch of things, nice.

palday:
I'll merge as soon as the CIs are done.

palday:
The coverage check seems to be a trivial fail.

PR #249, opened 2020-02-05
Profiling fixed-effects parameters

dmbates:
Initial approach to profiling the fixed-effects parameters for a linear mixed-effects model.  The quantities that are stored are ζ, the signed square root of the change in the objective (negative twice the log-likelihood) and the parameters being optimized, θ.  Derived parameters, β and σ, are also stored.

This approach depends strongly on the fact that the fixed-effects parameters are coefficients in a linear predictor with identity link and multivariate Gaussian distributions.  To profile such a coefficient you simply subtract its contribution to the predictor from the response and fit the reduced model.

To profile other parameters such as standard deviations or correlations of the random effects it will be necessary to change the evaluation of the objective.


palday:
The failing CI was related to a failure to download Gadfly for the docs. I've restarted it.

palday:
A quick look suggests that this is as embarrassingly parallel as the bootstrap, with the same shared memory issues for the model object. Should I add in multi-threading?

dmbates:
@palday The number of optimizations of the objective is much lower for profiling than for parametric bootstrapping. The default here is on the order of 10x the number of fixed-effects parameters.  I certainly have no objections to incorporating threads as you did for the bootstrapping but it may not be very noticeable.

palday:
Then I'll probably leave it alone for now. I've been working on various parallelizable odds and ends lately, so I might come back for the practice reasoning about various concurrency issues.

dmbates:
Starting over with this because the branch is too stale.

PR #254, opened 2020-02-11
Add functions testdata and testdatasets

dmbates:
Add and export functions `testdata` and `testdatasets` to import one of the datasets for testing.


src/MixedModels.jl
```diff
@@ -99,6 +100,8 @@ export @formula,
        statscholesky,
        std,
        stderror,
+       testdata,
+       testdatasets,
        updateL!,
```
palday: Let's not export this name -- it seems likely to collide with other packages and I think the explicit `MixedModels.testdata` is nice.

src/utilities.jl
```diff
@@ -115,3 +115,18 @@ function replicate(f::Function, n::Integer; use_threads=false)
     end
     results
 end
+
+"""
+    testdata(nm)
+
+Return a data frame of test data set named `nm`, which can be a `String` or `Symbol`
+"""
+testdata(nm::AbstractString) = Feather.read(joinpath(TestData, nm * ".feather"))
```
palday: Following up on the suggestion of @Nosferican, we could rename both `testdatasets` and `testdata` to `datasets` and let the compiler figure out which methods we need.

PR #257, opened 2020-02-12
fix up nesting (slash) syntax and add tests for fixefs too

kleinschmidt:
The existing implementation of `g/x` syntax (estimates effects of `x` for each
level of grouping factor `g`) almost works for fixed effects as well.

The syntax `g/x` is equivalent to `g + fulldummy(g) & x`.

This PR adds tests for fixed effects and fixes a few oversights that only apply
in the fixed effects setting (e.g., use the `FullRank` schema to make sure that
`g` is promoted to full rank if necessary; promote `g` to full rank dummy coding
for the nested component).

Ultimately this should be moved into a separate regression formulae package.

### DIFF
--- a/src/randomeffectsterm.jl
+++ b/src/randomeffectsterm.jl
@@ -27,8 +27,14 @@ function StatsModels.apply_schema(
     if length(t.args_parsed) ≠ 2
         throw(ArgumentError("malformed nesting term: $t (Exactly two arguments required"))
     end
+
     first, second = apply_schema.(t.args_parsed, Ref(sch.schema), Mod)
-    return first + first & second
+
+    if !(typeof(first) <: CategoricalTerm)
+        throw(ArgumentError("nesting terms requires categorical grouping term, got $first.  Manually specify $first as `CategoricalTerm` in hints/contrasts"))
+    end
+
+    return first + fulldummy(first) & second
 end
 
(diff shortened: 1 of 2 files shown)


src/randomeffectsterm.jl
```diff
@@ -27,8 +27,14 @@ function StatsModels.apply_schema(
...
         throw(ArgumentError("malformed nesting term: $t (Exactly two arguments required"))
     end
-    first, second = apply_schema.(t.args_parsed, Ref(sch.schema), Mod)
-    return first + first & second
+
+    first, second = apply_schema.(t.args_parsed, Ref(sch), Mod)
+    
+    if !(typeof(first) <: CategoricalTerm)
+        throw(ArgumentError("nesting terms requires categorical grouping term, got $first.  Manually specify $first as `CategoricalTerm` in hints/contrasts"))
+    end
```
palday: I think this error is the only thing not covered by the tests. Can you add one?

palday:
I agree about having a RegressionFormulae package ... oh the Future ....

src/randomeffectsterm.jl
kleinschmidt: Yup.  And one for nesting too deep, too (in case anyone gets any ideas ;) )

PR #258, opened 2020-02-12
Draft of using GH Actions

Nosferican:
- Replaces Travis and AppVeyor for GitHub Actions


palday:
@kleinschmidt We're going to need organization access for some of these changes.

palday:
I think we should test with current release (or rather all officially supported releases) and with the next releases, but allow failure for the next releases, so that we get a heads-up on potentially breaking changes.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,47 @@
...
+    types:
+      - created
+
+jobs:
+  ci:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: [1.3]
+        julia-arch: [x64, x86]
```
dmbates: x86 might cause problems in the tests - IIRC it did in the past and I just restricted the tests to run on x64.  It may be okay here but we should watch for the possibility of problems with it.

palday:
@Nosferican How do we view the generated docs before merging?

palday:
@Nosferican and it's broken again. :(

Nosferican:
@kleinschmidt, what was the name of the secret you named the SSH key for Documenter for the GitHub Actions?

kleinschmidt:
`DOCUMENTER_KEY`

kleinschmidt:
BTW some of the math doesn't render in the preview (for instance: https://juliastats.org/MixedModels.jl/previews/PR258/optimization/)  Is there a difference in the latex renderer for this version of documenter vs. previous versions?

kleinschmidt:
I think the issue is that neither `equation` nor `aligned` are supported by KaTeX: https://github.com/KaTeX/KaTeX/issues/445

kleinschmidt:
`aligned` is fine but `equation` is not; it's also redundant so I'll push a fix.

palday:
And now the doc preview is failing on #264 
```
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
┌ Error: Git failed to fetch git@github.com:JuliaStats/MixedModels.jl.git
│ This can be caused by a DOCUMENTER_KEY variable that is not correctly set up.
│ Make sure that the environment variable is properly set up as a Base64-encoded string
│ of the SSH private key. You may need to re-generate the keys with DocumenterTools.
└ @ Documenter ~/.julia/packages/Documenter/HmOsT/src/Documenter.jl:499
```

Nothing is awesome @kleinschmidt @Nosferican

kleinschmidt:
Try rebasing on master?

Nosferican:
If not, we (as in @kleinschmidt) could regenerate the DocumenterTools SSH keys.

palday:
Still fails after rebase :(

palday:
But works on #265 ...

kleinschmidt:
For some reason the `DOCUMENTER_KEY` secret exists but is empty on #264 :
```
Run julia-actions/julia-docdeploy@releases/v1
  env:
    GITHUB_TOKEN: ***
    DOCUMENTER_KEY: 
```

(contrast this with #265 , where the key has a non-empty value:
```
Run julia-actions/julia-docdeploy@releases/v1
  env:
    GITHUB_TOKEN: ***
    DOCUMENTER_KEY: ***
```

I have no idea why there would be a difference

kleinschmidt:
Ahhhhhh I see now "Secrets are not passed to workflows that are triggered by a pull request from a fork."

kleinschmidt:
The exception (according to the [docs](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets)) is the GITHUB_TOKEN secret, so maybe we should use that for documenter?  Or I guess not, since it doesn't trigger gh-pages builds...

https://github.com/JuliaDocs/Documenter.jl/issues/1177#issuecomment-586715840

Nosferican:
That should be fine... The docs will update when external PR is merged (triggered on commit to master).  If we wanted to preview the docs after making sure the PR is kosher we can trigger it ourselves and that should pass the secret.

PR #261, opened 2020-02-13
update docs, fix copy-paste error for parametricbootstrap methods

palday:
Include `use_threads` in docstring and fix a copy-paste error for the implict RNG method.


palday:
I would wait to merge until the #258 lands and we can easily check how the docs build.

PR #264, opened 2020-02-15
new rePCA: now on correlations and with optional extra info

palday:
Questions:

- [x] How does the `show` method for `PCA` with all of its extra parameters look? Should some of those be renamed?
- [x] Related to that: how are the properties for that?
- [x] Should we have a function that returns a NamedTuple of the full PCA struct for each grouping variable?
- [x] Related to that: how is the API for `rePCA`? I'm not sure how I feel about `loadings`
- [x] Related to that: should `corr` be a named argument?

@dmbates Note that this has been rebased onto current `#master`, which is why it's in my fork, but if you're mostly okay with the changes, I'll force push onto the main repo's `correlationPCA` branch.

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -620,7 +620,7 @@ end
 LinearAlgebra.rank(m::LinearMixedModel) = first(m.feterms).rank
 
 """
-    rePCA(m::LinearMixedModel; loadings::Bool=false, corr::Bool=true)
+    rePCA(m::LinearMixedModel; corr::Bool=true)
 
 Return a named tuple of the normalized cumulative variance of a principal components
 analysis of the random effects covariance matrices or correlation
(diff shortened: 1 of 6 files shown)


palday:
I think the ideal thing would be to have the property `rePCA` only show the cumulative variances, but give the function `rePCA` an option `cumvar_only` with `true`. For `cumvary_only=false`, return a named tuple of `PCA` objects.

kleinschmidt:
The reason that the doc preview is failing is because github does not pass on secrets to workflows triggered by PRs from forks (since that would allow anyone who forks to expose secrets).  So we can't see the preview docs build but this won't affect the docs deploy once this is merged.

palday:
@kleinschmidt Then I'll make sure to use the main repo for anything with a lot of doc changes ...

src/linearmixedmodel.jl
```diff
@@ -638,11 +641,31 @@ end
...
+principal component, the first two principal components, etc.  The last element is
+always 1.0 representing the complete proportion of the variance.
+"""
+function rePCA(m::LinearMixedModel; corr::Bool=true)
+    pca = PCA.(m.reterms, corr=corr)
+    NamedTuple{fnames(m)}(getproperty.(pca,:cumvar))
+end
+
+"""
+    rePCA(m::LinearMixedModel; corr::Bool=true)
```
dmbates: Should be just `PCA`, not `rePCA`.

kliegl:
Wow!

PR #265, opened 2020-02-16
LikelihoodRatioTest struct and pretty-printing via coeftable

palday:
Sample session:

```julia
using MixedModels
slp = MixedModels.dataset(:sleepstudy);
fm0 = fit(MixedModel,@formula(reaction ~ 1 + (1+days|subj)),slp);
fm1 = fit(MixedModel,@formula(reaction ~ 1 + days + (1+days|subj)),slp);
```

```julia
julia> lrt = MixedModels.likelihoodratiotest(fm0,fm1)
Model Formulae
1: reaction ~ 1 + (1 + days | subj)
2: reaction ~ 1 + days + (1 + days | subj)
─────────────────────────────────────────────
   model-dof  deviance     χ²  χ²-dof  P(>χ²)
─────────────────────────────────────────────
1        5.0   1775.48  -0.00     0.0  <1e-99
2        6.0   1751.94  23.54     1.0  <1e-5 
─────────────────────────────────────────────
```


src/likelihoodratiotest.jl
```diff
@@ -0,0 +1,102 @@
...
+    if s == :dof
+        lrt.models.dof
+    elseif s == :deviance
+        lrt.models.deviance
+    elseif s == :pvalues
+        lrt.tests.pvalues
+    elseif s == :formulae
+        lrt.formulas
+    else
+        getfield(lrt, s)
```
Nosferican: This should probably be `getproperty`.

Nosferican:
Do we want to have any checks on valid LR tests? For example, if REML and the fixed effects differ or something. Maybe as a `@warn`?

palday:
Maybe for REML general, but @dmbates has a warning in the doc string that the validity of LR tests is in some sense the responsibility of the user

src/likelihoodratiotest.jl
palday: It has to be `getfield` or it's a bottomless recursion.

palday: (I found this out the hard way.)

palday:
@Nosferican I've got the warning in there now, or more precisely an argument error when the coefnames for REML models aren't identical. 

@dmbates We can change the printing pretty easily at a later date, but does the `LikelihoodRatioTest` struct and accessors make sense? 

One thing this won't catch -- and shouldn't! -- is when the users have done something ridiculous and are comparing models fit to different data. In `lme4` this can arise from the interaction of NAs, even for models fit to the same dataframe, but because we currently do a very coarse, very early drop of rows with missings, that subtle case shouldn't arise here. Until we have better missing support. But that's a problem for future us, I think.

src/likelihoodratiotest.jl
Nosferican: Yeah. I realized that as I was walking to the zoo... all the sudden, wait it's overloading `getproperty`... that would be bad. #StrollingThroughTheForestRealizations

Issue #206, opened 2019-10-01
Deviance for families with dispersion parameter


palday:
Currently, the dispersion parameter for families with one isn't taken into account when calculating the deviance for GLMMs. For example, in Julia we have:

```julia
julia> lexdec_glmm_gamma
Generalized Linear Mixed Model fit by maximum likelihood (nAGQ = 1)
  rt_raw ~ 1 + Class + NativeLanguage + Class & NativeLanguage + (1 | Subject) + (1 | Word)
  Distribution: Gamma{Float64}
  Link: IdentityLink()

  Deviance: 91.9052

Variance components:
             Column    Variance Std.Dev. 
 Word    (Intercept)     0.0000  0.00000
 Subject (Intercept)  1508.7803 38.84302
 Number of obs: 1659; levels of grouping factors: 79, 21

Fixed-effects parameters:
──────────────────────────────────────────────────────────────────────────────
                                       Estimate  Std.Error    z value  P(>|z|)
──────────────────────────────────────────────────────────────────────────────
(Intercept)                           563.552      6.69785  84.1393     <1e-99
Class: plant                            6.43673    9.20547   0.699229   0.4844
NativeLanguage: Other                 114.543     11.2587   10.1737     <1e-23
Class: plant & NativeLanguage: Other  -29.3668    15.495    -1.89524    0.0581
──────────────────────────────────────────────────────────────────────────────

julia> loglikelihood(lexdec_glmm_gamma)
ERROR: MethodError: no method matching Gamma{Float64}(::Float64)
Closest candidates are:
  Gamma{Float64}(::Any, ::Any) where T at /home/XXX/.julia/packages/Distributions/Vcqls/src/univariate/continuous/gamma.jl:30
Stacktrace:
 [1] loglikelihood(::GeneralizedLinearMixedModel{Float64}) at /home/XXX/Work/MixedModels.jl/src/generalizedlinearmixedmodel.jl:301
 [2] top-level scope at none:0
```

But in R we have:

```R
R> summary(lexdec_glmm_gamma)
Generalized linear mixed model fit by maximum likelihood (Laplace
  Approximation) [glmerMod]
 Family: Gamma  ( identity )
Formula: rt_raw ~ 1 + Class * NativeLanguage + (1 | Subject) + (1 | Word)
   Data: lexdec

     AIC      BIC   logLik deviance df.resid 
 20339.4  20377.3 -10162.7  20325.4     1652 

Scaled residuals: 
[...]

PR #270, opened 2020-02-16
GLMM dispersion and dof_residual

### LINKED ISSUE #248: Gaussian  GLMMs with non identity link
I've found what breaks it:

```julia

julia> GeneralizedLinearMixedModel(@formula(yield ~ 1 + (1|batch)), data(:dyestuff),
                                Normal(),SqrtLink())
Generalized Linear Mixed Model fit by maximum likelihood (nAGQ = 1)
  yield ~ 1 + (1 | batch)
  Distribution: Normal{Float64}
  Link: SqrtLink()

  Deviance: 115249.4629

Error showing value of type GeneralizedLinearMixedModel{Float64}:
ERROR: dof_residual is not defined for GeneralizedLinearMixedModel{Float64}.
Stacktrace:
 [1] error(::String) at /usr/local/lib/julia/sys.so:?
 [2] dof_residual(::GeneralizedLinearMixedModel{Float64}) at /home/XXX/.julia/packages/StatsBase/DyWPR/src/statmodels.jl:350
 [3] dispersion(::GeneralizedLinearMixedModel{Float64}, ::Bool) at /home/XXX/Code/MixedModels.jl/src/generalizedlinearmixedmodel.jl:108 (repeats 2 times)
```

This isn't quite the same, but it's related to broader issues with GLMMs and families with a dispersion parameter (see e.g. #206).

palday:
This adds/fixes methods for  `GLM.dispersion` and `StatsBase.dof_residual` for `GeneralizedLinearMixedModel` (closes #248 and potentially #206).

Questions from late-night coding confusion:
- [x] The dispersion is just calculated as it would be in a GLM. Is this correct? 
- [x]  Residual degrees of freedom is done using the naive formula (n observation - model degrees of freedom). Is this correct?
- [x] If that is correct, then I'm not sure that `dof_residual(::LinearMixedModel)` is correct. It delivers the right answer for (its use in) the evaluation of the objective, but I don't think that the residual degree of freedom should match the number of observations nor be dependent on the fitting criterion (ML vs. REML). [`lme4` just uses the (nobs - model.df) formula](https://github.com/lme4/lme4/blob/bdf54de8298ecd505173801076ff00623e95d5eb/R/lmer.R#L1047). 

Here some samples that make me doubt myself.

# Gaussian with Non Identity Link

```julia
julia> using MixedModels, RCall
julia> glmm = GeneralizedLinearMixedModel(@formula(yield ~ 1 + (1|batch)), MixedModels.dataset(:dyestuff),Normal(),SqrtLink())
Generalized Linear Mixed Model fit by maximum likelihood (nAGQ = 1)
  yield ~ 1 + (1 | batch)
  Distribution: Normal{Float64}
  Link: SqrtLink()

  Deviance: 115249.4629

Variance components:
            Column    Variance  Std.Dev. 
batch    (Intercept)  1961.0615 44.283874
Residual              4266.2037 65.316183

Fixed-effects parameters:
             Estimate  Std.Error  z value  P(>|z|)
(Intercept)   39.0832    18.0791     2.16   0.0306
R> glmer(Yield ~ 1 + (1|Batch), data=Dyestuff, family=gaussian(link="sqrt"))
Generalized linear mixed model fit by maximum likelihood (Laplace
  Approximation) [glmerMod]
 Family: gaussian  ( sqrt )
Formula: Yield ~ 1 + (1 | Batch)
   Data: Dyestuff
      AIC       BIC    logLik  deviance  df.resid 
 381.4231  385.6267 -187.7115  375.4231        27 
Random effects:
 Groups   Name        Std.Dev.
 Batch    (Intercept) 24.49   
 Residual             44.29   
Number of obs: 30, groups:  Batch, 6
Fixed Effects:
(Intercept)  
      39.08 
```
(Note that we can't compare the deviance here because of #206 and lme4/lme4#375.)

It's a bit odd that the residual std. dev. in the lme4 fit matches the `batch` std. dev. in the MixedModels fit.

# Gamma model

I was curious whether this was a pecularity of Gaussians with non-identity link, so I tried a dumb model:
```julia
julia> kb07 = MixedModels.dataset(:kb07)
julia> gamma = fit(MixedModel,@formula(rt_raw ~ 1 + spkr*prec*load + (1+spkr+prec+load|subj)), kb07, Gamma(), LogLink())
Generalized Linear Mixed Model fit by maximum likelihood (nAGQ = 1)
  rt_raw ~ 1 + spkr + prec + load + spkr & prec + spkr & load + prec & load + spkr & prec & load + (1 + spkr + prec + load | subj)
  Distribution: Gamma{Float64}
  Link: LogLink()

  Deviance: 655.4913

Variance components:
             Column       Variance     Std.Dev.    Corr.
subj     (Intercept)     0.1382712720 0.371848453
         spkr: old       0.1382715354 0.371848807 -0.00
         prec: maintain  0.1382715355 0.371848807 -0.00  0.00
         load: yes       0.0013793501 0.037139603 -0.00  0.00  0.00
Residual                 0.1371768891 0.370373985
 Number of obs: 1789; levels of grouping factors: 56

Fixed-effects parameters:
────────────────────────────────────────────────────────────────────────────────
                                           Estimate  Std.Error  z value  P(>|z|)
────────────────────────────────────────────────────────────────────────────────
(Intercept)                              7.7793      0.0555826   139.96   <1e-99
spkr: old                                0.068266    0.0608828     1.12   0.2622
prec: maintain                          -0.301492    0.0608828    -4.95   <1e-6 
load: yes                                0.0877345   0.0355276     2.47   0.0135
spkr: old & prec: maintain              -0.0682468   0.0497819    -1.37   0.1704
spkr: old & load: yes                    0.00985952  0.0497207     0.20   0.8428
prec: maintain & load: yes              -0.0423475   0.0497207    -0.85   0.3944
spkr: old & prec: maintain & load: yes   0.0765511   0.0703376     1.09   0.2764
────────────────────────────────────────────────────────────────────────────────

julia> @rput kb07;
# this next step takes a while ....
R> glmer(rt_raw ~ 1 + spkr*prec*load + (1+spkr+prec+load|subj), kb07, family=Gamma(link="log"))
┌ Warning: RCall.jl: Warning in checkConv(attr(opt, "derivs"), opt$par, ctrl = control$checkConv,  :
│   Model failed to converge with max|grad| = 0.00332188 (tol = 0.001, component 1)
└ @ RCall ~/.julia/packages/RCall/g7dhB/src/io.jl:113
Generalized linear mixed model fit by maximum likelihood (Laplace
  Approximation) [glmerMod]
 Family: Gamma  ( log )
Formula: rt_raw ~ 1 + spkr * prec * load + (1 + spkr + prec + load | subj)
   Data: kb07
      AIC       BIC    logLik  deviance  df.resid 
 28720.41  28824.71 -14341.20  28682.41      1770 
Random effects:
 Groups   Name         Std.Dev. Corr             
 subj     (Intercept)  0.1744                    
          spkrold      0.1203   -0.36            
          precmaintain 0.1327   -0.61  0.13      
          loadyes      0.1141   -0.44  0.26  0.08
 Residual              0.3916                    
Number of obs: 1789, groups:  subj, 56
Fixed Effects:
                 (Intercept)                       spkrold  
                     7.74677                       0.07344  
                precmaintain                       loadyes  
                    -0.29274                       0.08844  
        spkrold:precmaintain               spkrold:loadyes  
                    -0.07189                      -0.00706  
        precmaintain:loadyes  spkrold:precmaintain:loadyes  
                    -0.03788                       0.08865  
convergence code 0; 1 optimizer warnings; 0 lme4 warnings 
```
Somewhat closer, despite the convergence warning. So maybe it is something about the Gaussian case.

### DIFF
--- a/src/generalizedlinearmixedmodel.jl
+++ b/src/generalizedlinearmixedmodel.jl
@@ -112,13 +112,27 @@ function deviance!(m::GeneralizedLinearMixedModel, nAGQ = 1)
     deviance(m, nAGQ)
 end
 
-GLM.dispersion(m::GeneralizedLinearMixedModel, sqr::Bool = false) =
-    dispersion(m.resp, dof_residual(m), sqr)
+function GLM.dispersion(m::GeneralizedLinearMixedModel, sqr::Bool = false)
+# adapted from GLM.dispersion(::AbstractGLM, ::Bool)
+# TODO: PR for a GLM.dispersion(resp::GLM.GlmResp, dof_residual::Int, sqr::Bool)
+    r = m.resp
+    if dispersion_parameter(r.d)
+        wrkwt, wrkresid = r.wrkwt, r.wrkresid
+        s = sum(i -> wrkwt[i] * abs2(wrkresid[i]), eachindex(wrkwt, wrkresid)) / dof_residual(m)
+        sqr ? s : sqrt(s)
+    else
+        one(eltype(r.mu))
+    end
+end
 
 GLM.dispersion_parameter(m::GeneralizedLinearMixedModel) = dispersion_parameter(m.resp.d)
 
-function StatsBase.dof(m::GeneralizedLinearMixedModel)
+function StatsBase.dof(m::GeneralizedLinearMixedModel)::Int
     length(m.β) + length(m.θ) + GLM.dispersion_parameter(m.resp.d)
+end
+
+function StatsBase.dof_residual(m::GeneralizedLinearMixedModel)::Int
+    nobs(m) - dof(m)
 end
 
 fit(
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -287,8 +287,12 @@ GLM.dispersion_parameter(m::LinearMixedModel) = true
 StatsBase.dof(m::LinearMixedModel) = size(m)[2] + nθ(m) + 1
 
 function StatsBase.dof_residual(m::LinearMixedModel)::Int
-    (n, p, q, k) = size(m)
-    n - m.optsum.REML * p
+    # nobs - rank(FE) - 1 (dispersion)
+    # this differs from lme4 by not including nθ
+    # a better estimate would be a number somewhere between the number of
+    # variance components and the number of conditional modes
+    # nobs, rank FE, num conditional modes, num grouping vars
+    nobs(m) - size(m)[2] - 1
 end
 
 """
@@ -528,7 +532,7 @@ Return negative twice the log-likelihood of model `m`
 """
 function objective(m::LinearMixedModel)
     wts = m.sqrtwts
-    logdet(m) + dof_residual(m) * (1 + log2π + log(varest(m))) -
+    logdet(m) + ssqdenom(m) * (1 + log2π + log(varest(m))) -
     (isempty(wts) ? 0 : 2 * sum(log, wts))
 end
 
@@ -778,8 +782,16 @@ function σρs(m::LinearMixedModel)
     NamedTuple{fnames(m)}(((σρs(t, σ) for t in m.reterms)...,))
 end
 
+"""
+    size(m::LinearMixedModel)
+
+Returns the size of a mixed model as a tuple of length four:
+the number of observations, the number of (non-singular) fixed-effects parameters,
+the number of conditional modes (random effects), the number of grouping variables
+"""
 function Base.size(m::LinearMixedModel)
-    n, p = size(first(m.feterms))
+    n = size(first(m.feterms),1)
+    p = first(m.feterms).rank
     n, p, sum(size.(m.reterms, 2)), length(m.reterms)
 end
 
@@ -791,6 +803,20 @@ Return the square root of the penalized, weighted residual sum-of-squares (pwrss
 This value is the contents of the `1 × 1` bottom right block of `m.L`
 """
 sqrtpwrss(m::LinearMixedModel) = first(m.L.blocks[end, end])
+
+"""
+    ssqdenom(m::LinearMixedModel)
+
+Return the denominator for penalized sums-of-squares.
+
+For MLE, this value is either the number of observation for ML. For REML, this
+value is the number of observations minus the rank of the fixed-effects matrix.
+The difference is analagous to the use of n or n-1 in the denominator when
+calculating the variance.
+"""
+function ssqdenom(m::LinearMixedModel)::Int
+    nobs(m) - m.optsum.REML * first(m.feterms).rank
+end
 
 """
     std(m::MixedModel)
@@ -898,7 +924,7 @@ end
 
 Returns the estimate of σ², the variance of the conditional distribution of Y given B.
 """
-varest(m::LinearMixedModel) = pwrss(m) / dof_residual(m)
+varest(m::LinearMixedModel) = pwrss(m) / ssqdenom(m)
 
 """
     vcov(m::LinearMixedModel)
--- a/test/fit.jl
+++ b/test/fit.jl
@@ -21,7 +21,7 @@ end
 end
 
 @testset "Normal Distribution GLMM" begin
-    @test_broken(isa(fit(MixedModel, @formula(yield ~ 1 + (1|batch)), MixedModels.dataset(:dyestuff),
-                         Normal(), LogLink),
-                     GeneralizedLinearMixedModel))
+    @test isa(fit(MixedModel, @formula(yield ~ 1 + (1|batch)), MixedModels.dataset(:dyestuff),
+                         Normal(), SqrtLink()),
+                     GeneralizedLinearMixedModel)
 end


palday:
@dmbates I think this is ready for a critical eye. :)

dmbates:
Sorry, I have lost track of the plot here.  This appears ready to squash and merge.  Anything else you want to add to it @palday ?

palday:
I think the code here is ready to merge, but there were two lingering questions:
1. Is `dof_residual` correct for LMM? I think it's fine as is, but it does give different answers than lme4. Rewriting to match lme4's behavior would require changing some things in the objective. 
2. What's happening with the Gaussian fit with non identity link (and potentially other families with dispersion parameters)?

I think the answer to (1) is "it's less work to leave as is and if the ratio of observations to parameters is so close as to make a difference, then you really should just collect more data" with a potentially future follow-up using some of the ddf ideas we've discussed with figuring out a good effective degrees of freedom that compromises between the  number of conditional modes vs. the number of variance parameters.

For (2), I think we should probably raise it as a separate issue and see if we can find some good test data for Gamma, inverse Gaussian and Gaussian with non-identity link.

dmbates:
Let's go with `dof_residual` for `LinearMixedModel` returning `n - rank(fetrm(m))`.  It will be necessary to add another internal function like `ssqdenom` to distinguish between ML and REML estimation.  Are you okay with doing this, @palday ?

palday:
Yep, I'll take care of that.

palday:
While I've been adding in that latest change, two questions arose:

1. The tuple returned by size includes singular columns for the fixed effects. Should this return the rank instead? (If not, then I need to make a quick change to `dof` to reflect the smaller effective model matrix for the rank deficient case.)
2. For `dof_residual`, I'm almost tending to nobs - dof_model as the very naive formula, which would match lme4 behavior. Otherwise, I think I would prefer nobs - rank(FE) - 1 to match fixed-effects regression. (This is the gift that keeps giving ....)

dmbates:
I think you are correct that size should return the rank not the number of columns in the fixed-effects model matrix.

Gotta love those degrees of freedom questions - it's always a case of "Which of these wrong answers should we return?"  Let's go ahead and use nobs - rank(FE) - 1.  It's an upper bound but, I suspect, not too far off the mark.  I will create a long-term projects issue and add "evaluate the trace of the hat matrix" as I think that the best answer will be nobs - tr(H).

palday:
@dmbates If you're happy with all the current wrong answers for `dof_residual` :) then this is ready for squash and merge.

Nosferican:
Should I open an issue for the approximation to the effective degrees of freedom à la [Satterthwaite](https://en.wikipedia.org/wiki/Welch%E2%80%93Satterthwaite_equation)?

dmbates:
@palday @Nosferican Does anyone know of a computationally feasible formula for Satterthwaite or Kenward-Roger?  I am not opposed to including them - it's just that I don't want them included with naive formulas that involve the inversion of n by n matrices.

palday:
I still don't know of a good one for K-R -- and I've read the source of `pbkrtest`. Satterthwaite is fairly straightforward, if I recall correctly (I also read the source of `lmerTest` a while back) -- on the order of the current `leverage` implementation.

Nosferican:
If not mistaken, Satterthwaite = Kenward-Roger for the approximated degrees of freedom formula which is given in the Wikipedia link (no matrix inversion required) as far as I can tell.

I believe the part that requires inverting the n by n matrix is the Kenward-Roger estimate for standard errors.

palday:
Ah, that might be the case. But I'm with @dmbates on this: if you're trying to invert a matrix in anything but theorem proving, then you should probably re-consider what life choices brought you to this point ...

Nosferican:
Tabling the standard error estimates aside (i.e., the ones with the naive inefficient computations). For the estimates of the degree of freedom, we should probably incorporate (Satterthwaite = Kenward-Roger) methods since the formula for that seems pretty efficient (e.g., doesn't require any matrix inversion).

palday:
Can you go ahead and open an issue with the formula in the issue body?

Nosferican:
@dmbates @palday Could y'all take a look at [Kenward Roger (2009)](https://doi.org/10.1016/j.csda.2008.12.013).
> From a computational point of view, matrices such as `X'SX` are typically comparatively small and can be calculated efficiently taking advantage of the block diagonal structure of, and its derivatives. In practice the values of Wij, the variances and covariances of the estimated covariance parameters σ, can be estimated via the information matrix in the conventional manner. However it should be noted that although the observed, expected, or average information matrices can be used to estimate these quantities in the part of the adjustment taken from the original Kenwardand Roger(1997) paper, this is not true for new additional component(3). For this, the inner values of Wij should be estimated using the expected information, while for the outer values, Wst, any of the three alternatives can be used.

It seems the 2009 article gives a revised estimator which is computationally sensible.

PR #272, opened 2020-02-17
add docstring for MixedModel supertype that points to concrete terms

kleinschmidt:
People often look for help on `MixedModel`.  This adds a pointer to
`LinearMixedModel` and `GeneralizedLinearMixedModel` in the docstring.

### DIFF
--- a/src/MixedModels.jl
+++ b/src/MixedModels.jl
@@ -109,6 +109,13 @@ export @formula,
 
 import Base: ==, *
 
+"""
+    MixedModel
+
+Abstract type for mixed models.  MixedModels.jl implements two subtypes:
+`LinearMixedModel` and `GeneralizedLinearMixedModel`.  See the documentation for
+each for more details.
+"""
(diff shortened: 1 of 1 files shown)


palday:
I'm going to take this opportunity to ask about whether we should create a non-release Git tag for the state of the repo at 19:00 Bielefeld time yesterday. Something like `zif-tutorial2020`, so that we can always point to that version, even as we continue adding stuff in. Would that irritate TagBot?

palday:
@kleinschmidt Can you extend the doc a little bit to include the note that's primarily used for the `fit` method without an explicit constructor? I think the easiest/best solution would be to move `fit(::GeneralizedLinearMixedModel)` to `fit(::MixedModel)` because that's what the superclass is actually used for by the user, and we don't have the comparable doc for `fit(::LinearMixedModel)`.

kleinschmidt:
WIll do.  I think tagbot ignores tags that it doesn't control and that don't conflict so there should be no problem.

PR #277, opened 2020-02-18
add corrmat function that generates full correlation matrix of ReMat

kleinschmidt:
This makes it more convenient to work with correlation matrices from fitted
models.  I've added a doctest to show that it's equivalent to `VarrCor` at least
for one model fit to the KB07 data, but probably should have more extensive
tests.


palday:
@kleinschmidt Can we get codecov to detect doctests / is there a way to run the doctests as part of `test`?

PR #279, opened 2020-02-19
LRT for GLMMs

### LINKED ISSUE #278: MixedModels.likelihoodratiotest for glmer models
currently only works for linear models :)

palday:
Closes #278,


src/likelihoodratiotest.jl
```diff
@@ -86,21 +103,61 @@ function Base.show(io::IO, lrt::LikelihoodRatioTest; digits=2)
...
+    outrows[2, :] = ["[1]",
+                    @sprintf("%.0d", lrt.dof[1]),
+                    @sprintf("%.4f", lrt.deviance[1]),
+                    " "," ", " "]
+
+    for i in 2:nr
+        outrows[i+1, :] = ["[$i]",
+                           @sprintf("%.0d", lrt.dof[i]),
+                           @sprintf("%.4f", lrt.deviance[i]),
+                           @sprintf("%.0d", Δdf[i-1]),
```
dmbates: I get
```julia
julia> MixedModels.likelihoodratiotest(fm2, fm1)
Model Formulae
1: reaction ~ 1 + days + MixedModels.ZeroCorr((1 + days | subj))
2: reaction ~ 1 + days + (1 + days | subj)
Error showing value of type MixedModels.LikelihoodRatioTest:
ERROR: UndefVarError: Δdf not defined
```
with the stack trace pointing here.

dmbates: It appears that you accidentally incorporate the line to define `Δdf` in a comment line above.  Instead of guessing what might have occurred, I'll let you undo that it it was indeed the problem.

Other than that, it looks good to go from me - well we might want to ponder why this wasn't caught in a test.

palday: Oh, that looks like I hit delete right before doing git add. Grrr. I'm also surprised this wasn't caught by a test ... I might have forgotten to test the `show` method itself. I'll investigate.

palday: Ah, because I didn't include the file with the new tests in `runtests.jl`. That's twice in the past two weeks ....

palday: And what do you know, when you actually enable the tests, coverage improves....

dmbates:
The column labels are backwards
```julia
julia> MixedModels.likelihoodratiotest(m2, m1)
Model Formulae
1: reaction ~ 1 + days + MixedModels.ZeroCorr((1 + days | subj))
2: reaction ~ 1 + days + (1 + days | subj)
─────────────────────────────────────────────
     model-dof   deviance  χ²  χ²-dof  P(>χ²)
─────────────────────────────────────────────
[1]          5  1752.0033                    
[2]          6  1751.9393   1  0.0639  0.8004
─────────────────────────────────────────────
```
I'll push a change, if that is okay with you.

dmbates:
Should `likelihoodratiotest` be exported?  Would it be better to change the name to `lrtest`?

palday:
I have mixed feelings on that. I would hesitate to consider the name `lrtest` until the associated StatsModel pull request (JuliaStats/StatsModels.jl#162) lands. The `lrtest` there is defined generically enough that it should work on `MixedModels` without difficulty, but it won't do a lot of the checking we do (all REML or ML, all same link and same family, same FE for REML).  My suggestion would be to include `likelihoodratiotest` but not export it and keep the name to mark it as being potentially different -- right now, the internal representations certainly are. 

If we want immediate compatibility, we can include their definition of the  `LRTestResult` struct and associated `show` method and remove that when `lrtest` makes it into a StatsModels release.  When the `lrtest` PR lands, we can add a method for `MixedModel`  that does our extra checks and then delegates to the more general method and then deprecate `likelihoodratiotest` and/or have it just call `lrtest`. 

Let me know which route you want to go. I think the advantage to having `likelihoodratiotest` right now is that we can include it in our next release without waiting for StatsModels and/or depending on a higher StatsModels version. Not exporting it highlights that it's our solution and not the general JuliaStats ecosystem solution. I can still make it as compatible as possible with the 'lrtest` proposal on StatsModels.

dmbates:
Thanks for reviewing the state of play for me.  I agree with keeping the name `likelihoodratiotest` but not export it yet.

dmbates:
This is good to go from my point of view.  Feel free to pull the trigger if you agree.

nalimilan:
> The `lrtest` there is defined generically enough that it should work on `MixedModels` without difficulty, but it won't do a lot of the checking we do (all REML or ML, all same link and same family, same FE for REML).

Have you noticed the `isnested` function that I also introduced at https://github.com/JuliaStats/StatsModels.jl/pull/162? My idea was that MixedModels would implement it to run any necessary checks.

palday:
@nalimilan At first, I thought `isnested` is the wrong name for all of the checks we do because we're not checking nesting per se, but rather general  comparability. Then it occurred to me that if the models don't have the same family, etc., then they definitely aren't nested. In some sense, REML models with different fixed effects aren't nested in terms of their optimization criteria, but they are nested in terms of their formulae. So I don't know if `isnested` is a misnomer there. Of course, it's no more of a `misnomer` than `likelihoodratiotest` on something that isn't a - or as @dmbates likes to emphasizes, *the* - likelihood....

When the StatsModels PR lands, I'll implement the necessary methods. Maybe sooner, if I get bored. :)

nalimilan:
OK. I guess we can adopt any definition of "nested" we want (there could also be arguments to choose a stricter definition). Let me know if you prefer another term. Strictly speaking, what we need could be called `iscomparableforlikelihoodratiotest`. :-)

PR #283, opened 2020-02-20
Up Docs

Nosferican:
This is a draft for the first update page.
- [x] `constructors`
- [x] `optimization`
- [x] `GaussHermite`
- [x] `bootstrap`


docs/jmd/constructors.jmd
```diff
@@ -107,38 +162,38 @@ zerocorr!
...
 The special syntax `zerocorr` can be applied to individual random effects terms inside the `@formula`:
-```@docs
-zerocorr
+```@example Main
+fm2zerocorr_fm = fit(MixedModel, @formula(reaction ~ 1 + days + zerocorr(1 + days|subj)), sleepstudy)
 ```
-```{julia;term=true}
-fit(MixedModel, @formula(reaction ~ 1 + days + zerocorr(1+days|subj)), sleepstudy)
+```@setup Main
+    all(fm2zerocorr == fm2zerocorr_fm
```
dmbates: Should there be a closing parenthesis here?

docs/jmd/constructors.jmd
```diff
@@ -167,30 +222,30 @@ In this form optional arguments `fast` and/or `nAGQ` can be passed to the optimi
...
+deviances = Dict{Symbol,Float64}()
+b = @benchmarkable deviances[:default] = deviance(fit(MixedModel, $contraform, $contraception, $bernoulli));
+run(b)
+b = @benchmarkable deviances[:fast] = deviance(fit(MixedModel, $contraform, $contraception, $bernoulli, fast = true));
+run(b)
+b = @benchmarkable deviances[:nAGQ] = deviance(fit(MixedModel, $contraform, $contraception, $bernoulli, nAGQ=9));
+run(b)
+b = @benchmarkable deviances[:nAGQ_fast] = deviance(fit(MixedModel, $contraform, $contraception, $bernoulli, nAGQ=9, fast=true));
+run(b)
+sort(deviances)
```
dmbates: I don't think this sort is doing what you intended.  It seems to sort on the keys, not on the values.
```julia
julia> sort(deviances)
OrderedCollections.OrderedDict{Symbol,Float64} with 4 entries:
  :default   => 2372.73
  :fast      => 2372.78
  :nAGQ      => 2372.46
  :nAGQ_fast => 2372.51
```

docs/jmd/constructors.jmd
```diff
@@ -248,49 +315,63 @@ The `coef` and `fixef` extractors both return the maximum likelihood estimates o
...
-show(fixef(fm1))
-show(fixef(gm1))
+```@example Main
+coef(fm1)
+```
+```@example Main
+fixef(fm1)
+```
+```@example Main
+fixef(fm1)
```
dmbates: Did you intend to repeat this chunk?

docs/jmd/constructors.jmd
Nosferican: I intended on doing on keys as those as I ran them in that order but I probably want to sort them by performance now that I think about it.

docs/jmd/constructors.jmd
Nosferican: An eager push... I was going to be testing the model with `==` on every property absent a 
`==(::LinearMixedModel, ::LinearMixedModel)`

dmbates:
@Nosferican Are you planning to make similar changes in the other .jmd files or should we merge this commit as is?

palday:
I think the plan was to update some additional pages, but to go ahead and get some review as he finished each one.

dmbates:
@palday and @Nosferican can you take a look.  I think this is ready to squash and merge.

Nosferican:
Will be reviewing it today.

palday:
In the [preview](https://juliastats.org/MixedModels.jl/previews/PR283/constructors/#Examples-of-linear-mixed-effects-model-fits-1), the doc tests are shown (without any results). Is this desired?

palday:
For ` vcov(m::LinearMixedModel)`, it looks like I forgot to update the prototype in the docstring. Can you integrate that into this PR?

palday:
I really should not be trusted to update doc strings.

Nosferican:
We can have the doctest show the results... I would say that should be the preferred way

dmbates:
I think if we just have a `@testset` we should put it in `@setup Main`, not `@example Main`.  Right now we are inconsistent about that.  I meant to change the ones that are currently in `@example Main`.

Nosferican:
For `doctest` yes... Those are docstrings with `jldoctest`.
For the testing we perform in the docs (not in the docstring),
We should hide the code and output with `@setup Main`.

dmbates:
I haven't tried using `@setup Main` for the plots' code.  I like to put explanatory axis labels and titles on plots so that they can be comprehended without needing to read the surrounding text.  But this makes  all the `Guide.xlabel(...)`, etc. calls in the plot call itself redundant for the reader.  I would like to hide the call to `plot(...)` but still have the plot itself appear.  I may try to see if I can get that to work on the train tomorrow.

Nosferican:
You could use `@example Main` and use `# hide` to hide the source code but still have the output show up https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@example-block-1

.gitignore
```diff
@@ -1,8 +1,8 @@
 docs/build
 docs/site
-docs/src/*.md
 docs/src/assets/*.svg
 test/scratch.jl
 tune.json
 Manifest.toml
 settings.json
+docs/jmd
```
Nosferican: Why is this?

docs/src/GaussHermite.md
```diff
@@ -29,41 +29,45 @@ More formally, a `k`th order rule is exact when `f` is a `k-1` order polynomial.
...
 show(abs2.(ev.vectors[1,:]))
 ```
 As a function of `k` this can be written as
-```{julia;results="hidden"}
+```@example Main
 function gausshermitenorm(k)
     ev = eigen(SymTridiagonal(zeros(k), sqrt.(1:k-1)))
     ev.values, abs2.(ev.vectors[1,:])
-end
+end;
```
Nosferican: Is the `;` needed here?

.gitignore
dmbates: The idea is to eliminate the jmd directory altogether and just use an src directory now that the weaving is no longer done.

docs/src/GaussHermite.md
dmbates: Not really.  I had hoped that I could suppress the output with the `;` but no such luck.

dmbates:
@palday, @Nosferican Shall we go ahead and merge this now so that other PRs can be rebased to use the cleaner docs generation?

PR #287, opened 2020-02-23
Add stderror! method, distinguish coef and fixef

dmbates:
The docs should be updated to indicate that `coef` always refers to the full coefficient vector, as does `stderror` and `fixef` refers to the truncated coefficient vector.  Furthermore this extends to associated functions like `coefnames` and `fixefnames`.  However, I think it is better to do this in the version from #283

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -382,7 +382,7 @@ StatsBase.fitted(m::LinearMixedModel{T}) where {T} = fitted!(Vector{T}(undef, no
 
 Overwrite `v` with the pivoted fixed-effects coefficients of model `m`
 
-For full rank models the length of `v` must be the rank of `X`.  For rank-deficient models
+For full-rank models the length of `v` must be the rank of `X`.  For rank-deficient models
 the length of `v` can be the rank of `X` or the number of columns of `X`.  In the latter
 case the calculated coefficients are padded with -0.0 out to the number of columns.
 """
@@ -824,6 +824,10 @@ function stderror!(v::AbstractVector{T}, m::LinearMixedModel{T}) where {T}
     end
     invpermute!(v, fetrm(m).piv)
     v
+end
+
+function StatsBase.stderror(m::LinearMixedModel{T}) where {T}
+    stderror!(similar(fetrm(m).piv, T), m)
 end
 
 """
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -159,15 +159,21 @@ fit(
     REML = REML,
 )
 
-StatsBase.coef(m::MixedModel) = fixef(m, false)
+function StatsBase.coef(m::LinearMixedModel{T}) where {T}
+    piv = fetrm(m).piv
+    invpermute!(fixef!(similar(piv, T), m), piv)
+end
 
-βs(m::LinearMixedModel) = NamedTuple{(Symbol.(fixefnames(m))...,)}((fixef(m)...,))
+βs(m::LinearMixedModel) = NamedTuple{(Symbol.(coefnames(m))...,)}(coef(m))
 
-StatsBase.coefnames(m::LinearMixedModel) = fixefnames(m, false)
+function StatsBase.coefnames(m::LinearMixedModel)
+    Xtrm = fetrm(m)
+    invpermute!(copy(Xtrm.cnames), Xtrm.piv)
+end
 
-function StatsBase.coeftable(m::MixedModel)
+function StatsBase.coeftable(m::LinearMixedModel)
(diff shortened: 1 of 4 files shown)


dmbates:
I cancelled the windows-2019 check because it had been running for nearly 40 minutes.

src/linearmixedmodel.jl
```diff
@@ -159,15 +159,21 @@ fit(
...
 )
 
-StatsBase.coef(m::MixedModel) = fixef(m, false)
+function StatsBase.coef(m::LinearMixedModel{T}) where {T}
+    piv = fetrm(m).piv
+    invpermute!(fixef!(similar(piv, T), m), piv)
+end
 
-βs(m::LinearMixedModel) = NamedTuple{(Symbol.(fixefnames(m))...,)}((fixef(m)...,))
+βs(m::LinearMixedModel) = NamedTuple{(Symbol.(coefnames(m))...,)}(coef(m))
```
palday: `GeneralizedLinearMixedModel` has a property `beta`: not the quite the same as the accessor here, but should standardize? And/or do a `MixedModel` method instead -- `coefnames` and `coef` are defined for both GLMM and LMM.

src/linearmixedmodel.jl
```diff
@@ -285,13 +291,30 @@ function StatsBase.dof_residual(m::LinearMixedModel)::Int
...
+    k = feind(m)
+    LowerTriangular(m.L.blocks[k, k])
+end
+
+"""
+    fetrm(m::LinearMixedModel)
+
+Return the fixed-effects term from `m.allterms`
+"""
+fetrm(m) = m.allterms[feind(m)]
```
palday: I like having these helpers around.

src/linearmixedmodel.jl
```diff
@@ -356,53 +380,44 @@ StatsBase.fitted(m::LinearMixedModel{T}) where {T} = fitted!(Vector{T}(undef, no
 """
     fixef!(v::Vector{T}, m::LinearMixedModel{T})
 
-Overwrite `v` with the pivoted and, possibly, truncated fixed-effects coefficients of model `m`
-"""
-fixef!(v::AbstractVector{T}, m::LinearMixedModel{T}) where {T} =
-    ldiv!(feL(m)', copyto!(v, m.L.blocks[end, end-1]))
+Overwrite `v` with the pivoted fixed-effects coefficients of model `m`
 
+For full rank models the length of `v` must be the rank of `X`.  For rank-deficient models
```
palday: "full-rank" for consistency with "rank-deficient"

src/linearmixedmodel.jl
```diff
@@ -880,7 +919,7 @@ function StatsBase.vcov(m::LinearMixedModel{T}; corr=false) where {T}
         vv = covmat[iperm, iperm]
     end
 
-    corr ?  StatsBase.cov2cor!(vv,stderror(m)) : vv
+    corr ?  StatsBase.cov2cor!(vv, stderror(m)) : vv
```
palday: Did you want to redefine `stderror` to redirect to `stderror!` plus a pre-allocated scratch like `fixef` and `fixef!`?

palday:
> I cancelled the windows-2019 check because it had been running for nearly 40 minutes.

It's Tier 2 anyway :)

src/linearmixedmodel.jl
dmbates: I think it is best to keep separate methods for `LinearMixedModel` and `GeneralizedLinearMixedModel`.  When beta is part of the "baba yaga" optimization you can't guarantee that the calculation using only the LMM field will produce the correct answer.  This is related to the distinction between `fast=true` and `fast=false`.

src/linearmixedmodel.jl
dmbates: :+1:

src/linearmixedmodel.jl
dmbates: Thanks for catching that.  I will fix it.

src/linearmixedmodel.jl
dmbates: I suppose now that I have written `stderror!` we should use it.  I expect we will see reductions in the hundreds of nanoseconds in the execution times. :smiley:

dmbates:
@palday In my local copy of the repository I did a rebase on current master, which worked without a hitch.  Is it necessary to do a rebase and force push or does the squash and merge take care of that because there is no overlap?

palday:
`squash` should remove any potential redundancies because collapses things down into final differences. You may want to edit the commit message to remove the notes from intermediate commits. Force-pushing at this point wouldn't hurt, but I don't think it's necessary.

PR #296, opened 2020-03-20
Leverage

dmbates:
Add a `leverage` method plus tests and docs.

This method can definitely be improved, taking advantage of the blocked structure and the sparsity of the (1,1) block of L.  However, it works and is easy to implement.

I did check the results against the `hatvalues` method in `lme4` on some examples.


palday:
Nice addition.

This actually seems deeply related to profiling somehow.

docs/src/constructors.md
```diff
@@ -433,3 +433,28 @@ condVar
...
+For a linear model without random effects the fitted values are obtained by orthogonal projection of the response onto the column span of the model matrix and the sum of the leverage values is the dimension of this column span.
+That is, the sum of the leverage values is the rank of the model matrix and `n - sum(leverage(m))` is the degrees of freedom for residuals.
+The sum of the leverage values is also the trace of the so-called "hat" matrix, `H`.
+(The name "hat matrix" reflects the fact that $\hat{\mathbf{y}} = \mathbf{H} \mathbf{y}$.  That is, `H` puts a hat on `y`.)
+
+For a linear mixed model the sum of the leverage values will be between `p`, the rank of the fixed-effects model matrix, and `p + q` where `q` is the total number of random effects.
+This number does not represent a dimension (or "degrees of freedom") of a linear subspace of all possible fitted values because the projection is not an orthogonal projection.
+Nevertheless, it is a reasonable measure of the effective degrees of freedom of the model and `n - sum(leverage(m))` can be considered the effective residual degrees of freedom.
+```@example Main
+n, p, q, k = size(fm1)
```
palday: you should probably print these values if you want to netter draw the comparison between the different sizes.

dmbates: I thought I replied to your comment but I may have forgotten to click the comment button.  I have added two more examples, one of which, `dyestuff2`, converges to a random-effects variance of zero therefore its effective degrees of freedom is at the lower bound.  To get to the upper bound we may try the `penicllin` example.

palday:
This can merged whenever you're ready. We still need to think about whether we should update `residual_dof` (maybe adding a method that takes an `approximation` argument with options `:asymptotic`, `:hat` and maybe `:satterthwaite`, which are all reasonably efficient, even for naive implementations) and the coeftable.

PR #297, opened 2020-03-23
Bump compatibility for Tables; patch version bump

palday:
If everybody is happy with this, then call JuliaRegistrator on the relevant commit. After TagBot has done it's job, close the PR without merging and delete the branch.  The tag will keep the relevant commit from being garbage collected by git.


palday:
@Nosferican Do you want to have the honor of making the release?

palday:
Yes.

palday:
Finally got to use `git cherry-pick`.

palday:
So bringing in those changes cuts Julia 1.2 support -- is that too much for a patch release?

palday:
In other words, should this move from a patch to a minor release?

dmbates:
Yes, minor release I think.

Nosferican:
What exactly cuts the support for Julia v1.2?

palday:
The `mul!` methods -- these got pulled in with cherry-pick to support newer Distributions (which required newer BlockArrays).

dmbates:
@JuliaRegistrator register()

PR #298, opened 2020-03-23
Replace Printf by Base.Ryu for Julia-1.4

### LINKED ISSUE #246: Switch from Printf to Base.Ryu with 1.4.0
This is just a reminder that after Julia v1.4.0 is released we can drop the dependency on the `Printf` package, which is used exactly twice, and use `Base.Ryu.writefixed` in those two places.

dmbates:
Closes #246


src/varcorr.jl
```diff
@@ -54,7 +54,7 @@ function Base.show(io::IO, vc::VarCorr)
             for l = 1:(j-1)
                 ρind += 1
                 ρval = ρ[ρind]
-                ρval === -0.0 ? write(io, "   .  ") : @printf(io, "%6.2f", ρval)
+                ρval === -0.0 ? write(io, "   .  ") : write(io, lpad(Ryu.writefixed(ρval, 2, true), 6))
```
palday: I was going to ask if the change to include explicit `+` in the correlations was intentional, but clearly it was.

PR #299, opened 2020-03-26
Nicer describeblocks output

### LINKED ISSUE #285: Clean up printing the output of `describeblocks`
The current `show` method is rather perfunctory.

dmbates:
Closes #285

### DIFF
--- a/src/blockdescription.jl
+++ b/src/blockdescription.jl
@@ -4,10 +4,13 @@
 Description of blocks of `A` and `L` in a [`LinearMixedModel`](@ref)
 
 ## Fields
-* `blknms`: Vector of block names as `String`s
-* `blkrows`: Vector of number of rows in each block
-* `Atypes`: Matrix of `DataType`s for blocks in `A`
-* `Ltypes`: Matrix of `DataType`s of blocks in `L`
+* `blknms`: Vector{String} of block names
+* `blkrows`: Vector{Int} of the number of rows in each block
+* `ALtypes`: Matrix{String} of datatypes for blocks in `A` and `L`.
+
+When a block in `L` is the same type as the corresponding block in `A`, it is
+described with a single name, such as `Dense`.  When the types differ the entry
+in `ALtypes` is of the form `Diag/Dense`, as determined by a `shorttype` method.
 """
 struct BlockDescription
     blknms::Vector{String}
--- a/src/blockdescription.jl
+++ b/src/blockdescription.jl
@@ -1,0 +1,57 @@
+"""
+    BlockDescription
+
+Description of blocks of `A` and `L` in a [`LinearMixedModel`](@ref)
+
+## Fields
+* `blknms`: Vector of block names as `String`s
+* `blkrows`: Vector of number of rows in each block
+* `Atypes`: Matrix of `DataType`s for blocks in `A`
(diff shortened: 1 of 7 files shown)


dmbates:
`describeblocks` now produces output like
```julia
julia> describeblocks(m1)
Blk:         A type                dim                L type
1,1:  UniformBlockDiagonal      (224, 224)      UniformBlockDiagonal
2,1:                 Dense      (128, 224)      Dense
2,2:  UniformBlockDiagonal      (128, 128)      Dense
3,1:                 Dense       (8, 224)       Dense
3,2:                 Dense       (8, 128)       Dense
3,3:                 Dense        (8, 8)        Dense
4,1:                 Dense       (1, 224)       Dense
4,2:                 Dense       (1, 128)       Dense
4,3:                 Dense        (1, 8)        Dense
4,4:                 Dense        (1, 1)        Dense
```

An alternative would be to condense the type names even further and lay them out in the form of the blocks themselves.  Something like
A types
size            subj        item      fixed
224:      BlkDiag
128:       Dense      BlkDiag
    8:       Dense        Dense   Dense

L types
size            subj        item      fixed
224:      BlkDiag
128:       Dense        Dense
    8:       Dense        Dense   Dense

Does this seem like a better choice?

palday:
Hmmmm.... I like both ways and think they have different use cases. The first variant is nicer for 'us' (developers and experts), who care more about the computational aspect. I think the latter version could be used as nice way for users to check their data (e.g. dense blocks from partial crossing from issues with with identifiers). 

@christinabergmann As a power user who's been exposed to the inner workings of theblocks, what do you think?

christinabergmann:
Oh I like the condensed form, indeed, it's more end-user friendly, so the question is how much you want to promote it for quick data checks versus actual development

dmbates:
I will implement the condensed form, use it as the default, and have the current form as an alternative selected by an optional Boolean argument.

src/blockdescription.jl
```diff
@@ -0,0 +1,58 @@
+"""
+    BlockDescription
+
+Description of blocks of `A` and `L` in a [`LinearMixedModel`](@ref)
+
+## Fields
+* `blknms`: Vector of block names as `String`s
+* `blkrows`: Vector of number of rows in each block
+* `Atypes`: Matrix of `DataType`s for blocks in `A`
+* `Ltypes`: Matrix of `DataType`s of blocks in `L`
```
palday, line 10: this doesn't match the `ALtypes` field

dmbates: Thanks for picking that up.

test/pls.jl
```diff
@@ -58,7 +58,7 @@ fits = Dict(k => fit.(MixedModel, forms[k], Ref(MixedModels.dataset[k])) for k i
     
     io = IOBuffer()
     show(io, BlockDescription(fm1))
-    @test_broken countlines(io) == 3
+    @test_broken countlines(io) == 3  # probably needs seekstart(io) but that will break the next test
```
palday, line 61: what happens if you swap the order of the tests here and use `seekstart`?

dmbates: Latest version works on my machine at least.

dmbates:
@palday Is this ready to go?

palday:
@dmbates I added a GLMM method. I'll squash and merge when all the tests pass.

PR #301, opened 2020-04-06
Tighten code, increase test coverage

dmbates:


test/pls.jl
```diff
@@ -32,31 +30,31 @@ const forms = Dict(
...
         @formula(reaction ~ 1 + days + (1+days|subj)),
     ],
 )
 
-#=         # replace all the model fits by
-fits = Dict(k => fit.(MixedModel, forms[k], Ref(MixedModels.dataset[k])) for k in keys(forms))
-=#
+fits = Dict(k => fit.(MixedModel, fms[k], Ref(MixedModels.dataset(k))) for k in keys(fms))
+
+const io = IOBuffer()
```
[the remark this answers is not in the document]
dmbates: I think you are right that it is not a good idea to all the heavy lifting of fitting many models at the beginning of the script before any testsets are active.  I will change that.

test/pls.jl
```diff
@@ -35,12 +37,19 @@ const fms = Dict(
...
 
-fits = Dict(k => fit.(MixedModel, fms[k], Ref(MixedModels.dataset(k))) for k in keys(fms))
+const fittedmodels = Dict{Symbol,Vector{LinearMixedModel}}();
+function models(nm::Symbol)
+    if haskey(fittedmodels, nm)
+        fittedmodels[nm]
+    else
+        fittedmodels[nm] = fit.(MixedModel, fms[nm], Ref(dataset(nm)))
+    end
+end
```
palday: This somehow makes me very happy, as does the commit message.

dmbates: There is also a special method for `get!` that allows you to use a thunk so the method could be written
```julia
models(nm::Symbol) = get!(fittedmodels, nm) do nm
     fit.(MixedModel, fms[nm], Ref(dataset(nm))
end
```
except that the "thunk" function cannot take an argument.  Thus this, slightly wordier, method.

dmbates:
Grrr.  It seems that the `issingular` method is producing different answers under Windows.  Probably need to tweak the tolerances.  @palday do you have easy access to a Windows machine for testing?  I can reboot a computer into Windows but I only have 64-bit Windows.

palday:
Since I can no longer go into the office, the same holds for me -- I have a laptop I can boot into 64-bit Windows. 

Actually .... 64-bit Windows gets its 32-bit compatibility by basically running 32-bit programs a 32-Windows running on top of 64-bit Windows (hence the WoW acronym in some of the filenames). If you have 64-bit Windows, you may be able to replicate the 32-bit failure just by using the 32-bit julia build.

Otherwise for such things, I tend to modify the test so that gives a bit more debugging output. Once it's fixed, I remove the debugging output and push again -- the commits adding and removing the output cancel out in the squash commit anyway. 

There's also the bigger question of whether we want to support x86 Windows .... I haven't seen an x86 Windows box in years, just lots of never-ported 32-bit software running on 64-bit Windows

dmbates:
@palday I think this is ready to go.  Can you take a look at it please?  If it seems okay then please squash and merge.

src/linearmixedmodel.jl
```diff
@@ -303,17 +303,15 @@ objective and the parameters are printed on stdout at each function evaluation.
 function fit!(m::LinearMixedModel{T}; verbose::Bool = false, REML::Bool = false) where {T}
     optsum = m.optsum
     opt = Opt(optsum)
-    feval = 0
     optsum.REML = REML
     function obj(x, g)
-        isempty(g) || error("gradient not defined")
-        feval += 1
+        isempty(g) || throw(ArgumentError("g should be empty for this objective"))
```
palday, line 308: I think this error message is more informative.

src/linearmixedmodel.jl
```diff
@@ -303,17 +303,15 @@ objective and the parameters are printed on stdout at each function evaluation.
...
         val = objective(updateL!(setθ!(m, x)))
-        feval == 1 && (optsum.finitial = val)
-        verbose && println("f_", feval, ": ", round(val, digits = 5), " ", x)
+        verbose && println(round(val, digits = 5), " ", x)
```
palday, line 310: I'm okay with leaving the current evaluation number out of the verbose output, but this might be something we add back in later if users prefer it ....

src/linearmixedmodel.jl
```diff
@@ -474,8 +472,10 @@ end
     issingular(m::LinearMixedModel, θ=m.θ)
 
 Test whether the model `m` is singular if the parameter vector is `θ`.
+
+Equality comparisons are used b/c small non-negative θ values are replaced by 0 in `fit!`.
 """
-issingular(m::LinearMixedModel, θ=m.θ) = any(isapprox.(lowerbd(m), θ))
+issingular(m::LinearMixedModel, θ=m.θ) = any(lowerbd(m) .== θ)
```
palday, line 478: I get perverse pleasure in warranted exact comparisons in floating point. :)

test/pirls.jl
```diff
@@ -53,22 +58,20 @@ end
...
-        MixedModels.dataset(:verbagg), Bernoulli())
+    gm3 = fit(MixedModel, only(fms[:verbagg]), dataset(:verbagg), Bernoulli())
     @test deviance(gm3) ≈ 8151.40 rtol=1e-5
     @test lowerbd(gm3) == vcat(fill(-Inf, 6), zeros(2))
     @test fitted(gm3) == predict(gm3)
     # these two values are not well defined at the optimum
-    @test isapprox(sum(x -> sum(abs2, x), gm3.u), 273.29266717430795, rtol=1e-3)
-    @test sum(gm3.resp.devresid) ≈ 7156.547357801238 rtol=1e-4
+    @test isapprox(sum(x -> sum(abs2, x), gm3.u), 273.29646346940785, rtol=1e-3)
+    @test sum(gm3.resp.devresid) ≈ 7156.550941446312 rtol=1e-4
```
palday, line 67: These are irrelevant changes in the deviance residuals, but still what changed?

test/pirls.jl
```diff
@@ -1,25 +1,31 @@
...
 
+using MixedModels: dataset
+
+const fms = Dict(
+    :cbpp => [@formula((incid/hsz) ~ 1 + period + (1|herd))],
+    :contra => [@formula(use ~ 1+age+abs2(age)+urban+livch+(1|urbdist))],
+    :grouseticks => [@formula(ticks ~ 1+year+ch+ (1|index) + (1|brood) + (1|location))],
+    :verbagg => [@formula(r2 ~ 1+anger+gender+btype+situ+(1|subj)+(1|item))],
+)
+
```
palday, line 12: Will this migrate at some point to the `tests/pls.jl` just in time fitting? For now, this extra layer of indirection doesn't help readability of the tests.

test/pls.jl
```diff
@@ -94,6 +99,7 @@ fits = Dict(k => fit.(MixedModel, forms[k], Ref(MixedModels.dataset[k])) for k i
     @test fm1.objective == objective(fm1)
     @test fm1.σ ≈ 49.510099986291145 atol=1.e-5
     @test fm1.X == ones(30,1)
+    ds = MixedModels.dataset(:dyestuff)
```
palday, line 102: since you used fancy `using`:

```suggestion
    ds = dataset(:dyestuff)
```

test/pls.jl
```diff
@@ -139,42 +149,49 @@ end
...
+    @test fm.optsum.initial == ones(2)
     @test lowerbd(fm) == zeros(2)
 
-    fit!(fm)
-
     @test objective(fm) ≈ 332.18834867227616 atol=0.001
     @test coef(fm) ≈ [22.97222222222222] atol=0.001
     @test fixef(fm) ≈ [22.97222222222222] atol=0.001
-    @test coef(fm)[1] ≈ mean(pen.diameter)
+    @test coef(fm)[1] ≈ mean(MixedModels.dataset(:penicillin).diameter)
```
palday, line 166:
```suggestion
    @test coef(fm)[1] ≈ mean(dataset(:penicillin).diameter)
```

test/pls.jl
```diff
@@ -276,9 +302,9 @@ end
     @test fmnc.θ == [fm.θ[1], fm.θ[3]]
     @test lowerbd(fmnc) == zeros(2)
 
-    fmnc = zerocorr!(LMM(@formula(reaction ~ 1 + days + (1+days|subj)), slp));
```
palday, line 279: do we test `zerocorr!` applied to a whole model anywhere now?

test/pls.jl
```diff
@@ -302,25 +326,8 @@ end
...
-    @test fixef(fmnc2) ≈ [251.40510484848477, 10.467285959595715]
-    @test stderror(fmnc2) ≈ [6.707710260366577, 1.5193083237479683] atol=0.001
-    @test fmnc2.θ ≈ [0.9458106880922268, 0.22692826607677266] atol=0.0001
-    @test first(std(fmnc2)) ≈ [24.171449463289047, 5.799379721123582]
-    @test last(std(fmnc2)) ≈ [25.556130034081047]
-    @test logdet(fmnc2) ≈ 74.46952585564611 atol=0.001
-
     MixedModels.likelihoodratiotest(fm, fmnc)
-
+    slp = MixedModels.dataset(:sleepstudy)
```
palday, line 330:
```suggestion
    slp = dataset(:sleepstudy)
```

test/pls.jl
```diff
@@ -336,9 +335,9 @@ end
 
     # combining [ReMat{T,S1}, ReMat{T,S2}] for S1 ≠ S2
     slpcat = categorical!(deepcopy(slp), [:days])
-    fm_cat = LMM(@formula(reaction ~ 1 + days + (1|subj) + (0 + days|subj)), slpcat)
-    @test fm_cat isa LMM
-    σρ = fit!(fm_cat).σρs
+    fm_cat = fit(MixedModel, @formula(reaction ~ 1+days+(1|subj)+(0+days|subj)),slpcat)
```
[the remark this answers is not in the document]
palday: Probably not without a lot of special casing, so just leave it here.

test/pls.jl
```diff
@@ -348,22 +355,41 @@ end
...
 
 @testset "d3" begin
-    fm = updateL!(LMM(@formula(y ~ 1 + u + (1+u|g) + (1+u|h) + (1+u|i)), MixedModels.dataset(:d3)));
-    @test pwrss(fm) ≈ 5.1261847180180885e6 rtol = 1e-6
-    @test objective(fm) ≈ 901641.2930413672 rtol = 1e-6
-    fit!(fm)
-    @test objective(fm) ≈ 884957.5540213 rtol = 1e-6
+    fm = only(models(:d3))
+    @test pwrss(fm) ≈ 5.30480294295329e6 rtol=1.e-4
+    @test objective(fm) ≈ 884957.5540213 rtol = 1e-4
```
palday, line 372: this is a big change in the objective -- what happened?

src/linearmixedmodel.jl
dmbates: The reason I did this was because I discovered that `NLopt` now provides `numevals` so we don't need to keep track of the number of evaluations separately.

src/linearmixedmodel.jl
dmbates: I am making the assumption that the value of θ is the result of a call to `fit!`, which should be the case. Nevertheless it is a hidden assumption on the validity of the method but I don't think it is necessary to be too pedantic about this one.

test/pirls.jl
dmbates: From time to time I update those numbers to what seems to be the latest results - on the vague assumption that things might have gotten more accurate/reliable. Those numbers are the result of my running the examples by hand on my Intel i5 / Ubuntu 19.10 / Julia 1.4.0 system

src/linearmixedmodel.jl
palday: I completely agree and no snark was intended. I think the exact comparison here is actually nice because it means we only have one set of thresholds (those in `fit!`) for numerical singularity.

test/pirls.jl
dmbates: I agree that this may not be a good approach here because there are additional arguments needed or used when fitting a GLMM.  Perhaps we could leave this in here for the time being with a mental note to check back and see if there are better ways to organize the dataset/formula/distribution/link/options combinations later.  It may be that reverting to individual formula expressions inside each of the testsets will be the best choice after all.

test/pls.jl
dmbates: I guess we should add such a test back in. 

An alternative is to drop the use of `zerocorr!` entirely in favor of `zerocorr` on individual terms.  It is always good when there are multiple ways of doing something to ask if there is anything to be gained from having said multiple ways.  @kleinschmidt Do you have any opinions?

test/pls.jl
dmbates: I'm not sure I understand the comment.  I don't see a difference in the objective.  I'm not sure why I dropped the tolerance.  Right now if I run it on the local machine under julia-1.4.0 I get an objective of 884957.5539700153.  Perhaps I was confusing `rtol` and `atol`.

palday: Ah, this was a mistake on my part reading the diff:

```diff
-    fm = updateL!(LMM(@formula(y ~ 1 + u + (1+u|g) + (1+u|h) + (1+u|i)), MixedModels.dataset(:d3)));
-    @test pwrss(fm) ≈ 5.1261847180180885e6 rtol = 1e-6
-    @test objective(fm) ≈ 901641.2930413672 rtol = 1e-6 # <= READ THIS LINE
-    fit!(fm)
-    @test objective(fm) ≈ 884957.5540213 rtol = 1e-6  # <= INSTEAD OF THIS LINE
+    fm = only(models(:d3))
+    @test pwrss(fm) ≈ 5.30480294295329e6 rtol=1.e-4
+    @test objective(fm) ≈ 884957.5540213 rtol = 1e-4
```

test/pls.jl
kleinschmidt: I think for most of the cases where someone might want to use `zerocorr!` they're probably going to have a small number of grouping factors with a lot of terms in each one, so there's not much to be gained from `zerocorr!`

palday: And post-hoc modification of fitted models (i.e. the RK approach)?

dmbates: To answer the original question by @palday yes we do test `zerocorr!` elsewhere - a few lines above this.  It is line 310 in the current numbering.

Let's leave this as it is and remind ourselves to revisit the question of whether to keep `zerocorr!` in the package at a later time.

palday:
The last remaining test failure looks semi-legitimate.: it can't find the correlation value in the `kb07` output.

dmbates:
I think the test failure was spurious.  If you agree @palday please squash and merge.

dmbates:
But all of the other architectures did find that value and previous CI runs on this architecture found it.  Is there a way of restarting a single test?

dmbates:
Also, I don't understand why there is a Tier 2 test of the same combination of Julia version, architecture and operating system.  @Nosferican have I made a mistake in changing `Tier1.yml` and `Tier2.yml` for the minimum Julia version of 1.4?   We do want a minimum Julia version of 1.4 - what I am asking is did I create an accidental duplication of test runs?

PR #303, opened 2020-04-11
Updated Test Coverage for Current Release

Nosferican:
Updated /README.md
Updated /.github/workflows/Tier.yml
Dropped testing on v1.3 and duplicate runs.
Related https://github.com/JuliaStats/MixedModels.jl/pull/301#issuecomment-612257345


palday:
Can you rebase and force push? Then we can double check it the spurious failure on #301.

Nosferican:
Seems like a legitimate error.

dmbates:
I apparently chose a poor example - the "semi-maximal" model for the `kb07` data - to check on printing of the estimated correlations in the random effects.  I thought the estimate would be stable to 2 significant digits across architectures but apparently not.  I'm pretty sure that github.com uses AMD processors which may be these values are different from the ones I get on an Intel-based (1.4, x64, ubuntu-19.10) system.

In any case I changed the example to a more stable example in commit 19c87caa1215db5fc10820ea1306428d41ce81a8 but now I can't figure out how to incorporate that commit into this PR.  Can anyone help?

palday:
@dmbates I cherry-picked it over. @Nosferican thanks for marking that members can edit. :)

palday:
@dmbates I find such numeric differences across architectures somehow delightful and terrible at the same time. I was planning on upgrading my desktop in the near-to-milddle future and found some of the stuff AMD has been doing interesting. Since I already have Intel processors elsewhere, this might be enough to push me over ...

dmbates:
@palday Those sound like some of the explanations I give to my wife about why I should go out and buy yet another computer.

Issue #307, opened 2020-04-24
values of GLMM parameter estimates in show method


dmbates:
In a message to R-SIG-Mixed-Models@R-Project.org with subject "nAGQ > 1 in lme4::glmer gives unexpected likelihood" Ben Goldstein provided simulated data
```julia
julia> goldstein = categorical!(DataFrame(y = Int32[83, 3, 8, 78, 901, 21, 4, 1, 1, 39, 82, 3, 2, 82, 874, 18, 5, 1, 3, 50, 87, 7, 3, 67, 914, 18, 0, 1, 1, 38, 86, 13, 5, 65, 913, 13, 2, 0, 0, 48, 90, 5, 5, 71, 886, 19, 3, 0, 2, 32, 96, 1, 1, 87, 860, 21, 3, 0, 1, 54, 83, 2, 4, 70, 874, 19, 5, 0, 4, 36, 100, 11, 3, 71, 950, 21, 6, 0, 1, 40, 89, 5, 5, 73, 859, 29, 3, 0, 2, 38, 78, 13, 6, 100, 852, 24, 5, 0, 1, 39], group = repeat(1:10, outer=10)), :group);

julia> by(goldstein, :group, mn = :y => mean)
10×2 DataFrame
│ Row │ group        │ mn      │
│     │ Categorical… │ Float64 │
├─────┼──────────────┼─────────┤
│ 1   │ 1            │ 87.4    │
│ 2   │ 2            │ 6.3     │
│ 3   │ 3            │ 4.2     │
│ 4   │ 4            │ 76.4    │
│ 5   │ 5            │ 888.3   │
│ 6   │ 6            │ 20.3    │
│ 7   │ 7            │ 3.6     │
│ 8   │ 8            │ 0.3     │
│ 9   │ 9            │ 1.6     │
│ 10  │ 10           │ 41.4    │
```
The peculiar thing is that fitting a simple GLMM model by Laplace approximation and with `nAGQ=11` produces identical parameter estimates in the show method.
```julia
julia> m1 = fit(MixedModel, @formula(y ~ 1 + (1|group)), dd, Poisson())
Generalized Linear Mixed Model fit by maximum likelihood (nAGQ = 1)
  y ~ 1 + (1 | group)
  Distribution: Poisson{Float64}
  Link: LogLink()

  Deviance: 193.5587

Variance components:
         Column    Variance  Std.Dev. 
group (Intercept)  3.9577026 1.9893975

[...]

PR #291, opened 2020-02-26
Take the dispersion into account for GLMM familes with a dispersion param

palday:
Closes #206 

- [x] `varest`
- [ ] `deviance`
- [ ] `loglikelihood`
- [ ] export all the usual links
    - [ ] identity
    - [ ] sqrt
    - [ ] log
    - [ ] inverse
    - [ ] probit
    - [ ] logit
- [ ] tests for this functionality in all families
    - [ ] Bernoulli
    - [ ] Binomial 
    - [ ] Gamma 
    - [ ] Gaussian with non-identity link
    - [ ] InverseGaussian
    - [ ] Poisson 
 
Note (to self) that some distributions (e.g. Gamma) require changing the linear predictor + dispersion into a different parameterization (e.g. shape, scale).

PR #308, opened 2020-04-25
correct coeftable for GLMMs, add tests

dmbates:
See  #307 

Note that the example for this PR ("goldstein") still has the peculiar property that the deviances change between the Laplace approximation and `nAGQ=11` but the parameter estimates are exactly the same.  Now the multipliers, `m11.mult` are all close to 1.0 but not that close. 

I don't think this is right still


dmbates:
It looks as if the identical parameter estimates for the Laplace approximation and the nAGQ=11 fit for the goldstein data may be a coincidence.  Fitting such models to the contra data does produce slightly different values for the parameter estimates.

src/generalizedlinearmixedmodel.jl
```diff
@@ -567,6 +578,16 @@ function Base.show(io::IO, m::GeneralizedLinearMixedModel)
...
+function σs(m::MixedModel)
+    σ = dispersion(m)
+    NamedTuple{fnames(m)}(((σs(t, σ) for t in m.reterms)...,))
+end
+
+function σρs(m::MixedModel)
+    σ = dispersion(m)
+    NamedTuple{fnames(m)}(((σρs(t, σ) for t in m.reterms)...,))
+end
+
```
palday, line 590: should we have a separate `mixedmodel.jl` file for methods defined for the superclass?

src/generalizedlinearmixedmodel.jl
```diff
@@ -401,18 +413,17 @@ function Base.getproperty(m::GeneralizedLinearMixedModel, s::Symbol)
...
+        if D <: Binomial
+            sum(logpdf(D(round(Int, n), μ), round(Int, y * n)) 
+                for (μ, y, n) in zip(r.mu, r.y, m.wt))
+        else
+            sum(logpdf(D(μ), y) for (μ, y) in zip(r.mu, r.y))
         end
-    end
-    accum - (mapreduce(u -> sum(abs2, u), +, m.u) + logdet(m)) / 2
+    )
+    accum - (sum(sum(abs2, u) for u in m.u) + logdet(m)) / 2
```
palday, line 426: This will go away with the proposed changes in #291. (No change necessary here, just a note to self.)

src/generalizedlinearmixedmodel.jl
```diff
@@ -589,16 +610,14 @@ varest(m::GeneralizedLinearMixedModel{T}) where {T} = one(T)
...
-    :describeblocks,
     :feL,
+    :fetrm,
     :(LinearAlgebra.logdet),
     :lowerbd,
     :PCA,
     :rePCA,
+    :(StatsBase.coefnames),
     :(StatsModels.modelmatrix),
-    :(StatsBase.vcov),
```
palday, line 599: There is now no `vcov` method for `GeneralizedLinearMixedModel`.

src/generalizedlinearmixedmodel.jl
```diff
@@ -53,10 +53,19 @@ struct GeneralizedLinearMixedModel{T<:AbstractFloat} <: MixedModel{T}
...
+    z = co ./ se
+    pvalue = ccdf.(Chisq(1), abs2.(z))
+    CoefTable(
+        hcat(co, se, z, pvalue),
+        ["Estimate", "Std.Error", "z value", "P(>|z|)"],
+        coefnames(m),
+        4, # pvalcol
+        3, # teststatcol
+    )
+end
```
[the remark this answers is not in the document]
dmbates: Certainly.  Thanks for catching that.  I will create a new file with the methods that are common to all forms of `MixedModel`.

src/generalizedlinearmixedmodel.jl
dmbates: Yes.

dmbates: The trap I always fall into here is using both the upper-case and lower-case names `MixedModels.jl` and `mixedmodels.jl` then finding out that certain operating systems still use case-insensitive file names.  Very important for those who still want to do their computing on punched cards.

palday: This also occurred to me after I reviewed ... but then we have the distinction between `MixedModels.jl` for the package and `mixedmodel.jl` for the class. While I typically don't like have an important identifier differ only singular/plural, I can't think of a better name for the file with the superclass methods.

dmbates:
@palday Will you have the opportunity to look at the latest commit to see if it addresses the questions you had?

palday:
There are a couple of spots where the superclass methods aren't there yet (`coeftable` that depends on `coef`, which isn't defined for GLMM), but I'll add them in.

palday:
Ah, can't do it for the cherry picking after all -- the `fixef` methods for GLMM and LMM take different arguments (GLMM has a `permuted` argument, but LMM doesn't -- it instead depends on the difference between `fixef` and `coef`). So that would be a breaking change.  I'll spin that re-factoring off into a different PR.

palday:
Ah, so good news -- the `coeftable` method for GLMM is actually correct in the last release. The wrong table was introduced in #287 when the static allocation for the coefficients and errors was introduced for LMM. It still makes sense to add in the GLMM parallels for those things, then we can go back to more generic methods for things like `coeftable` that build upon them.

PR #314, opened 2020-05-13
Remove some tests dependency on DataFrames and RNG results

### LINKED ISSUE #306: possible test failure in upcoming Julia version 1.5
A PkgEval run for a [PR](https://github.com/JuliaLang/julia/pull/35078) which changes the generated numbers for `randn!` indicates that the tests of this package might fail in Julia 1.5 (and on Julia current master). Apologies if this is a false positive.
cf. https://github.com/JuliaCI/NanosoldierReports/blob/7de24e455342298cbef56826b5827f0d7640d2c1/pkgeval/by_hash/b89e35c_vs_098ef24/logs/MixedModels/1.5.0-DEV-71a4a114c2.log

dmbates:
Closes #306.


test/linalg.jl
```diff
@@ -31,38 +38,45 @@ using DataFrames, LinearAlgebra, MixedModels, Random, SparseArrays, StatsModels,
...
+    rng = MersenneTwister(1234321)
+    df = (
+        Y = randn(rng, 400),
+        A = repeat(PooledArray(["N","Y"]), outer=200),
+        G = repeat(PooledArray(string.('A':'T')), inner = 2, outer=10),
+        H = repeat(PooledArray(string.('a':'j')), inner=40),
+        )
+    m1 = fit(MixedModel, @formula(Y ~ 1 + A + (1+A|G) + (1+A|H)), df)
+    wm1 = fit(MixedModel, @formula(Y ~ 1+A+(1+A|G)+(1+A|H)), df, wts  = ones(400))
+    @test loglikelihood(wm1) ≈ loglikelihood(m1)
```
palday, line 50: is this necessary? or could you just deepcopy `m1`, the do `reweight!` and check the fits? (Or do we need weights in the original fit to able to re-weight?)

test/linalg.jl
```diff
@@ -31,38 +38,45 @@ using DataFrames, LinearAlgebra, MixedModels, Random, SparseArrays, StatsModels,
...
-                       G = rand(rng, 1:ng, n),
-                       H = rand(rng, 1:ng, n))
-        categorical!(df, [:G, :H])
+    function gendata(rng::AbstractRNG, n::Integer, ng::Integer, nh::Integer)
+        (
+            Y = randn(rng, n),
+            X = rand(rng, n),
+            G = PooledArray(rand(rng, levs(ng, 'G'), n)),
+            H = PooledArray(rand(rng, levs(nh, 'H'), n)),
+        )
```
palday, line 76: I like removing the dependency on `DataFrames`

test/linalg.jl
```diff
@@ -31,38 +38,45 @@ using DataFrames, LinearAlgebra, MixedModels, Random, SparseArrays, StatsModels,
...
     end
+    gendata(n::Integer, ng::Integer, nh::Integer) = gendata(MersenneTwister(42), n, ng, nh)
+    gendata(n::Integer, ng::Integer) = gendata(MersenneTwister(42), n, ng, ng)
 
     @testset "Adjoint{T,ReMat{T,1}}, BlockedSparse{T,1,2}" begin
         # this is an indirect test of lmulΛ! for a blocking structure found in
```
palday, line 82: I'm not sure this test is still doing what we want it to do for other random numbers. From the [original issue](https://github.com/JuliaStats/MixedModels.jl/issues/123#issuecomment-535497527 ):
> This highlights the point about needing a reproducible RNG -- the particular blocking here arises from a particular set of pseudorandom numbers.

palday:
@dmbates Despite my previous concerns, I think we can merge this if you agree that those concerns can be ignored.

PR #317, opened 2020-05-14
support NLopt 0.6 (and thus Julia 1.5)

### LINKED ISSUE #315: Conflict with the latest version of NLopt v0.6.0 #master
Hello!
I have difficulty to add the current master branch of MixedModels.jl due to the conflict with the working version of NLOpt v0.6.0 #master. Currently, that version of NLopt is the only version that works. Previous versions, e.g., v0.5.0 and v0.5.1 do not work under Julia 1.4. So, I added this master branch of NLopt first. Then, I tried to add the MixedModels.jl, but got the following error:
```
(@v1.4) pkg> add MixedModels#master
    Cloning git-repo `https://github.com/JuliaStats/MixedModels.jl.git`
   Updating git-repo `https://github.com/JuliaStats/MixedModels.jl.git`
  Resolving package versions...
ERROR: Unsatisfiable requirements detected for package NLopt [76087f3c]:
 NLopt [76087f3c] log:
 ├─possible versions are: 0.6.0 or uninstalled
 ├─restricted to versions 0.5 by MixedModels [ff71e718] — no versions left
 │ └─MixedModels [ff71e718] log:
 │   ├─possible versions are: 2.2.0 or uninstalled
 │   └─MixedModels [ff71e718] is fixed to version 2.2.0
 └─NLopt [76087f3c] is fixed to version 0.6.0
```
Would things work if `Project.toml` of the MixedModels.jl is updated to the NLopt v0.6.0? Currently, it is set as v0.5.0. Thanks for your help!

palday:
Still keeping NLopt 0.5 support in for now because the relevant bits of outward-facing API haven't changed. We can drop it at a later date when more people and packages have migrated.

Closes #315.


palday:
Closes #316 (which came in while I was running the tests locally!).

dmbates:
> Can we modify `.github/workflows/Tier[12].yml` to test on 1.5 in addition to 1.4?
> 
> I have been unable to find documentation describing what the valid entries for `julia-version` are.

The answer, apparently, is not as simple as changing the `julia-version:` value to `[1.4, 1.5]`.  @Nosferican do you have any idea on where to find a list of value Julia versions?

dmbates:
From https://github.com/julia-actions/setup-julia/blob/master/src/installer.ts it looks like the most recent version available for testing is 1.4 which remaps to 1.4.1.  Apparently one can also use 'nightly', which i have just tried.  It will probably fail because of the problem with non-reproducible RNG results.

It looks as though we will need to wait for 1.5.0-rc1 before being able to test on 1.5.0

palday:
why don't we get #314 merged and I'll rebase onto that, so that we can test on nightly?

dmbates:
Sounds good to me.  Please do go ahead and merge #314 and feel free to modify it beforehand if it seems that I lost the plot at some time in my changes.

palday:
Oh boy, it looks like the new RNG behavior is messing up my prized guarantee that threaded and unthreaded bootstrap return identical results. 

@dmbates My preference would be to merge this and address the bootstrap issue in a separate PR.

dmbates:
@palday Whatever you think makes sense.  I have kind of lost the plot on the RNG issues.

PR #319, opened 2020-05-15
change tests for simulate! and bootstrap for new Julia 1.5 RNG behavior 

palday:
Let's not get me started on how a given RNG's behavior can change for the same seed.


palday:
(And how much debugging code that involves both threading and an explicitly stochastic process leads to madness.)

PR #322, opened 2020-05-19
Change development version

palday:
Change the development version to unambiguously indicate that it's not part of the 2.x series (especially since we've now released 2.3).


palday:
The test failure on nightly is another RNG-related issue that comes up, well, stochastically.

PR #323, opened 2020-05-19
remove RNG from cholesky test

palday:
The tests for the Cholesky decomposition and pivot don't actually require us to meet any distributional assumptions, so I replaced the draws from the RNG with simple ranges.


palday:
Sometimes this test still fails and I have no idea why. I'm not able to replicate on nightly on my own machine and there is no longer any stochastic element to it, so I'm not sure what's occasionally causing us to get a different pivot.  Given that we're using factors and thus all elements of XtX are either precisely 0 or 1, we shouldn't be near machine precision, so that's not where the factorization is breaking down. I'm wondering if there is yet another BLAS difference we're running into and thus we only see it when we get "unlucky" with the testing host.

PR #326, opened 2020-05-29
test on Julia 1.5 in Future workflow

palday:
Now that there's a Julia 1.5 beta, let's see if we can test on it in GitHub Actions.


palday:
So 1.5 apparently isn't yet available as a GitHub action. :(

Also, the Cholesky strikes again.

dmbates:
Closed by #327

PR #327, opened 2020-05-29
restore the use of views not @simd workarounds

dmbates:
- replace and remove code to by-pass creation of views in some linear algebra operations, especially those related to `UniformBlockDiagonal` arrays and some pre- and post-multiplications by a repeated block diagonal array.
- prior to Julia v1.5.0 expressing these operations using views required a lot of allocations and garbage collection.  v1.5.0 does away with these allocations.  Many thanks to those who made this happen.

### DIFF
--- a/src/remat.jl
+++ b/src/remat.jl
@@ -389,32 +389,15 @@ function reweight!(A::ReMat, sqrtwts::Vector)
     A
 end
 
-rmulΛ!(A::M, B::ReMat{T,1}) where{M<:AbstractMatrix{T}} where{T} = rmul!(A, first(B.λ))
-
-"""
-    _rmullowertriblk!(A::AbstractMatrix{T}, Bd::Matrix, coloffset, S)
-
-Multiply in-place a block of `S` columns of `A` starting at `coloffset+1` on the right by the lower triangular `Bd`.
-"""
-function _rmullowertriblk!(A::AbstractMatrix, Bd::Matrix, coloffset, S)
-    @inbounds @simd for i = 1:size(A, 1)
-        for j = 1:S
-            Aij = A[i, j + coloffset] * Bd[j, j]
-            for k = j + 1:S
-                Aij += A[i, k + coloffset] * Bd[k, j]
-            end
-            A[i, j + coloffset] = Aij
-        end
-    end
-end
-
+rmulΛ!(A::M, B::ReMat{T,1}) where{M<:AbstractMatrix{T}} where{T} = rmul!(A, only(B.λ))
 function rmulΛ!(A::M, B::ReMat{T,S}) where {M<:AbstractMatrix{T}} where {T,S}
     m, n = size(A)
     q, r = divrem(n, S)
     iszero(r) || throw(DimensionMismatch("size(A, 2) is not a multiple of block size"))
-    Bd = B.λ.data
+    λ = B.λ
     for k = 1:q
-        _rmullowertriblk!(A, Bd, (k - 1) * S, S)
+        coloffset = (k - 1) * S
+        rmul!(view(A, :, coloffset+1:coloffset+S), λ)
     end
     A
 end
@@ -458,40 +441,14 @@ function scaleinflate!(Ljj::Matrix{T}, Λj::ReMat{T,1}) where {T}
     Ljj
 end
 
-"""
-    _scaleinflateface!(A::Array, λ::Matrix, f, S)
-
-Replace the `f`'th `S×S` face of `A` by `λ'A[:,:,f]*λ + I` where `λ` is `S×S` lower triangular
-"""
-function _scaleinflateface!(A::Array, λ::Matrix, f, S)
-    @inbounds @simd for i = 1:S                   # rmul!(A[:,:,f], λ)
-        for j = 1:S
-            Aij = A[i, j, f] * λ[j, j]
-            for k = j + 1:S
-                Aij += A[i, k, f] * λ[k, j]
-            end
-            A[i, j, f] = Aij
-        end
-    end
-    for j = 1:S                  # lmul!(λ, A[:,:,f])
-        for i = 1:S
-            Aij = λ[i, i]'A[i, j, f]
-            for k = i + 1:S
-                Aij += λ[k, i]'A[k, j, f]
-            end
-            A[i, j, f] = Aij
-        end
-        A[j, j, f] += true      # inflate diagonal
-    end
-end
-
 function scaleinflate!(Ljj::UniformBlockDiagonal{T}, Λj::ReMat{T,S}) where {T,S}
-    A = Ljj.data
-    m, n, l = size(A)
-    m == n == S || throw(DimensionMismatch())
-    λ = Λj.λ.data
-    for f in 1:l
-        _scaleinflateface!(A, λ, f, S)
+    λ = Λj.λ
+    dind = diagind(S, S)
+    for f in Ljj.facevec
+        lmul!(λ', rmul!(f, λ))
+        for i in dind
+            f[i] += one(T)
+        end
     end
     Ljj
 end
--- a/src/remat.jl
+++ b/src/remat.jl
@@ -444,10 +444,12 @@ end
 function scaleinflate!(Ljj::UniformBlockDiagonal{T}, Λj::ReMat{T,S}) where {T,S}
     λ = Λj.λ
     dind = diagind(S, S)
-    for f in Ljj.facevec
+    Ldat = Ljj.data
+    for k in axes(Ldat, 3)
+        f = view(Ldat, :, :, k)
         lmul!(λ', rmul!(f, λ))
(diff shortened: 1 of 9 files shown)


test/pls.jl
```diff
@@ -327,8 +327,8 @@ end
 
     @testset "zerocorr PCA" begin
         @test length(fmnc.rePCA) == 1
-        @test fmnc.rePCA.subj == [0.5, 1.0]
-        @test MixedModels.PCA(fmnc).subj.loadings == I(2)
+        @test fmnc.rePCA.subj ≈ [0.5, 1.0]
+        @test_broken MixedModels.PCA(fmnc).subj.loadings ≈ I(2)
```
dmbates, line 331: @palday Could you take a look at this, please?  `git blame` thinks you added this testset and I am not sure the results are well-defined but I don't know enough about what you wanted to have happen.

palday: I think the thought process there was "zerocorr models should already have an orthogonal variance-covariance matrix, so the PC rotation should be the identity", but it occurs to me that that may not be true.

src/remat.jl
```diff
@@ -458,40 +441,14 @@ function scaleinflate!(Ljj::Matrix{T}, Λj::ReMat{T,1}) where {T}
...
-    m == n == S || throw(DimensionMismatch())
-    λ = Λj.λ.data
-    for f in 1:l
-        _scaleinflateface!(A, λ, f, S)
+    λ = Λj.λ
+    dind = diagind(S, S)
+    for f in Ljj.facevec
+        lmul!(λ', rmul!(f, λ))
+        for i in dind
+            f[i] += one(T)
```
palday, line 450: Adding the comment back in will help me remember what's going on when I look at this later :)
```suggestion
            f[i] += one(T) # inflate diagonal
```

palday:
It looks like the test data isn't installing and that's why the CIs are failing. I'll see what I can do.

palday:
Hmmm, it looks like the xz format might not work:

```julia
gzip: stdin: not in gzip format
tar: Child returned status 1
ERROR: LoadError: InitError: Unable to automatically install 'TestData' 
```

palday:
So, Julia itself [does support using `xz`](https://github.com/JuliaLang/Pkg.jl/blob/release-1.4/src/PlatformEngines.jl#L331-L337) (assuming the user has it installed -- we still need to check what happens on Windows, where it might not be there by default), but it's sensitive to file extension. Now it's time to check what happens to the file extension when auto-downloading....

palday:
And the problem comes back to `wget`, which without extra flags defaults to taking the file name from the URL and not the HTTP headers (for which you need the `--content-disposition` flag). I think I found the necessary lines in `Pkg` to add in the extra options, but this won't change before Julia 1.6, so my recommendation is that we use `gz` for now (which is the default compression method and thus works when the file extension isn't available).

palday:
So I think there might be some less than ideal code in the relevant part of `Pkg`, but I also think that I'm messing up something here. I'll keep trying to fix this, but you might get a bunch of notifications while I work on it....

palday:
Note to self and to explain a bit what's going on:
1. Because different tar versions can't be relied on to automatically determine which (de)compression algorithm to do, Julia uses the file extension to do this.
1. If the extension isn't `xz` or `bz2`, then Julia defaults to `gzip`.
1. Julia calls `wget` or `curl` on different systems, but the default options for `wget` determine the filename based on the URL and not the metadata. The OSF URL doesn't include the file extension, so we get the default `gzip` behavior. There are `wget` flags to trust the metadata, but there may be security issues with that. Still, I can make a PR and see what happens.
1. There are better OSF URLs that include a file extension via HTTP `?` (`GET` or `POST`, can never remember which is which) syntax. But this seems to interact weirdly with the shell call to `wget` or `curl` on the CIs. 

I suspect the last point could be fixed via appropriate escaping, but I don't know whether that would work well across platforms. I'm running into other difficulties when testing locally -- I seem to have a stale cache somewhere that restarting Julia doesn't fix (I get "expected hash doesn't match"), but the computed hash matches `Artifact.toml` and I can't find a reference to the "expected" anywhere in the project....).

dmbates:
I have been testing it as well and am frustrated trying to find out how to force a new version to be downloaded.  I create sequences like this
```
julia> using Pkg.Artifacts

julia> artifact"TestData"
"/home/bates/.julia/artifacts/6614babc98083a65c25b7ec1ef2e38b345de65c9"

julia> using MixedModels

julia> MixedModels.TestData
"/home/bates/.julia/artifacts/d3fe9d4e2e432b1ec1d8f0a80cf778a94f81e7c6"
```
so I don't know what is happening in the call to the __init__ function for the package.

test/pls.jl
palday: Ah, of course, it's only true up to a permutation of the columns.

palday:
For now, I would say just leave the artifacts as they are now and I'll see if I can an `xz` download working elsewhere. As long as we don't add additional testdata, the git-tree-hash won't change and so we can change the download URL and hash at a later point without negative impact. 

If the CIs are happy with the new artifact, then that suggests it should work on users' systems...

dmbates:
I must admit that I am well, frustrated, that I have no idea where the information on these downloads comes from.  I, perhaps naively, expected that the information on artifacts was the information we stored in `Artifacts.toml` but apparently not.
```julia
julia> using MixedModels
Downloading artifact: TestData
curl: (22) The requested URL returned error: 404 Not Found                    
Downloading artifact: TestData
######################################################################## 100.0%#=#=-#  #                                                                ######################################################################## 100.0%
┌ Error: Hash Mismatch!
│   Expected sha256:   7bb283955521bd513cdd0a8c57e8440b3d40450ba338769806cf4fba0b9662f4
│   Calculated sha256: e85dc5f76251707034ff1e086775349650d81ddebabba67c0f065143c5cee134
ERROR: InitError: Unable to automatically install 'TestData' from '/home/bates/.julia/dev/MixedModels/Artifacts.toml'
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] ensure_artifact_installed(::String, ::Dict{String,Any}, ::String; platform::Pkg.BinaryPlatforms.Platform, verbose::Bool, quiet_download::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Artifacts.jl:894
 [3] do_artifact_str(::String, ::Dict{String,Any}, ::String, ::Module) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Artifacts.jl:1023
 [4] #invokelatest#1 at ./essentials.jl:710 [inlined]
 [5] invokelatest at ./essentials.jl:709 [inlined]
 [6] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Artifacts.jl:1064 [inlined]
 [7] __init__() at /home/bates/.julia/dev/MixedModels/src/MixedModels.jl:134
 [8] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:697
 [9] _require_search_from_serialized(::Base.PkgId, ::String) at ./loading.jl:782
 [10] _require(::Base.PkgId) at ./loading.jl:1007
 [11] require(::Base.PkgId) at ./loading.jl:928
 [12] require(::Module, ::Symbol) at ./loading.jl:923
during initialization of module MixedModels

shell> cat Artifacts.toml
[TestData]
git-tree-sha1 = "6614babc98083a65c25b7ec1ef2e38b345de65c9"
lazy = true

    [[TestData.download]]
    sha256 = "92e1d57966e6d4716d9925450b4a100d923e3df4c40060d110860cf968ec6559"
    # when updating this, making sure to include the version (change to new) and
    # displayName (don't change) because the default link always points to the newest
    # (and so will not match the SHA for users not yet on the most recent version)
    # and Julia invokes wget without using HTTP metadata, so we need the link
    # to end with the right extension unless we want to fall back to gzip compression
    url = "https://osf.io/pcjk6/download"
```
I have no idea where that expected sha256 nor the observed sha256 came from.

dmbates:
Good news.  I think I am beginning to understand.  You need to make a change in the package code to force it to precompile if you want the changes in `Artifacts.toml` to take effect.  In retrospect it makes sense.  So I am able to function again.  Sorry for the noise.

palday:
> You need to make a change in the package code to force it to precompile if you want the changes in Artifacts.toml to take effect.

AHHH, that actually helps me a lot.

dmbates:
The purpose of removing the `facevec` field in the `UniformBlockDiagonal` struct is to make the code a bit cleaner.  As of v1.5.0 there is no major difference in the speed - in fact it is slightly faster in my timings.  Pre-1.5.0 versions of Julia may see a significant difference on some benchmarks.

dmbates:
@palday I think this is ready to go.  Please pull the trigger if you agree.

PR #329, opened 2020-06-25
Add and use parmap field.

dmbates:
The `parmap` field in a `LinearMixedModel` is a vector of triples of the form `(term, row, column)` giving the mapping from the elements of the θ parameter vector to the positions in the λ matrices for each random-effects term.  Storing this as a separate field allows for cleaner logic in methods for `nθ`, `setθ!`, and `getθ!`.

### DIFF
--- a/src/femat.jl
+++ b/src/femat.jl
@@ -83,5 +83,3 @@ LinearAlgebra.mul!(C, adjA::Adjoint{T,<:FeMat{T}}, B::FeMat{T}) where {T} =
 Does `A` have full column rank?
 """
 isfullrank(A::FeMat) = A.rank == length(A.piv)
-
-nθ(A::FeMat) = 0
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -8,6 +8,7 @@ Linear mixed-effects model representation
 * `formula`: the formula for the model
 * `allterms`: a vector of random-effects terms, the fixed-effects terms and the response
 * `sqrtwts`: vector of square roots of the case weights.  Can be empty.
+* `parmap` : Vector{NTuple{3,Int}} of (block, row, column) mapping of θ to λ
 * `A`: an `nt × nt` symmetric `BlockMatrix` of matrices representing `hcat(Z,X,y)'hcat(Z,X,y)`
 * `L`: a `nt × nt` `BlockMatrix` - the lower Cholesky factor of `Λ'AΛ+I`
 * `optsum`: an [`OptSummary`](@ref) object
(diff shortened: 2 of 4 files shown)


src/linearmixedmodel.jl
```diff
@@ -920,6 +943,9 @@ function zerocorr!(m::LinearMixedModel{T}, trmns) where {T}
             zerocorr!(trm)
         end
     end
+    newparmap = mkparmap(reterms)
+    copyto!(m.parmap, newparmap)
+    resize!(m.parmap, length(newparmap))
```
palday, line 948: This dance is necessary because `LinearMixedModel` is an immutable struct, right? (Not suggesting a change, just double checking my understanding.)

src/femat.jl
```diff
@@ -83,5 +83,3 @@ LinearAlgebra.mul!(C, adjA::Adjoint{T,<:FeMat{T}}, B::FeMat{T}) where {T} =
 Does `A` have full column rank?
 """
 isfullrank(A::FeMat) = A.rank == length(A.piv)
-
-nθ(A::FeMat) = 0
```
palday, line 87: Not arguing for keeping this around, but why remove it? (Or rather what was the original use case that this was included? `allterms` stuff?)

src/linearmixedmodel.jl
dmbates: Exactly

src/femat.jl
dmbates: I guess there is no problem with keeping it around.  You are correct that it was added in the conversion to using `allterms` so that nθ of a model could be written as the sum of nθ applied to the `allterms` field.

My feeling is that it is better to avoid creating specific methods if we know we won't use them because deleting it cuts down on clutter and on those reading the code thinking "why is this here?".  It would probably be possible to drop all the methods for nθ but I haven't checked.

dmbates:
@palday Thank you for reviewing this given everything else on your plate right now.

dmbates:
@palday Did you notice that the test failure under 1.5.0-beta1 is in the `missingcells` test where `ch.piv` comes back as 1:100?

src/femat.jl
palday: Let's go ahead and cut it -- I think we're slowly converging on a somewhat more refined internal structure and trimming methods makes it easier to keep an overview of everything.

palday:
@dmbates Yeah, I saw that. So I'm happy to go ahead and merge this if you're done tinkering on this PR. :)

dmbates:
Yes, I'm done tinkering with this so please do squash and merge.  I am still tinkering with #331 however.

PR #331, opened 2020-06-25
Rank update

### LINKED ISSUE #290: Error When Adding Random Slopes
I'm trying to fit a model with two grouping factors and one covariate. One has many levels (more than 20,000) and one has only a few hundred. There are ~70,000 data. I'm getting an error whenever I add a random slope of the covariate by the factor with 200 levels. The models fits without issues if I exclude random slopes.

```
ERROR: MethodError: rmulΛ!(::BlockedSparse{Float64,2,1}, ::ReMat{Float64,1}) is ambiguous. Candidates:
  rmulΛ!(A::M, B::ReMat{T,1}) where {T, M<:AbstractArray{T,2}} in MixedModels at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/remat.jl:391
  rmulΛ!(A::BlockedSparse{T,S,P}, B::ReMat{T,P}) where {T, S, P} in MixedModels at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/remat.jl:405
Possible fix, define
  rmulΛ!(::Union{M<:BlockedSparse{T,S,1}, M<:BlockedSparse{T,S,1}, M<:(BlockedSparse{T,S,1} where S)}, ::Union{ReMat{T,1}, ReMat{T,1}, ReMat{T,1}})

Stacktrace:
 [1] updateL!(::LinearMixedModel{Float64}) at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/linearmixedmodel.jl:678
 [2] #fit!#38(::Bool, ::Bool, ::typeof(fit!), ::LinearMixedModel{Float64}) at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/linearmixedmodel.jl:251
 [3] (::StatsBase.var"#kw##fit!")(::NamedTuple{(:verbose, :REML),Tuple{Bool,Bool}}, ::typeof(fit!), ::LinearMixedModel{Float64}) at ./none:0
 [4] #fit#35(::Array{Any,1}, ::Dict{Symbol,Any}, ::Bool, ::Bool, ::typeof(fit), ::Type{LinearMixedModel}, ::StatsModels.FormulaTerm{StatsModels.Term,Tuple{StatsModels.Term,StatsModels.FunctionTerm{typeof(|),var"#25#27",(:Item,)},StatsModels.FunctionTerm{typeof(|),var"#26#28",(:FreqZipfUS, :Subject)}}}, ::NamedTuple{(:RT, :FreqZipfUS, :Subject, :Item),Tuple{CSV.Column{Float64,Float64},CSV.Column{Union{Missing, Float64},Union{Missing, Float64}},CategoricalArray{Int64,1,UInt32,Int64,CategoricalValue{Int64,UInt32},Union{}},CategoricalArray{String,1,UInt32,String,CategoricalString{UInt32},Union{}}}}) at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/linearmixedmodel.jl:116
 [5] (::StatsBase.var"#kw##fit")(::NamedTuple{(:wts, :contrasts, :verbose, :REML),Tuple{Array{Any,1},Dict{Symbol,Any},Bool,Bool}}, ::typeof(fit), ::Type{LinearMixedModel}, ::StatsModels.FormulaTerm{StatsModels.Term,Tuple{StatsModels.Term,StatsModels.FunctionTerm{typeof(|),var"#25#27",(:Item,)},StatsModels.FunctionTerm{typeof(|),var"#26#28",(:FreqZipfUS, :Subject)}}}, ::NamedTuple{(:RT, :FreqZipfUS, :Subject, :Item),Tuple{CSV.Column{Float64,Float64},CSV.Column{Union{Missing, Float64},Union{Missing, Float64}},CategoricalArray{Int64,1,UInt32,Int64,CategoricalValue{Int64,UInt32},Union{}},CategoricalArray{String,1,UInt32,String,CategoricalString{UInt32},Union{}}}}) at ./none:0
 [6] #fit#34(::Array{Any,1}, ::Dict{Symbol,Any}, ::Bool, ::Bool, ::typeof(fit), ::Type{LinearMixedModel}, ::StatsModels.FormulaTerm{StatsModels.Term,Tuple{StatsModels.Term,StatsModels.FunctionTerm{typeof(|),var"#25#27",(:Item,)},StatsModels.FunctionTerm{typeof(|),var"#26#28",(:FreqZipfUS, :Subject)}}}, ::DataFrame) at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/linearmixedmodel.jl:110
 [7] fit(::Type{LinearMixedModel}, ::StatsModels.FormulaTerm{StatsModels.Term,Tuple{StatsModels.Term,StatsModels.FunctionTerm{typeof(|),var"#25#27",(:Item,)},StatsModels.FunctionTerm{typeof(|),var"#26#28",(:FreqZipfUS, :Subject)}}}, ::DataFrame) at /home/bryor/.juliapro/JuliaPro_v1.3.1-1/packages/MixedModels/qyKOf/src/linearmixedmodel.jl:110
 [8] top-level scope at REPL[32]:1
```

dmbates:
Clean up `rankUpdate!` methods.  In particular use Hermitian{T, Diagonal{T, Vector{T}}} instead of branching on the diagonal case in `updateL!`.  We want to use method dispatch instead of `if` statements that I had in the code.  This should be rebased on master after #329 is merged.

Closes #290.

### DIFF
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -11,58 +11,25 @@ The order of the arguments
 """
 function rankUpdate! end
 
-function rankUpdate!(
-    C::AbstractMatrix{T},
-    A::AbstractMatrix{T},
-    α=true,
-    β=true) where {T}
-
-    m, n = size(A)
-    m == size(C, 2) || throw(DimensionMismatch(""))
-    @info "it is surprising that this function is called - please report a use case as a MixedModels issue"
-    @warn "using generic method, this will be slower than usual"
-
-    A = A'A
-
-    for (i, el) in enumerate(C)
-        C[i] = β * el + α * A[i]
-    end
-    C
-end
-
-function rankUpdate!(
-    C::HermOrSym{T,S},
-    a::StridedVector{T},
-    α = true,
-) where {T<:BlasReal,S<:StridedMatrix}
+function rankUpdate!(C::HermOrSym{T,S}, a::StridedVector{T}, α, β) where {T,S}
+    isone(β) || throw(ArgumentError("isone(β) is false"))
     BLAS.syr!(C.uplo, T(α), a, C.data)
     C  ## to ensure that the return value is HermOrSym
 end
 
-function rankUpdate!(
-    C::HermOrSym{T,S},
-    A::StridedMatrix{T},
-    α = true,
-    β = true,
-) where {T<:BlasReal,S<:StridedMatrix}
+function rankUpdate!(C::HermOrSym{T,S}, A::StridedMatrix{T}, α, β) where {T,S}
     BLAS.syrk!(C.uplo, 'N', T(α), A, T(β), C.data)
     C
 end
 
-function rankUpdate!(
-    C::HermOrSym{T,Matrix{T}},
-    A::SparseMatrixCSC{T},
-    α = true,
-    β = true,
-) where {T}
-    m, n = size(A)
-    m == size(C, 2) || throw(DimensionMismatch(""))
+function rankUpdate!(C::HermOrSym{T,S}, A::SparseMatrixCSC{T}, α, β) where {T,S}
+    A.m == size(C, 2) || throw(DimensionMismatch())
     C.uplo == 'L' || throw(ArgumentError("C.uplo must be 'L'"))
     Cd = C.data
     isone(β) || rmul!(LowerTriangular(Cd), β)
     rv = rowvals(A)
     nz = nonzeros(A)
-    @inbounds for jj = 1:n
+    @inbounds for jj = 1:A.n
         rangejj = nzrange(A, jj)
         lenrngjj = length(rangejj)
         for (k, j) in enumerate(rangejj)
@@ -77,43 +44,38 @@ function rankUpdate!(
     C
 end
 
-rankUpdate!(C::HermOrSym, A::BlockedSparse, α = true, β = true) =
+function rankUpdate!(C::HermOrSym, A::BlockedSparse, α, β)
     rankUpdate!(C, sparse(A), α, β)
+end
 
-function rankUpdate!(
-    C::Diagonal{T},
-    A::SparseMatrixCSC{T},
-    α = true,
-    β = true,
-) where {T<:Number}
-    m, n = size(A)
-    dd = C.diag
-    length(dd) == m || throw(DimensionMismatch(""))
+function rankUpdate!(C::HermOrSym{T,Diagonal{T}}, A::SparseMatrixCSC{T}, α, β) where {T}
+    dd = C.data.diag
+    A.m == length(dd) || throw(DimensionMismatch())
     isone(β) || rmul!(dd, β)
     nz = nonzeros(A)
     rv = rowvals(A)
-    @inbounds for j = 1:n
-        nzr = nzrange(A, j)
-        if !isempty(nzr)
-            length(nzr) == 1 || throw(ArgumentError("A*A' has off-diagonal elements"))
-            k = nzr[1]
-            dd[rv[k]] += α * abs2(nz[k])
-        end
+    @inbounds for j = 1:A.n
+        k = only(nzrange(A, j))
+        dd[rv[k]] += α * abs2(nz[k])
     end
     C
 end
 
-rankUpdate!(C::Diagonal{T}, A::BlockedSparse{T}, α = true, β = true) where {T<:Number} =
+function rankUpdate!(C::HermOrSym{T,Diagonal{T}}, A::BlockedSparse{T}, α, β) where {T}
     rankUpdate!(C, sparse(A), α, β)
+end
 
 function rankUpdate!(
-    C::HermOrSym{T,UniformBlockDiagonal{T}},
+    C::HermOrSym{T,UniformBlockDiagonal{T}}, 
     A::BlockedSparse{T,S},
-    α = true,
+    α,
+    β,
 ) where {T,S}
     Ac = A.cscmat
     cp = Ac.colptr
-    all(diff(cp) .== S) || throw(ArgumentError("Each column of A must contain exactly S nonzeros"))
+    if any(diff(cp) .≠ S)
+        throw(ArgumentError("Each column of A must contain exactly S nonzeros"))
+    end
     Cdat = C.data.data
     j, k, l = size(Cdat)
     S == j == k && div(Ac.m, S) == l ||
@@ -127,36 +89,21 @@ function rankUpdate!(
     C
 end
 
-#=
-rankUpdate!(C::HermOrSym{T,Matrix{T}}, A::BlockedSparse{T}, α = true) where {T} =
-    rankUpdate!(C, A.cscmat, α)
-=#
+function rankUpdate!(C::HermOrSym{T,Diagonal{T}}, A::Diagonal{T}, α, β) where {T}
+    Cdiag = C.data.diag
+    if length(Cdiag) ≠ length(A.diag)
+        throw(DimensionMismatch("length(C.data.diag) ≠ length(A.diag)"))
+    end
 
-function rankUpdate!(
-    C::Diagonal{T,S},
-    A::Diagonal{T,S},
-    α::Number = true,
-    β::Number = true,
-) where {T,S}
-    length(C.diag) == length(A.diag) || throw(DimensionMismatch("length(C.diag) ≠ length(A.diag)"))
-
-    C.diag .= β .* C.diag .+ α .* abs2.(A.diag)
+    Cdiag = β .* Cdiag .+ α .* abs2.(A.diag)
     C
 end
 
-
-function rankUpdate!(
-    C::HermOrSym{T,Matrix{T}},
-    A::Diagonal{T},
-    α::Number = true,
-    β::Number = true,
-) where {T}
-    m, n = size(A)
-    m == size(C, 2) || throw(DimensionMismatch(""))
-    adiag = A.diag
-    cdata = C.data
-    for (i, a) in zip(diagind(cdata), adiag)
-        cdata[i] = β * cdata[i] + α * abs2(a)
+function rankUpdate!(C::HermOrSym{T,Matrix{T}}, A::Diagonal{T}, α, β) where {T}
+    Adiag, Cdata = A.diag, C.data
+    length(Adiag) == size(C, 2) || throw(DimensionMismatch())
+    for (i, a) in zip(diagind(Cdata), Adiag)
+        Cdata[i] = β * Cdata[i] + α * abs2(a)
     end
     C
 end
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -11,6 +11,10 @@ The order of the arguments
 """
 function rankUpdate! end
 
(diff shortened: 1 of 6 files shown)


src/linalg/rankUpdate.jl
```diff
@@ -77,43 +44,39 @@ function rankUpdate!(
...
-    nz = nonzeros(A)
-    rv = rowvals(A)
-    @inbounds for j = 1:n
-        nzr = nzrange(A, j)
-        if !isempty(nzr)
-            length(nzr) == 1 || throw(ArgumentError("A*A' has off-diagonal elements"))
-            k = nzr[1]
-            dd[rv[k]] += α * abs2(nz[k])
-        end
+    all(isone.(diff(A.colptr))) || throw(ArgumentError("Columns of A must have exactly 1 nonzero"))
```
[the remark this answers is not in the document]
palday: Ah and I get to the next method and see that the new error message is more consistent with the error message for a different optimization.

src/remat.jl
```diff
@@ -43,7 +43,7 @@ function _amalgamate(reterms::Vector, T::Type)
     value = ReMat{T}[]
     for (f, inds) in factordict
         if isone(length(inds))
-            push!(value, reterms[first(inds)])
+            push!(value, reterms[only(inds)])
```
palday, line 46: yay for `only()` :)

src/remat.jl
```diff
@@ -374,8 +376,10 @@ function *(adjA::Adjoint{T,<:ReMat{T,S}}, B::ReMat{T,P}) where {T,S,P}
     BlockedSparse{T,S,P}(cscmat, reshape(cscmat.nzval, S, :), cscmat.colptr[1:P:(cscmat.n + 1)])
 end
 
-PCA(A::ReMat{T,1}; corr::Bool=true) where {T} =
-    PCA(corr ? reshape(abs.(vec(A.λ)),1,1) : reshape(ones(T,1),1,1), corr=corr)
+function PCA(A::ReMat{T,1}; corr::Bool=true) where {T}
+    val = ones(T, 1, 1)
+    PCA(corr ? val : abs(only(A.λ)) * val, corr=corr)
+end
```
palday, line 382: this helps readability nicely

test/linalg.jl
```diff
@@ -52,6 +52,7 @@ end
     @test loglikelihood(fit!(wm1)) ≈ loglikelihood(m1)
 end
 
+#=  I don't see this testset as meaningful b/c diagonal A does not occur after amalgamation of ReMat's for the same grouping factor - D.B.
```
palday, line 55: I'm trying to think of a case where diagonal A could occur. Perhaps in one of the custom model types some of the ZiF spinoff projects are working on? Maybe this is something re-visit in a few weeks to months and see if we can further trim.

src/linalg/rankUpdate.jl
```diff
@@ -126,37 +89,23 @@ function rankUpdate!(
...
-function rankUpdate!(
-    C::Diagonal{T,S},
-    A::Diagonal{T,S},
-    α::Number = true,
-    β::Number = true,
-) where {T,S}
-    length(C.diag) == length(A.diag) || throw(DimensionMismatch("length(C.diag) ≠ length(A.diag)"))
-
-    C.diag .= β .* C.diag .+ α .* abs2.(A.diag)
+    Cdiag = β .* Cdiag .+ α .* abs2.(A.diag)
```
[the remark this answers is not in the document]
palday: Ah, of course, this is in the commented-out function. :facepalm:

palday, line 99:
```suggestion
    Cdiag .= β .* Cdiag .+ α .* abs2.(A.diag)
```

dmbates:
> If I understand everything correctly, then it looks pretty good. I have one question about assignment vs. dot assignment in a single method that needs attention, everything else is small.
> 
> The two big changes that I didn't comment on are:
> 
>     1. No default values for alpha and beta in the rank update methods. I think this is fine, but this is definitely a breaking change (admittedly on non-exported methods), so it's a good thing to get in now before the big release. Is this also in line with the related methods in LinearAlgebra?
> 
>     2. Removal of a catch-all `rankUpdate!` method. I'm wondering if we should still have that in there. Or maybe even something 'dumber', such as
> 
> 
> ```julia
> function rankUpdate(A::Any, B::Any, alpha, beta) 
>    error("We haven't implemented a method for $(typeof(A)), $(typeof(B)). Please file an issue on GitHub")
> end
> ```

I don't see the benefit of adding a "catch-all" method to generate an error.  Not having a method has the same effect, doesn't it?

dmbates:
> If I understand everything correctly, then it looks pretty good. I have one question about assignment vs. dot assignment in a single method that needs attention, everything else is small.
> 
> The two big changes that I didn't comment on are:
> 
>     1. No default values for alpha and beta in the rank update methods. I think this is fine, but this is definitely a breaking change (admittedly on non-exported methods), so it's a good thing to get in now before the big release. Is this also in line with the related methods in LinearAlgebra?

I felt that this generic would be internal to the package and I dropped the default values for alpha and beta because there is only one call to generic in the package.  It seemed it was easier to specify the values there than to add defaults for all the methods.

palday:
> > If I understand everything correctly, then it looks pretty good. I have one question about assignment vs. dot assignment in a single method that needs attention, everything else is small.
> > The two big changes that I didn't comment on are:
> > ```
> > 1. No default values for alpha and beta in the rank update methods. I think this is fine, but this is definitely a breaking change (admittedly on non-exported methods), so it's a good thing to get in now before the big release. Is this also in line with the related methods in LinearAlgebra?
> > 
> > 2. Removal of a catch-all `rankUpdate!` method. I'm wondering if we should still have that in there. Or maybe even something 'dumber', such as
> > ```
> > 
> > 
> > ```julia
> > function rankUpdate(A::Any, B::Any, alpha, beta) 
> >    error("We haven't implemented a method for $(typeof(A)), $(typeof(B)). Please file an issue on GitHub")
> > end
> > ```
> 
> I don't see the benefit of adding a "catch-all" method to generate an error. Not having a method has the same effect, doesn't it?

For us, yes. But for the average user, a `MethodError` followed by a stacktrace is big and scary and potentially looks like a bug. An error with "we haven't implemented this yet" seems less scary. But then again, both will generate huge stacktraces and that's probably the hardest thing for normal users anyway.

Ultimately, I'm neutral. Do we have a "normal" person we can ask?

test/linalg.jl
dmbates: It is best to think of how `rankUpdate!` gets called.  After pre- and post-multiplication by Lambda and inflation of the diagonal the `rankUpdate!` methods get called with `C` being one of the diagonal blocks and `A` being a block to the left of `C`.  So `C` is `m.L[Block(2,2)]` or higher because there are no blocks to the left of `m.L[Block(1,1)]`.  The cases that matter are when there is more than one grouping factor for the random effects.  If `A` were to be diagonal then it would need to be generated from two scalar random-effects terms, as in `(1|G) + (1|H)` and the number of levels of `G` and `H` would need to be equal and the levels would match.  You can get that from `(1|G) + (0+x|G)` but that is now collapsed to a single r.e. term either through `zerocorr(1+x|G)` or by amalgamation of the random effects terms.  So I don't think there will be real-life cases where `A` is diagonal.  `C` is always a square matrix and either `Diagonal` or `UniformBlockDiagonal` or, through fill-in, `Dense`.   The only special cases for `A` are `Dense`, `SparseMatrixCSC` or, in a new branch that I haven't pushed yet, `KHotColumn`.  The last one is a special type of sparse matrix generated by nested grouping factors for vector-valued random effects where each column of `A` has exactly `K` adjacent non-zeros.

dmbates:
> > > If I understand everything correctly, then it looks pretty good. I have one question about assignment vs. dot assignment in a single method that needs attention, everything else is small.
> > > The two big changes that I didn't comment on are:
> > > ```
> > > 1. No default values for alpha and beta in the rank update methods. I think this is fine, but this is definitely a breaking change (admittedly on non-exported methods), so it's a good thing to get in now before the big release. Is this also in line with the related methods in LinearAlgebra?
> > > 
> > > 2. Removal of a catch-all `rankUpdate!` method. I'm wondering if we should still have that in there. Or maybe even something 'dumber', such as
> > > ```
> > > 
> > > 
> > > ```julia
> > > function rankUpdate(A::Any, B::Any, alpha, beta) 
> > >    error("We haven't implemented a method for $(typeof(A)), $(typeof(B)). Please file an issue on GitHub")
> > > end
> > > ```
> > 
> > 
> > I don't see the benefit of adding a "catch-all" method to generate an error. Not having a method has the same effect, doesn't it?
> 
> For us, yes. But for the average user, a `MethodError` followed by a stacktrace is big and scary and potentially looks like a bug. An error with "we haven't implemented this yet" seems less scary. But then again, both will generate huge stacktraces and that's probably the hardest thing for normal users anyway.
> 
> Ultimately, I'm neutral. Do we have a "normal" person we can ask?

I would value the opinions of either @christinabergmann or @debruine

dmbates:
> > > > If I understand everything correctly, then it looks pretty good. I have one question about assignment vs. dot assignment in a single method that needs attention, everything else is small.
> > > > The two big changes that I didn't comment on are:
> > > > ```
> > > > 1. No default values for alpha and beta in the rank update methods. I think this is fine, but this is definitely a breaking change (admittedly on non-exported methods), so it's a good thing to get in now before the big release. Is this also in line with the related methods in LinearAlgebra?
> > > > 
> > > > 2. Removal of a catch-all `rankUpdate!` method. I'm wondering if we should still have that in there. Or maybe even something 'dumber', such as
> > > > ```
> > > > 
> > > > 
> > > > ```julia
> > > > function rankUpdate(A::Any, B::Any, alpha, beta) 
> > > >    error("We haven't implemented a method for $(typeof(A)), $(typeof(B)). Please file an issue on GitHub")
> > > > end
> > > > ```
> > > 
> > > 
> > > I don't see the benefit of adding a "catch-all" method to generate an error. Not having a method has the same effect, doesn't it?
> > 
> > 
> > For us, yes. But for the average user, a `MethodError` followed by a stacktrace is big and scary and potentially looks like a bug. An error with "we haven't implemented this yet" seems less scary. But then again, both will generate huge stacktraces and that's probably the hardest thing for normal users anyway.

In retrospect I think it is fine to go ahead with adding a fallback-error method like this.  Would you (@palday)  be willing to add a test to ensure test coverage?

palday:
@dmbates I'll take care of it. :smile:

test/linalg.jl
palday: Right, so the only way that this could occur would be if someone had two identical grouping variables so that they wouldn't be amalgamated. Until we see a legitimate use case for that, I think we can safely ignore that possibility.

Thanks for the refresher! :smile:

palday:
@dmbates If you're happy with the new error message, please squash and merge.

Issue #255, opened 2020-02-11
document fulldummy


kleinschmidt:
the `zerocorr*` functions are documented (in constructors.jmd) but `fulldummy` isn't.  Might be a  good idea to have a separate "formula extensions" page to document all the special syntax that mixed models introduces.

kleinschmidt:
Since this came up on Zulip after merging #269, it might be good to have a docs page dedicated specifically to the various ways of controlling the variance/correlation parameters in the random effects, which would cover `zerocorr`, `fulldummy`, and the difference between `(1+x|g)` and `(1|g) + (0+x|g)`.

palday:
If we're talking about magic .... does it make sense to have `(1|g) + (x|g)` detect the presence of an intercept and not do an implicit intercept?  Or more generally, remove redundant terms? (Maybe the latter is already caught be the normal formula handling and `amalgamate`.)

kleinschmidt:
Yeah I noticed that when I was playing around last night.  I'd have thought that works but it doesn't for some reason, I'll have to go back and see why.  I think it's because the default intercept behavior doesn't take into account the other terms, since they're only linked by the schema...

palday:
@kleinschmidt Do you have time/energy/capacity to do this in the next two weeks?

kleinschmidt:
If it's blocking I can.  Ping me again in a few days...

dmbates:
@palday @kleinschmidt Is the issue here documenting the `MixedModels.fulldummy` function or the `StatsModels.FullDummy` type?  If it is the function I can do that.

dmbates:
Do we have a canonical example of the use of `fulldummy`, maybe the machines data?  I can create a branch with the function documentation string and can add some text to `docs/src/constructors.md` but I should incorporate an example in the docs.

kleinschmidt:
Yes the function. But also things like zerocorr and splitting terms across multiple RES to remove correlations. Basically all the ways you can control the correlation structure you fit. 

I'm not sure about a canonical dataset but maybe whatever we have in the tests? Unless that's just made up stuff, I can't remember...

PR #338, opened 2020-07-11
make (1|x) + (y|x) equivalent to (1|x) + (0+y|x)

kleinschmidt:
After #269, a formula like `(1|x) + (y|x)` produces an error because we're
trying to add an extra intercept term to `(y|x)`, producing a non-unique
`(Intercept)` term.  This PR checks whether the schema associated with `|x` has
already seen or generated an intercept term in any earlier term before adding
one.  With this change, `(1|x) + (y|x)` is equivalent to `(1|x) + (0+y|x)`.  I
have a really hard time seeing why someone would want the behavior in those
cases to be different but maybe I just haven't thought about it hard enough

Inspired by @palday's comment in #255.

### DIFF
--- a/src/randomeffectsterm.jl
+++ b/src/randomeffectsterm.jl
@@ -60,7 +60,7 @@ function StatsModels.apply_schema(
         !StatsModels.hasintercept(lhs) &&
         !StatsModels.omitsintercept(lhs) &&
         ConstantTerm(1) ∉ schema.already &&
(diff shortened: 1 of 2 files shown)


palday:
Thanks. I think we're converging on "if you want overparameterized random effects, you must explicitly invoke `fulldummy`" which seems internally consistent and a reasonable default.

PR #339, opened 2020-07-13
add Grouping contrasts

kleinschmidt:
This is in lieu of having these defined in StatsModels.jl; as we've discussed
they may be a bit to specialized for the specific demands of grouping variables
in MixedModels without defining enough functionality to be of use in other
settings, so it makes sense to have them here until some kind of sensible
handling of grouping is incorporated into StatsModels.

The `Grouping` contrasts are a "pseudocontrast" that never defines a contrasts
matrix, just keeps track of the unique values and an "inverse index" which maps
levels to a numeric index.  This allows you to get around the `OutOfMemoryError`
that's triggered when you have a grouping variable with a very large number of
levels.  The catch is that, at the moment, you have to manually specify the
grouping contrasts when constructing the model.  I've added a guard to catch OOM
errors during schema construction which prints a warning, but we could also
throw a different error there instead to be clearer.  The downside with that
would be that if it's NOT a problem with the ranef grouping term, then they'd
just get useless advice and no useful stacktrace, so that's why I went with
printing a warning.


kleinschmidt:
based on JuliaStats/StatsModels#181

src/grouping.jl
```diff
@@ -0,0 +1,35 @@
...
+    struct Grouping <: StatsModels.AbstractContrasts end
+
+A placeholder type to indicate that a categorical variable is only used for
+grouping and not for contrasts.  When creating a [`CategoricalTerm`](@ref), this
+skips constructing the contrasts matrix which makes it robust to large numbers
+of levels, while still holding onto the vector of levels and constructing the
+level-to-index mapping (`invindex` field of the [`ContrastsMatrix`](@ref).).
+
+Note that calling `modelcols` on a `CategoricalTerm{Grouping}` is an error.
+This is intended only as a placeholder for other packages to build on.
```
palday, line 12:
```suggestion
```

(Because that was true in StatsModels, but not here.)

src/linearmixedmodel.jl
```diff
@@ -52,7 +52,15 @@ function LinearMixedModel(
...
     # missing support is in a StatsModels release
     tbl, _ = StatsModels.missing_omit(tbl, f)
-    form = apply_schema(f, schema(f, tbl, contrasts), LinearMixedModel)
+    sch = try 
+        schema(f, tbl, contrasts)
+    catch e
+        if isa(e, OutOfMemoryError)
+            @warn "Random effects grouping variables with many levels can cause out-of-memory errors.  Try manually specifying `Grouping()` contrasts for those variables."
+        end
+        rethrow(e)
```
palday, line 61: I think this is the way to go right now, but I'm thinking about whether we should automatically set `Grouping` contrasts for grouping variables.

kleinschmidt:
Yeah I agree @palday, it would be nice to be able to do this automatically. It actually may be possible now but would take some annoying fiddling; you'd have to unpack the `FunctionTerm{typeof(|)}`, or conver them to `ReTerm`s in order to figure out what the grouping variables are *before* constructing the schema, because that's when the contrasts are needed to avoid the OOM.  This would basically amount to circumventing the normal StatsModels pipeline which is why I'd proposed somewhere (JuliaStats/StatsModels.jl#154) to add an `apply_context` stage where you don't have the schema yet but you do have the context in which the formula will be used, so you could do things like create all the `ReTerm`s you need before creating the schema.

kleinschmidt:
I think at least some of the CI failures are not unrelated to this PR but have to do with how the OS handles OOM.  The MacOS one failed before too, apparently during the tests for the added functionality where I intentionally trigger an OOM error.  And the cancelled jobs might be related to this as well, but I'm not sure, although the passing tests on at least one windows build tells me that OOM doesn't always result in a canceled job.  I guess the thing to do is to remove the tests which intentionally cause an OOM because otherwise we'll always have failing CI which...defeats the purpose.

palday:
@kleinschmidt The failures on Future are a known issue. For Windows .... I'll have to look at the details of the GitHub Action / its docker image to figure out why it just cancels. For OS X: do you have a local Mac?

kleinschmidt:
I'm afraid not.

kliegl:
> I'm afraid not. @palday @kleinschmidt 

I have many local macs

kleinschmidt:
@kliegl can you check out this PR branch and run the tests?  The procedure is to

```
] dev MixedModels
] activate ~/.julia/dev/MixedModels
; git checkout dfk/grouping
] test
```

kliegl:
> @kliegl can you check out this PR branch and run the tests? The procedure is to
> 
> ```
> ] dev MixedModels
> ] activate ~/.julia/dev/MixedModels
> ; git checkout dfk/grouping
> ] test
> ```

The `;git checkout dfk/grouping` fails. Do I first have to clone this?

kleinschmidt:
ah yeah you'll need to first go to that directory, with 

```
; cd .julia/dev/MixedModels
; git fetch origin && git checkout dfk/grouping
```

kliegl:
Sorry. 
```
shell> cd .julia/dev/MixedModels
/Users/reinholdkliegl/.julia/dev/MixedModels
shell> git fetch origin && git checkout dfk/grouping
fatal: couldn't find remote ref &&
```

kliegl:
ERROR: Package MixedModels errored during testing (received signal: KILL)

[Test_2020-07-14.txt](https://github.com/JuliaStats/MixedModels.jl/files/4921309/Test_2020-07-14.txt)

kleinschmidt:
Yeah looks from that like the process is just killed when it hits the OOM tests.  I'll push a change that only tests the good path (I'm sufficiently convinced that trying to construct a 1,000,000x1,000,000 dense matrix will OOM...)

kliegl:
Just ran it again:
  Testing MixedModels tests passed 

Do you need the report?

kleinschmidt:
Looks like the CI now is failing because of a problem with the artifacts...

palday:
@kleinschmidt Yeah, I'm still sorting out the artifact link stuff. Right now, the default link always points to the newest version, but that of course doesn't match the SHA for older versions. I'll see if I can't fix that as part of #341

palday:
@kleinschmidt assuming that all the CIs (potentially except Future) pass, I think this is good to go.

PR #341, opened 2020-07-14
New version of data.tar.gz with machines dataset

dmbates:
`machines` data added to the `.tar.gz` file per @kliegl 's request.  Not sure I have followed the instructions in the comment correctly b/c I didn't really understand them.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,9 +1,9 @@
 [TestData]
-git-tree-sha1 = "6614babc98083a65c25b7ec1ef2e38b345de65c9"
+git-tree-sha1 = "8e1cb0be00e404a4a5e162bddc1d0235cc720ece"
 lazy = true
 
     [[TestData.download]]
-    sha256 = "92e1d57966e6d4716d9925450b4a100d923e3df4c40060d110860cf968ec6559"
+    sha256 = "c4b979ef93a10a6517eebf88d660500ff47ac1586e9cc897e4d10ffaa6378eb2"
     # when updating this, making sure to include the version (change to new) and
     # displayName (don't change) because the default link always points to the newest
     # (and so will not match the SHA for users not yet on the most recent version)
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -3,7 +3,12 @@
 # (so that we don't forget datasets)
 
 [TestData]
-# this is black magic to compute. Just let it error and copy over the new value.
+# compute this using
+# using Tar, Inflate, SHA
+# filename = "download?version=4" # I just used wget for the URL below and this is how it saved it
+# println("sha256: ", bytes2hex(open(sha256, filename)))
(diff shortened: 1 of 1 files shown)


palday:
There's some type of caching mechanism that happens locally that makes this very hard to test and I don't remember how we got around it last time, but that's something I would add to notes in `Artifacts.toml`.

palday:
And I have no idea how to compute the tree hash other than to let it error and copy-paste that from the error. The way I thought I found apparently doesn't work and that's using git.

Artifacts.toml
```diff
@@ -1,12 +1,21 @@
+# when generating a new artifact, add the new feather file to the datasets folder at
+# https://osf.io/a94tr/, then download that folder to generate the tarball
+# (so that we don't forget datasets)
+
 [TestData]
-git-tree-sha1 = "6614babc98083a65c25b7ec1ef2e38b345de65c9"
+# this is black magic to compute. Just let it error and copy over the new value.
+git-tree-sha1 = "4d1410cd290622e426411273ef379fe82b749ca4"
```
palday, line 7:
```julia
 ┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
│   Calculated git-tree-sha1: 4d1410cd290622e426411273ef379fe82b749ca4
```

The documentation for artifacts needs to be improved a LOT.

palday: @Nosferican Do you recall if this is related to a version number issue?

Nosferican: It's a bug in Pkg which computes different git-tree-sha1 depending on the platform if not mistaken. Last I heard, the Pkg team was planning on getting away from it or something but the current patch was to ignore that error. It is great for consistency and validating the content but it is currently broken.

palday: @Nosferican That's an additional bug -- look at the calculated SHA and the SHA in the diff. For some reason, Pkg is expected a different one than the one we have explicitly in the file  and which it is actually computing. :frowning:

Nosferican: Checking if that issue is still present. Give it a shot and copy the one it is expecting? Might help us identify if it is a version mismatch.

Nosferican: This is what I was thinking of https://github.com/JuliaLang/Pkg.jl/pull/1885.

dmbates:
> And I have no idea how to compute the tree hash other than to let it error and copy-paste that from the error. The way I thought I found apparently doesn't work and that's using git.

As I mentioned on Zulip, a recipe for how to obtain the sha values is available at https://julialang.github.io/Pkg.jl/dev/artifacts/.  It looks like
```
using Tar, Inflate, SHA

filename = "data.tar.gz"
println("sha256: ", bytes2hex(open(sha256, filename)))
println("git-tree-sha1: ", Tar.tree_hash(IOBuffer(inflate_gzip(filename))))
```

PR #345, opened 2020-07-16
increase granularity of locks

palday:
It seems that the RNG bits and not the fitting bits are the slow part of the bootstrap now, so it's time to revisit `simulate!` and the granularity of locking.

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -79,9 +79,9 @@ function parametricbootstrap(
         mod = m_threads[Threads.threadid()]
         local βsc = βsc_threads[Threads.threadid()]
         local θsc = θsc_threads[Threads.threadid()]
-        lock(rnglock)
-        mod = simulate!(rng, mod, β = β, σ = σ, θ = θ)
-        unlock(rnglock)
+        #lock(rnglock)
+        mod = simulate!(rng, mod, β = β, σ = σ, θ = θ, rnglock=rnglock)
+        #unlock(rnglock)
         refit!(mod)
         (
          objective = mod.objective,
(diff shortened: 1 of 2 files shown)


dmbates:
Have you profiled this, as in `@profview` in the VS Code REPL?  When I profile the version without threads the time spent in `simulate!` is small compared to the time spent in `refit!`.

palday:
That's interesting and good news. Getting familiar with profiling in VSCode is on my todo list for the next few days: the bootstrap is a bottleneck in things at the day job, so I would like to give this some attention in the very near future. 

My suspicion is that things have changed in the locking/threads in Julia itself in the last few iterations, but I haven't had time to investigate this thoroughly yet.

dmbates:
I think the slowdown under multiple threads is the problem of multi-threaded BLAS called by multi-threaded code.  If you fit, say
```julia
julia> rng = MersenneTwister(42);

julia> m1 = fit(MixedModel, @formula(reaction ~ 1 + days + (1+days|subj)), MixedModels.dataset(:sleepstudy));
```
run a small bootstrap to force compilation then, in VS Code, run
```julia
julia> @profview parametricbootstrap(rng, 1000, m1, use_threads=false) # adjust sample size for approx 5 s running time
```
it will open a panel with the execution times by function sorted by time spent in the function itself.  In my case the majority of the self-time is in `LinearAlgebra.BLAS.trmm!` (about 1.3 sec) and `trsm!` (about 0.7 sec).  If I change to use 4 threads on this 4-core computer I get 2.3 s in `trmm!` and 1.3 s in `trsm!`.

So it seems that multiple threads will not buy us much is the fitting time is dominated by BLAS operations.  You can check that fitting time is dominated by evaluation of the `obj!` function defined in `fit!` and that time is dominated by `updateL!`.  In this case the part of `updateL!` that dominates is `MixedModels.rmulΛ!
/home/bates/.julia/dev/MixedModels/src/remat.jl:403` and `MixedModels.rdiv!
/home/bates/.julia/dev/MixedModels/src/linalg.jl:59`.  In cases with multiple grouping factors there might also be `lmulΛ!` calls.

One possible approach is to copy the small lower triangular λ matrices to StaticArrays and use those in the multiplications.  I will do some timings with that.

PR #352, opened 2020-08-05
Julia 1.5 support for 2.x series

### LINKED ISSUE #351: Julia 1.5 fit! error
1.5 seems to break the LinearMixedModel.fit! - worked fine in 1.4.2.

 Error below:

```
fm1 = @time MixedModels.fit!(
    LinearMixedModel(
        @formula(
            xsret ~ xsrm + smb + hml + (xsrm + smb + hml|ticker) + (xsrm + smb + hml|sector) + 
            (xsrm + smb + hml|industry) + (xsrm + smb + hml|industry_group) + (xsrm + smb + hml|sub_industry)
        ),
        fin_df
    )
    , REML=true
)

TypeError: in typeassert, expected NLopt.Callback_Data, got a value of type Core.Compiler.UseRef

Stacktrace:
 [1] nlopt_callback_wrapper(::UInt32, ::Ptr{Float64}, ::Ptr{Float64}, ::Ptr{Nothing}) at /home/gsinha/.julia/packages/NLopt/eqN9a/src/NLopt.jl:411
 [2] optimize! at /home/gsinha/.julia/packages/NLopt/eqN9a/src/NLopt.jl:604 [inlined]
 [3] fit!(::LinearMixedModel{Float64}; verbose::Bool, REML::Bool) at /home/gsinha/.julia/packages/MixedModels/ALNok/src/linearmixedmodel.jl:251
 [4] top-level scope at timing.jl:174
 [5] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

palday:
This is just a dependency bump so that the last release works on Julia 1.5. Since I'm a bit behind on marshalling the final changes for the big 3.0 release, this seems like a nice way to give something back to the people. 

~Note that this should NOT be merged.~ This can be merged because I've created a `v2.x` series branch as the merge base that I may use to backport a few easy but good features from the current development branch (e.g. `Grouping`).

 When approved, I'll call the registrator and close the PR.

Closes #351.


palday:
The failure (1.5 on Linux) is related to the Cholesky issue (#324). I can't restart the Travis build, but it completes on 1.5 on Mac, so I would just say that's good enough for a backport.

palday:
The modified test for rank-deficiency works here even with AVX instructions.  While it's less than ideal and the rank deficient issue is still not completely resolved (#324), it looks like any resolution will involve a breaking change, so wouldn't be compatible with the 2.x series anyway. If I don't receive any objections by the end of the day, I'm going to merge this to get a compatibility release out.

test/linalg.jl
```diff
@@ -35,11 +35,11 @@ end
...
     df = (Y = randn(rng, 400), A = repeat(['N','Y'], outer=200),
         G = repeat('A':'T', inner = 2, outer=10), H = repeat('a':'j', inner=40))
-    lmm1 = fit!(LinearMixedModel(@formula(Y ~ 1+A+(1+A|G)+(1+A|H)), df,
-        wts  = ones(400)))
-    @test loglikelihood(lmm1) ≈ -578.9080978272708
-    MixedModels.reweight!(lmm1, ones(400))
-    @test loglikelihood(fit!(lmm1)) ≈ -578.9080978272708
+    m1 = fit(MixedModel, @formula(Y ~ 1 + A + (1+A|G) + (1+A|H)), df)
+    wm1 = fit(MixedModel, @formula(Y ~ 1+A+(1+A|G)+(1+A|H)), df, wts  = ones(400))
+    @test loglikelihood(wm1) ≈ loglikelihood(m1)
```
nalimilan, line 40: Could keep the hardcoded value, doesn't hurt.

test/statschol.jl
```diff
@@ -1,16 +1,14 @@
...
-
 const simdat = (
     G = repeat('A':'T', inner=10),
     H = repeat('a':'e', inner=2, outer=20),
-    U = repeat(0.:9, outer=20), 
+    U = repeat(0.:9, outer=20),
     V = repeat(-4.5:4.5, outer=20),
-    Y = 0.1 * randn(rng, 200),
-    Z = rand(rng, 200)
+    Y = (1:200)/200,
```
nalimilan, line 10: Random data seems better in general than a regular range. Any reason to change this?

palday: @nalimilan for both of these issues, the change in RNG in Julia 1.5 breaks things (see #306). I prefer the old tests too.  :man_shrugging: 

(That said, I'm going to see about using hard-coded RNG values that I generate once on my local machine as the next step in looking at the Cholesky issue #324.()

nalimilan: Yes, that's annoying, but how about using StableRNGs? (I'm surprised the test is so sensitive to changing the exact values. Maybe that's just bad luck?)

palday: Yeah, it's really bad luck. Also, StableRNGs is partly the result of the fit I threw when I was told about this change. ;)

test/linalg.jl
palday: For this, the most important thing isn't the unweighted fit, but rather that the unity-weighted fit has the same LL as the unweighted one.

There's a bug in re-weighting anyway (#334) that I may backport as well if I don't get the show-stoppers for 3.0 done in the next week.

test/statschol.jl
palday: Back to using (pseudo)0randomness.

PR #356, opened 2020-08-07
Use views, clean up ssqdenom

dmbates:
- For Julia v1.5.0 and later it is an advantage to use views, as in this `rankUpdate` method for sparse `A`
- Slightly cleaner code for `ssqdenom`.  If I am reading the profile timings correctly, `ssqdenom` is a non-negligible part of the time spent in `objective`.


palday:
I'm good to merge on this: looking at the `ssqdenom`: I'm wondering if it would make sense to memoize it / store it as part of the model if it's really taking up that much time.

dmbates:
The problem is that you don't know `ssqdenom` at the time the model is constructed - only when you go to fit it and set the REML property.  Let me look at the profile timings again.  I may be talking nonsense.

I checked and I was talking nonsense.  `ssqdenom` accounts for 4 ms. out of the total of 8.5 s fitting the model so not a big deal.

PR #361, opened 2020-08-27
Fix grouping name and amalgamate for interaction grouping terms

kleinschmidt:
There was a bug in how random effects like `(1 | a&b)` worked with amalgamate
and `fnames` (and hence `VarCorr` and display), because those paths were still
trying to use the `.sym` field on the grouping term.  I've replaced all
instances where `.sym` is used on a grouping term with a call to `fname`
(introduced in #197).

Also adds tests for `fit` with nesting and `&` in random effect grouping to
check the whole pipeline and to test equivalency of `(1 | a/b)` and `(1 | a&b) +
(1 | a)` (using the pastes data).

### DIFF
--- a/.github/workflows/Future.yml
+++ b/.github/workflows/Future.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        julia-version: [1.5.0-rc1, nightly]
+        julia-version: [1.5, nightly]
         julia-arch: [x64]
         os: [ubuntu-18.04]
     steps:
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -417,7 +417,7 @@ end
 
 Return the names of the grouping factors for the random-effects terms.
 """
-fnames(m::MixedModel) = ((tr.trm.sym for tr in m.reterms)...,)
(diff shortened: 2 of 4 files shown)


kleinschmidt:
I don't know why the tests fail on nightly or 1.5 but it seems unrelated to my changes, and they pass locally for me (1.5.0)...

palday:
For the failures: #324  (I've got a fix of sorts, but it needs some extra documentation to discuss the underlying issues and what guarantees we're making long-term).

palday:
LGTM, but can I get you to change the Future GH Actions to run  on `1.5` instead of `1.5-rc1`?

palday (approved):

kleinschmidt:
Shouldn't that be Tier 1 anyway?  I'm happy to fix that...

kleinschmidt:
I guess without updating the minimum version supported to 1.5 instead of 1.4 better keep that in the tests

palday:
Yep, exactly -- we want everything to work on 1.5, but (for now), we're still officially supporting 1.5. We certainly take advantage of some 1.5 performance improvements (especially with views), but the code should still run just fine on 1.4.

PR #362, opened 2020-08-30
Add raneftables

dmbates:
- add `raneftables` function to produce a `NamedTuple` of `columntables` of conditional means of the random effects
- currently applies only to `LinearMixedModel` although probably could apply to GLMMs too
- with this we should consider dropping the `named` optional argument from `ranef`, which would allow us to drop the dependency on the `NamedArrays` package.


src/linearmixedmodel.jl
```diff
@@ -417,7 +417,7 @@ end
 
 Return the names of the grouping factors for the random-effects terms.
 """
-fnames(m::MixedModel) = (fname.(m.reterms)...,)
+fnames(m::MixedModel) = ((tr.trm.sym for tr in m.reterms)...,)
```
dmbates, line 420: Looks like this change was problematic.  I will revert it.

palday: See #361

src/linearmixedmodel.jl
```diff
@@ -673,6 +673,22 @@ function ranef(m::LinearMixedModel{T}; uscale = false, named = false) where {T}
     vnmd
 end
 
+function retbl(mat, trm)
+    merge(
+        NamedTuple{(trm.trm.sym,)}((trm.levels,)),
```
palday, line 678: I suspect this will run into issues with grouping variables that are interactions (see #361).

src/linearmixedmodel.jl
```diff
@@ -673,6 +673,22 @@ function ranef(m::LinearMixedModel{T}; uscale = false, named = false) where {T}
...
+        columntable(Tables.table(transpose(mat), header=Symbol.(trm.cnames))),
+        )
+end
+
+"""
+    raneftables(m::LinearMixedModel; uscale = false)
+
+Return the conditional means of the random effects as a NamedTuple of columntables
+"""
+function raneftables(m::LinearMixedModel{T}; uscale = false) where {T}
```
palday, line 688:
```suggestion
function raneftables(m::MixedModel{T}; uscale = false) where {T}
```

This only depends on things that are defined for both LMM and GLMM. (Also, maybe move it to `mixedmodel.jl`? )

palday (commented):
At first I was opposed to dropping the `named` kwarg from `ranef`, but I grow increasingly amenable to the idea.

src/linearmixedmodel.jl
dmbates: Thanks for catching that.

src/linearmixedmodel.jl
dmbates: Good suggestion - now implemented.

dmbates:
> At first I was opposed to dropping the `named` kwarg from `ranef`, but I grow increasingly amenable to the idea.

Perhaps a separate PR.

palday (approved):

palday:
Looks good to me. Are you done tinkering?

dmbates:
> Looks good to me. Are you done tinkering?

Yep.  Please squash and merge if it seems okay to you.

PR #364, opened 2020-09-07
Remove dependency on Showoff package

### LINKED ISSUE #363: Failure showing fitted models on julia-1.6.0-DEV.855 and later
As I reported in https://github.com/JuliaGraphics/Showoff.jl/issues/24 the re-write of `Printf` in https://github.com/JuliaLang/julia/pull/32859 removed `Base.Grisu`, which has broken the `Showoff` package.

I have felt for a while that the use of `Showoff` is more ornamental than useful so maybe now is the time to bite the bullet and strip the use of `Showoff.showoff` from `MixedModels`.

One fewer dependency sounds like a good idea to me.  Opinions?

dmbates:
Drop dependency on the `Showoff` package. Closes #363
- The `show` method for the `VarCorr` type now uses a function `aligncompact` that uses `Base.alignment` on a `IOContext` with `:compact => true` and `Base.Ryu.writefixed` to align decimal points in the `Variance` and `Std.Dev.` columns.
- Also drop redundant checks in the CI.  We had updated the versions of Julia used in `Tier1.yml` and some of these combinations of architecture and operating system in `Tier2.yml` became redundant.
- The CI failure on `nightly` is from the instability of the column permutation in a rank-deficient test, not from these changes.


palday:
You completely eliminated Tier 2, but I think this is okay for now -- supporting 1.4 and 1.5 at the Tier 1 level makes sense until the next LTS release, especially since the change between those versions largely impact performance and not anything relevant for our internal API (unlike say the `mul!` or `Ryu` methods).

palday (approved):

palday:
I'm merging now -- but we should add figuring out official Tier 1 and Tier 2 standards pre 3.0 to our to do list. My suggestion would be:

1. Tier 1 is whatever is Tier 1 when we get close. No Tier 2.
2. At the next Julia LTS release, Tier 1 moves to Tier 2, and LTS becomes Tier 1.

dmbates:
> I'm merging now -- but we should add figuring out official Tier 1 and Tier 2 standards pre 3.0 to our to do list. My suggestion would be:
> 
>     1. Tier 1 is whatever is Tier 1 when we get close. No Tier 2.
> 
>     2. At the next Julia LTS release, Tier 1 moves to Tier 2, and LTS becomes Tier 1.

Thanks for merging.  Your proposal for Tier 1 and Tier 2 sounds good to me.  Did you know that 1.6.x will become the LTS version once 1.7.0 has been released?

I have been following the 1.6.0-DEV series and it really is a great improvement on precompile and loading times.  It also provides performance gains on some of our test cases.  Some of the gains are from allocating less memory.

kliegl:
@dmbates Do you recommend switching to 1.6.0-DEV?

dmbates:
@kliegl I wouldn't switch to 1.6.0-DEV at this time.  When it goes into alpha or beta test it might be worthwhile switching.  The speed changes are noticeable but not revolutionary.

PR #367, opened 2020-09-09
Fix for one tiny aspect of the rank deficiency issues

palday:
This fixes one tiny facet of the bigger issue lurking in #324.

In particular, OpenBLAS on SkyLakeX can do something weird for some (full rank!) matrices:

- The pivoted Cholesky estimates rank below full.
- The unpivoted Cholesky has `info ==0`, which means that the unpivoted Cholesky succeeded, suggesting numerical full rank.

In this case, the old code generated an invalid permutation consisting of the identity permutation with an extra element `0` appended to the end. The fix here trusts the success of the unpivoted Cholesky more than the rank of the pivoted Cholesky.

### DIFF
--- a/src/linalg/statschol.jl
+++ b/src/linalg/statschol.jl
@@ -10,9 +10,9 @@ function statscholesky(xtx::Symmetric{T}, tol::Real = -1) where {T<:AbstractFloa
     n = size(xtx, 2)
     chpiv = cholesky(xtx, Val(true), tol = T(-1), check = false)
     chunp = cholesky(xtx, check = false)
-    r = chpiv.rank
+    r = (chunp.info == 0) ? n : chpiv.rank
     piv = [1:n;]
-    if chunp.info > 0
+    if r < n
         nleft = n
         while r < nleft
             k = chunp.info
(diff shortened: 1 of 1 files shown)


dmbates (approved):
Looks good.  Let's wait for the checks to run then feel free to merge.

dmbates:
Is the failure on 1.5 a problem?  The failure on nightly is because Showoff hasn't been updated for the Grisu -> Ryu that took place back in the Julia-1.4.0 release.

palday:
The failure for 1.5 is the main rank deficiency issue (#324) and the failure for nightly is the `Showoff` issue (#363).

palday:
It's not much comfort, but on my other Linux machine without SkyLakeX, the tests pass on 1.5.

(I have learned far too much about numerical linear algebra and its interaction with different SIMD instructions for this issue ... and yet I still know that I know nothing.)

dmbates:
> It's not much comfort, but on my other Linux machine without SkyLakeX, the tests pass on 1.5.
> 
> (I have learned far too much about numerical linear algebra and its interaction with different SIMD instructions for this issue ... and yet I still know that I know nothing.)

One of Piet Hein's "grooks" is

> Knowing what thou knowest not is, in a sense, omniscience.

PR #368, opened 2020-09-09
Add coeftable method and property for MixedModelBootstrap

dmbates:
- A `MixedModelBootstrap` object already contained the standard errors for the fixed-effects coefficients.  This adds a method for `coeftable` and a `coeftable` property for `MixedModelBootstrap`.

- Needs tests - will do when I am more wide awake.

- I calculated the p-values here using `2*StatsFuns.normccdf(abs(z))` instead of `Distributions.ccdf(Distributions.Chisq(1), abs2(z))`.  If we change the `2*normccdf(abs(z))` call to a call to `SpecialFunctions.erfc(abs(z)*invsqrt2)` (where `invsqrt2` is an irrational of the value of `inv(sqrt(2))`) we can drop several indirect dependencies.


src/bootstrap.jl
```diff
@@ -165,10 +165,12 @@ function allpars(bsamp::MixedModelBootstrap{T}) where {T}
 end
 
 function Base.getproperty(bsamp::MixedModelBootstrap, s::Symbol)
-    if s ∈ [:objective, :σ, :θ]
+    if s ∈ [:objective, :σ, :θ, :se]
         getproperty.(getfield(bsamp, :bstr), s)
     elseif s == :β
         tidyβ(bsamp)
+    elseif s == :coeftable
+        tidyβsezp(bsamp)
```
palday, line 173: this doesn't line up with the current function name `coeftable` below

src/bootstrap.jl
```diff
@@ -243,6 +245,24 @@ function tidyβ(bsamp::MixedModelBootstrap{T}) where {T}
     result
 end
 
+function coeftable(bsamp::MixedModelBootstrap{T}) where {T}
```
palday, line 248: if we're going to name this `coeftable`, should we follow the `StatsBase.coeftable` API and return a `CoefTable` object? See e.g. the [method for `LinearMixedModel`](https://github.com/JuliaStats/MixedModels.jl/blob/9b7a748080c0f7904f00adde17a716db43caa26a/src/linearmixedmodel.jl#L197-L211)?

palday:
I think we still wind up depending on `Distributions` via `GLM`.

palday:
(Also, this is cool functionality! I think it might also be interesting to have a shortest coverage interval table.)

src/bootstrap.jl
dmbates: You're right - I changed the name then forgot to search for all occurrences of the old name.  Will fix.

src/bootstrap.jl
dmbates: Are you thinking of a vector of `coeftable` of `CoefTable` objects?  A single `CoefTable` of `nsamp * ncoef` rows might be unwieldy to work with.  Currently the returned object is a `Vector{NamedTuple}`, which is a row table, and can be converted to other tabular forms such as a DataFrame.  My recollection is that a `CoefTable` is not as easy to work with.  It was primarily created for its `show` method.

dmbates:
> I think we still wind up depending on `Distributions` via `GLM`.

And I always forget about GLMMs when I am on one of my "eliminate large packages in the dependencies" crusades.  Good to be reminded.

dmbates:
> (Also, this is cool functionality! I think it might also be interesting to have a shortest coverage interval table.)

Isn't that what we effectively get from
```
combine(groupby(DataFrame(bsamp.allpars, [:type, :group, :names])), :value => shortestcovint => :interval)
```

Are you thinking of an enhancement to this in some way or just adding a property name to do this automatically?  I would prefer If it could be done without creating a dependency on `DataFrames` either directly or indirectly through `DataFramesMeta`.

src/bootstrap.jl
palday: d'oh. I wasn't thinking clearly / hadn't read carefully enough. Returning a single `CoefTable` of that size would be unwieldy. Likewise for returning a `Vector{CoefTable}`. I don't know what the right course is here. The current variant is nice in terms of memory and dependencies, but does drift awy from the `StatsBase.coeftable` API expectation.

palday:
> > (Also, this is cool functionality! I think it might also be interesting to have a shortest coverage interval table.)
> 
> Isn't that what we effectively get from
> 
> ```
> combine(groupby(DataFrame(bsamp.allpars, [:type, :group, :names])), :value => shortestcovint => :interval)
> ```
> 
> Are you thinking of an enhancement to this in some way or just adding a property name to do this automatically? I would prefer If it could be done without creating a dependency on `DataFrames` either directly or indirectly through `DataFramesMeta`.

Once again, I thought quickly instead of clearly. Let's not introduce any extra dependencies. This is something for the long planned `MixedModelsExtras` package which will have heavyweight dependencies like `DataFrames` and something for plotting.

palday:
To close this out:

I think not using `CoefTable` makes sense here for the type of wrangling we want to do later. That said, we should probably rename this back to something like `tidyβsezp` or `tidysummary` or `tidystats` because [`StatsBase.coeftable` should return a `CoefTable`](https://juliastats.org/StatsBase.jl/stable/statmodels/#StatsBase.coeftable). In that case, we could then add in a `coeftable` method that represents the shortest coverage intervals or the like as `CoefTable`.

Sorry for the run-around on naming here.

dmbates:
No problem @palday with considering names carefully.  Naming things is known to be a difficult problem, https://www.martinfowler.com/bliki/TwoHardThings.html

I agree that it would be misleading to use `coeftable` as a name if the method/property did not return a `CoefTable`.  Would `coefpvalues` be okay, do you think?  I want to work `coef` into the name somehow because this method/property returns information on the fixed-effects coefficients.  And, despite our best efforts, having `pvalue` in the name will attract attention.

palday:
`coefpvalues` seems like a good way to get attention and has the two bits of that table people will care about in the name. And besides, we need at least some functionality that we'll regret later on. :wink: :smile:

dmbates:
I'll make the change to `coefpvalues` then.

src/MixedModels.jl
```diff
@@ -67,6 +67,7 @@ export @formula,
        bic,
        coef,
        coefnames,
+       coefpvalues,
        coeftable,
```
palday, line 71:
```suggestion
```

dmbates: I don't think we want to drop `coeftable` because there is a legitimate `coeftable` method for `MixedModel`.

palday: Ooops. Yes, you're absolutely right.

dmbates:
Should we merge this PR or would it be best to wait until after the course is finished?

palday:
I'm going to be a little rude and push to your branch to add tests. If you don't like my changes, feel free to revert the commit! Then I'll let you squash and merge when you're ready.

palday (approved):

dmbates:
> I'm going to be a little rude and push to your branch to add tests. If you don't like my changes, feel free to revert the commit! Then I'll let you squash and merge when you're ready.

Please feel free to do so.  I don't consider your doing that to be rude at all.

PR #369, opened 2020-09-12
Delete an extra backtick

likanzhan:


dmbates (approved):
Thanks for catching that.

PR #371, opened 2020-09-14
Remove dependency on BlockDiagonals

### LINKED ISSUE #366: Remove dependency on BlockDiagonals
Right now an update in the BlockDiagonals package is causing several other packages in the MixedModels dependencies to be downdated.  It turns out that BlockDiagonals is used in exactly one place so it should be straightforward to eliminate that dependency.

dmbates:
- Closes #366
- The dependency on `BlockDiagonals` was blocking updates on several other packages

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,7 +5,6 @@ version = "3.0.0-DEV"
 
 [deps]
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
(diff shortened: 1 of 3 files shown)


palday (commented):
This looks like it'll be fine, but it also looks like the type of thing that's easy to make thinko's on, so I'll approve and merge when the CIs are happy.

palday (approved):

PR #374, opened 2020-09-16
StatsModels.lrtest support methods

### LINKED ISSUE #320: Add lrtest support methods
As soon as JuliaStats/StatsModels.jl#162 lands, we should add in the following methods:

- [x] `loglikelihood` and ~`deviance`~ for REML-fitted models that throw an interpretable error.
- [x] `isnested` for `GeneralizedLinearMixedModel`
- [x] `isnested` for `LinearMixedModel`
- [x] refactor `likelihoodratiotest` to use `isnested`, perhaps add in deprecation (see previous discussion [here](https://github.com/JuliaStats/MixedModels.jl/pull/279#issuecomment-590055141)) 

Currently `likelihoodratiotest` can do one thing that `lrtest` cannot: model comparison on REML-fitted models. I'm wondering if we should keep it around for that purpose, give it some other name (suggestions?), or add in a custrom `lrtest` method that calls `objective`. There is one additional possibility: have `deviance` return -2 REML but issue a warning "This is not the deviance you are looking for."  Given that some funding providers and standards-givers demand REML-fitted models, I don't think we can just omit support for comparing REML-fitted models. 

There is one final option: I have code to do a bootstrap-based likelihood-ratio test. We could make our internal model comparison function do a bootstrapped comparison on the objective function and then it really would be different than the `lrtest` (and not depend on asymptotic assumptions about the deviance that the REML-criterion might not fulfill). In that case, what should we call it?

@dmbates Which of these is the smallest heresy?

palday:
For now, we're keeping `likelihoodratiotest` in. Two main reasons for that:

1. We print the formulae (but this can change)
2. We use don't check nesting (just basic compatibility in terms of family, links, objective, etc.). This means we get fewer false positives in cases like alternative specifications of compound symmetric random effects.

Closes #320.

### DIFF
--- a/src/likelihoodratiotest.jl
+++ b/src/likelihoodratiotest.jl
@@ -51,7 +51,7 @@ Likeihood ratio test applied to a set of nested models.
 
 Note that nesting of the models is not checked.  It is incumbent on the user 
 to check this. This differs from [`StatsModels.lrtest`](@ref) as nesting in 
-mixed models, especially in the random effects specification may be non obvious.
+mixed models, especially in the random effects specification, may be non obvious.
 
 This functionality may be deprecated in the future in favor of [`StatsModels.lrtest`](@ref).
 """
--- a/src/likelihoodratiotest.jl
+++ b/src/likelihoodratiotest.jl
@@ -82,7 +82,13 @@ function _likelihoodratiotest(m::Vararg{T}) where T <: MixedModel
     devs = objective.(m)[ord]
     dofdiffs = diff(dofs)
     devdiffs = .-(diff(devs))
-    pvals = ccdf.(Chisq.(dofdiffs), devdiffs)
+    pvals = map(zip(dofdiffs, devdiffs)) do (dof, dev)
+        if dev > 0
+            ccdf(Chisq(dof), dev)
+        else
+            NaN
+        end
+    end
 
     LikelihoodRatioTest(
         formulas,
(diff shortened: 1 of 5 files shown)


src/likelihoodratiotest.jl
```diff
@@ -45,34 +45,22 @@ end
...
-function likelihoodratiotest(m::GeneralizedLinearMixedModel...)
-    # TODO: test that all models are fit with same fast/nAGQ option?
-    glms = getproperty.(m,:resp);
-    allequal(Distribution.(glms)) ||
-        throw(ArgumentError("Models must be fit to the same distribution"))
-    allequal(string.(Link.(glms))) ||
-        throw(ArgumentError("Models must have the same link function"))
+Note that nesting of the models is not checked.  It is incumbent on the user 
+to check this. This differs from [`StatsModels.lrtest`](@ref) as nesting in 
+mixed models, especially in the random effects specification may be non obvious.
```
palday, line 54:
```suggestion
mixed models, especially in the random effects specification, may be non obvious.
```

dmbates (approved):
I agree that it is best to leave this as `likelihoodratiotest` for v3.0 and worry about compatibility with `StatsModels.lrtest` at a later time.  Thanks for adding the many tests.

Is this ready to merge now?

palday:
Good for squash and merge, if there's nothing else you want to add. :grinning:

PR #378, opened 2020-09-17
keep user-specified formula for display purposes

palday:
After using `apply_schema`, the user-specified formula is expanded out such that `a * b` is now represented as `a + b + a&b`. This is undesirable for display purposes, which is the only reason we keep the formula object around after model construction. If we use the original formula specified by the user, then everything stays nice and pretty.

Cons:
1. The fully expanded formula isn't immediately available. But we don't need this and if the user wants to see which terms there are in the model, then they can get this via `coefnames`. 
2. Information about contrast coding is no longer explicitly stored in the model. (In the expanded formula, the contrasts are part of the schema and so are stored with the terms.)  It's implicitly there in the fixed-efffects model matrix. I'm not sure how useful this information is after the fact. If this is really an issue, we could just store the `contrasts` dictionary as part of the `LinearMixedModel` struct.

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -98,7 +98,7 @@ function LinearMixedModel(
     optsum = OptSummary(θ, lbd, :LN_BOBYQA, ftol_rel = T(1.0e-12), ftol_abs = T(1.0e-8))
     fill!(optsum.xtol_abs, 1.0e-10)
     LinearMixedModel(
-        form,
+        f,
         allterms,
         sqrtwts,
         mkparmap(reterms),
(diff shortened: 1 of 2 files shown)


kliegl:
(1) Personally, I find it helpful to see how the program expands my model formula. This is not a strong preference, it also depends on the context. 
(2) I would very much be in favor to store the contrasts dictionary as part of the LinearMixedModel struct.

palday:
Okay, it turns out that this doesn't work the way I thought it did

Nosferican:
I do that in Econometrics.jl. Part for printing and part for deconstructing `predict` on new datasets. Is there other use cases here other than printing for which it might be helpful?

palday:
@Nosferican No, really just for printing. In models with big fixed-effects structures, the expanded formula is quite unwieldy for display purposes.

PR #384, opened 2020-09-22
DataAPI dependency and ref names

### LINKED ISSUE #383: Add dependency on DataAPI and change levs to levels
I think it would generally be a good idea to add a dependency on `DataAPI.jl` just to get access to certain standardized names and docstrings.  Currently the function to extract the levels of an `ReMat` is called `levs` but it really should be `levels` for consistency with other packages.

dmbates:
- direct dependency on `DataAPI`.  Several packages used by `MixedModels` depend on `DataAPI` so it was always in the indirect dependencies.
- add (unexported) methods for `levels`, `refarray`, `refpool`, and `refvalue`
- closes #383


test/FactorReTerm.jl
```diff
@@ -30,7 +30,7 @@ const LMM = LinearMixedModel
     end
 
     @testset "utilities" begin
-        @test levs(sf) == string.('A':'F')
+        @test refpool(sf) == string.('A':'F')
```
palday, line 33: They're equivalent in our implementation, but `levels` (instead of `refpool`) is closer to the direct replacement to `levs`.

dmbates: Milan pointed out in another discussion that the value of `refpool` can include unused levels and `missing` whereas the value of `levels` cannot include `missing` and should drop unused levels.  I will change that here.

src/remat.jl
```diff
@@ -133,7 +133,15 @@ function getθ!(v::AbstractVector{T}, A::ReMat{T}) where {T}
     v
 end
 
-levels(A::ReMat) = A.levels
+function DataAPI.levels(A::ReMat)
+    pool = A.levels
+    present = falses(size(pool))
+    @inbounds for i in A.refs
+        present[i] = true
+        all(present) && return pool
```
palday, line 141: this short-circuits the `for` at the cost of iterating over `present` a lot via `all`. It might be easier to check whether `length(A.levels) == length(A.refs)` and short-circuit before the loop

dmbates: Not sure I understand.  `length(A.levels) == length(A.refs)` doesn't really tell you anything.  `length(A.refs)` is always the number of observations.  The only interesting cases for categorical data are when `length(A.levels)` is less than `length(A.refs)`, otherwise there is no compression in the representation.

My motive in doing the short-circuit check is that in the common cases, where `length(A.levels)` is small, `all(present)` would be a fast operation on this, comparatively small, `BitArray`, which is what `falses` produces.  If the number of levels does not exceed the word size this check can be done by masking in a register.  It takes a few nanoseconds on my machine.

I suspect that the short circuit will never make a noticeable difference so we can remove it if doing so will make the code clearer.

palday: Ah, okay, that was a thinko on my part. I think this will be fine given the typical  nlevels << nobs.

dmbates:
Okay to squash and merge then?

palday:
One final thing: do we need to re-export `levels, refpool, refarray, refvalue` ?

dmbates:
> One final thing: do we need to re-export `levels, refpool, refarray, refvalue` ?

You're correct.  Will do.

palday:
I'll squash and merge when the CIs finish running :grinning:

dmbates:
> I'll squash and merge when the CIs finish running grinning

Thanks.

PR #385, opened 2020-09-24
Documentation for fulldummy

### LINKED ISSUE #255: document fulldummy

dmbates:
- closes #255 
- add a docstring for `fulldummy` and an example in the `constructors.md` source file for the online documentation

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "MixedModels"
 uuid = "ff71e718-51f3-5ec2-a782-8ffcbfa3c316"
-author = ["Phillip Alday <phillip.alday@mpi.nl>", "Douglas Bates <dmbates@gmail.com>", "Jose Bayoan Santiago Calderon <jbs3hp@virginia.edu>"]
+author = ["Phillip Alday <me@phillipalday.com>", "Douglas Bates <dmbates@gmail.com>", "Jose Bayoan Santiago Calderon <jbs3hp@virginia.edu>"]
 version = "3.0.0-DEV"
 
 [deps]
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -11,8 +11,8 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 BenchmarkTools = "0.5"
-DataFrames = "0.20"
-Documenter = "0.24"
-FreqTables = "0.3"
+DataFrames = "0.21"
+Documenter = "0.25"
+FreqTables = "0.4"
 Gadfly = "1"
-StatsBase = "0.32"
+StatsBase = "0.33"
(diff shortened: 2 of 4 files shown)


palday (commented):
Looks good so far. I think one of the things @kleinschmidt has been working on is what happens when you have `(1 + fulldummy(x)|g)` vs. `(1|g)` + `(0+fulldummy(x)|g)`. Like `||`, the `lme4` behavior for `dummy()` is somewhat surprising about when the intercept is repeated and when not. I think the current implementation (#269) is consistent and nice, but I suspect edge cases will keep cropping up.

kleinschmidt:
I think after #338 you should be able to do `(1|g) + (fulldummy(x) | g)` (without needing the `0+`).

palday:
@kleinschmidt Yeah, that makes sense and is consistent (which is what you were going for :grin: ).

dmbates:
I am a bit queasy about the use of the sleepstudy data to illustrate `fulldummy` and especially `zerocorr(1+fulldummy(days))`.  This is obviously longitudinal data - in fact it is a classic example of multiple, short, time series - and using indicators for the days and then declaring that the random effects could be independent is, well, questionable.  It is fine for illustrating the capabilities but its interpretation is problematic.

If anyone can suggest an alternative it would be welcome.  Right now I am just looking to close issues tagged as Release 3.

palday:
What about `machines`?

dmbates:
I suppose `machines` could be used.  @kliegl can you suggest models to fit?

palday:
@kleinschmidt Since you've obsessed about gotchas in `fulldummy`: are there any major ones we should clarify? We can always improve the documentation later, but known edge cases that we've made an API decision on are something we should probably also just document for ourselves.

kleinschmidt:
I don't know of any major gotchas but then again I'm probably the wrong person to ask :)  I think the biggest potential one is that it's going to not "play nicely" with other categorical predictors and the intercept I think...so like `(fulldummy(x) | g)` might add an intercept...or if you have `(fulldummy(x) + y | g)` then you might still get promotion of `y` to full rank, but I haven't checked that.  But those seem like a real edge case and if you're doing that then...you probably have a good reason.

palday:
So maybe we just add a byline:

> The interaction of `fulldummy` with complex random effects is subtle and complex with numerous potential edge cases. As we discover these edge cases, we will document and determine their behavior. Until such time, please check the model summary to verify that the expansion is working as you expected. If it is not, please report a use case on GitHub.

dmbates:
@palday Can you handle the clean-up and squash and merge on this one?

palday:
@dmbates will do :grinning:

palday (approved):

PR #386, opened 2020-09-26
PCA row labels

### LINKED ISSUE #340: Labels for PCA output rows
Could we display the MixedModel "Variance components" labels also with the corresponding rows in the PCA output, e.g. "Component loadings"?

palday:
Closes #340.

Sample of output:
```julia
julia> m1 = fit(MixedModel, @formula(score ~ 1 + Machine + (1 + Machine | Worker)), MixedModels.dataset(:machines))
 score ~ 1 + Machine + (1 + Machine | Worker)
   logLik   -2 logLik     AIC        BIC    
  -108.2089   216.4178   236.4178   256.3077

Variance components:
            Column    Variance Std.Dev.   Corr.
Worker   (Intercept)  13.813236 3.716616
         Machine: B   28.684049 5.355749 +0.49
         Machine: C   11.240468 3.352681 -0.36 +0.30
Residual               0.924687 0.961606
 Number of obs: 54; levels of grouping factors: 6

  Fixed-effects parameters:
(Intercept)  52.3556      1.53414  34.13    <1e-99
Machine: B    7.96667     2.20985   3.61    0.0003
Machine: C   13.9167      1.40576   9.90    <1e-22

julia> m1.PCA.Worker

Principal components based on correlation matrix
 (Intercept)   1.0     .     .
 Machine: B    0.49   1.0    .
 Machine: C   -0.36   0.3   1.0

Normalized cumulative variances:
[0.4988, 0.9235, 1.0]

Component loadings
                PC1    PC2    PC3
 (Intercept)  -0.75   0.23   0.62
 Machine: B   -0.64  -0.49  -0.59
 Machine: C    0.17  -0.84   0.52
```

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -246,8 +246,9 @@ function Base.show(io::IO, pca::PCA;
         if pca.rnames !== missing 
             n = length(pca.rnames)
             cv = string.(printmat)
+            dotpad = lpad(".", div(maximum(length, cv),2))
             for i = 1:n, j = (i+1):n
-                cv[i, j] = ""
+                cv[i, j] = dotpad
             end
             neg = startswith.(cv, "-")
             if any(neg)
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -173,11 +173,13 @@ Principal Components Analysis
 
 * `covcorr` covariance or correlation matrix
 * `sv` singular value decomposition
-* `corr` is this a correlation matrix
+* `rnames` rownames of the original matrix
+* `corr` is this a correlation matrix?
 """
 struct PCA{T<:AbstractFloat}
     covcor::Symmetric{T,Matrix{T}}
     sv::SVD{T,T,Matrix{T}}
+    rnames::Union{Vector{String}, Missing}
     corr::Bool
 end
 
@@ -193,9 +195,9 @@ For `LinearMixedModel`, a named tuple of PCA on each of the random-effects terms
 If `corr=true`, then the covariance is first standardized to the correlation scale.
 """
 
-function PCA(covfac::AbstractMatrix; corr::Bool=true)
+function PCA(covfac::AbstractMatrix, rnames=missing; corr::Bool=true)
     covf = corr ? rownormalize!(copy(covfac)) : copy(covfac)
-    PCA(Symmetric(covf*covf', :L), svd(covf), corr)
+    PCA(Symmetric(covf*covf', :L), svd(covf), rnames, corr)
 end
 
 function Base.getproperty(pca::PCA, s::Symbol)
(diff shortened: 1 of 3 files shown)


src/utilities.jl
```diff
@@ -240,27 +242,54 @@ function Base.show(io::IO, pca::PCA;
...
+            n = length(pca.rnames)
+            cv = string.(printmat)
+            dotpad = lpad(".", div(maximum(length, cv),2))
+            for i = 1:n, j = (i+1):n
+                cv[i, j] = dotpad
+            end
+            neg = startswith.(cv, "-")
+            if any(neg)
+                cv[.!neg] .= " ".* cv[.!neg]
+            end
```
ararslan, line 256:
```suggestion
            for i = 1:length(cv)
                if !startswith(cv[i], "-")
                    cv[i] = " " * cv[i]
                end
            end
```
Slightly longer but it has the benefit of only making one pass over `cv`. Otherwise you're doing one to broadcast `startswith`, one to check `any`, and then one for the broadcasted indexed assignment.

palday: But this will always add the space prefix for positive numbers, even if there are no negative numbers, right? I guess the extra space really doesn't matter for printing because everything remains aligned.

dmbates: I agree with Phillip that the check for any negative numbers is probably needed.  Could the check be written as
```
if any(startswith("-"), cv)
```
so that it dispatches to the `any(f::Function, a::AbstractArray; dims) in Base at reducedim.jl:732` method? I think that would avoid allocation of the intermediate result.

dmbates: I guess my scheme doesn't work out that well because you then have to do the `startswith(cv[i], "-")` check again to add the spaces.

It seems that in @palday 's version `neg` will be a BitArray and cv vector is usually short so I don't think there will be much overhead in allocating `neg`.

src/utilities.jl
```diff
@@ -240,27 +242,54 @@ function Base.show(io::IO, pca::PCA;
...
     show(io, round.(pca.cumvar, digits=ndigitscum))
     println(io)
     if loadings
-        println(io, "Component loadings")
-        Base.print_matrix(io, round.(pca.loadings, digits=ndigitsmat))
+        println(io, "\nComponent loadings")
+        printmat = round.(pca.loadings, digits=ndigitsmat)
+        
+        if pca.rnames !== missing 
+            pclabs = [Text(""); Text.( "PC$i" for i in 1:length(pca.rnames))]
```
dmbates, line 285: Will there be an alignment problem if `length(pca.rnames) > 9`?

dmbates (approved):

src/utilities.jl
palday: I think the bigger alignment problem will be if the numerical column width (arising `ndigitsmat`) is less than `maximum(length, pca.rnames)`.  But `Base.print_matrix` is pretty smart when it comes to printing `String`/`Text`.

PR #388, opened 2020-09-26
coef and fixef (and names versions) for GLMMs

### LINKED ISSUE #318: fixef and coef for GLMM
The memory-efficient versions of `fixef!`, `coef`, etc. introduced for LMM in #287 don't yet GLMM methods. Since the new versions also introduced a new way to handle rank-deficiency in extractor methods, we should probably update the GLMM methods. Potentially related is this warning I get when running the testsuite for #317.

```julia
WARNING: Method definition fixef!(AbstractArray{T, 1}, MixedModels.LinearMixedModel{T}) where {T} in module MixedModels at /XXX/MixedModels.jl/src/linearmixedmodel.jl:337 overwritten at /XXX/MixedModels.jl/src/generalizedlinearmixedmodel.jl:281.
```

dmbates:
- closes #318 
- needs tests
- also provides methods for `coefnames` and `fixefnames`


palday:
Naive approaches to creating a rank deficient GLMM fail even to do initial model construction because Σ isn't positive definite. I'm working on something and then I'll have tests.

Do we also need a `fixef!` method for GLMM? I guess consistency of the existing methods is most important and we can add in those extra bits later.

dmbates:
> Naive approaches to creating a rank deficient GLMM fail even to do initial model construction because Σ isn't positive definite. I'm working on something and then I'll have tests.
> 
> Do we also need a `fixef!` method for GLMM? I guess consistency of the existing methods is most important and we can add in those extra bits later.

I don't think it should be necessary.  For a GLMM both the fixed-effects coefficients and the spherical random effects are stored in the struct itself.  The fixef! method is for evaluating the fixed-effects estimates from the information in the blocked Cholesky factor and wouldn't apply to a GLMM.

palday:
@dmbates That makes sense and then we have one less thing to worry about.

palday:
I've explored the rank deficient case a bit further. It seems that the new rank detection isn't picking up on the rank deficiency and then the Cholesky decomposition fails in `updateL!` because the matrix is only positive semidefinite. 

This is actually an interesting pathology. I don't want to mess with the rank deficiency stuff anymore right now.  I also  don't want to add in any error checking into the `updateL!` or anything it calls, but we could catch the `PosDefException` in `fit!`, add an `@info` (or `@error`?) that this occurs as the result of a misspecified model, including undetected rank deficiency and then rethrow the exception. @dmbates Thoughts?

dmbates:
@palday Sounds like a plan.  This whole area of rank deficiency is a quagmire and I think we are devoting considerable effort to cases that should not occur frequently.

dmbates:
Shall we squash and merge this?

palday:
Let's. The helpful error message is low priority and really needs another PR (also because I need to make sure it doesn't impact performance.)

PR #390, opened 2020-09-28
Drop dependency on NamedArrays package

### LINKED ISSUE #389: Drop named argument to ranef now that we have raneftables
If we drop the `named` argument to `ranef`, now that we have `raneftables`, we can remove the dependency on the `NamedArrays` package.  Shall we do that?

dmbates:
- drop the `named` argument to `ranef` and the dependency on the `NamedArrays` package that it required
- the `raneftables` function provides an alternative that, in most cases, is easier to work with.
- closes #389


src/bootstrap.jl
```diff
@@ -15,7 +15,7 @@ Tables.Schema:
  :σ          T
  :β          NamedTuple{β_names}{NTuple{p,T}}
  :se         StaticArrays.SArray{Tuple{p},T,1,p}
- :θ          StaticArrays.SArray{Tuple{p},T,1,k}
+ :θ          StaticArrays.SArray{Tuple{k},T,1,k}
```
palday, line 18: This was a copy-paste error I introduced, right?

palday (approved):

src/bootstrap.jl
dmbates: Yes.

Sorry for putting that in this PR without explanation.  I was trying to decide if the `StaticArrays` package was still needed and found this while doing a search on the name.

palday: I thought as much, no worries!

palday:
I'm happy to squash and merge when you're ready.

PR #391, opened 2020-09-28
note on threading in bootstrap

palday:
For now, I would like to leave the threading option in the API for a few reasons:

1. There are some machines with dozens (or even > 100 cores) where there might actually be some boost from threading at both the BLAS and simultaneous fits levels. These aren't so rare in university clusters -- I have seen a few in neuroscience.
2. I suspect it might be possible to eek out a bit more of a performance boost with finer grained locking, but that's not for today.3
3. If someone is using an alternative BLAS that isn't multi-threaded but still has multiple cores (maybe this is the case on ARM?), then they can still get a boost from multithreading.


dmbates (approved):
Looks good.  Feel free to squash and merge (or tell me to do it).

PR #393, opened 2020-09-29
Initial version of NEWS

dmbates:
- Initial version of `NEWS.md` and script to add footnotes for commit numbers


dmbates:
The ordering of the sections is from the order in which I encountered the changes.  I think it should be in terms of importance which I would suggest is
- ReMat and FeMat types
- New formula features
- Rank deficiency
- Principal components
- Parametric bootstrap
- Availability of test data sets
- Replacement capabilities
- Package dependencies

How does that look?  Alternatives are welcome.

The purpose of the NEWS-update.jl file is to be run before the branch release to add URLs for issue numbers.  It was blatantly stolen from the Julia repository.

palday:
I'm torn about the position of Re- and FeMat. On the one hand, those are important changes, but on the other hand, it's mostly our close collaborators, who are developing functionality based on those, that care. 

The rest of the ordering looks great.

dmbates:
Okay, I agree with what you said.  How about moving `ReMat and FeMat types` to just before `Availability of test data sets`?

palday:
@dmbates Works for me!

dmbates:
Should I make these changes or will you?  Either is okay with me.

palday (approved):

palday:
Looks good to me now!

PR #394, opened 2020-09-30
Delete the  redundant word `either'

likanzhan:


src/linearmixedmodel.jl
```diff
@@ -815,7 +815,7 @@ sqrtpwrss(m::LinearMixedModel) = last(m.L)
 
 Return the denominator for penalized sums-of-squares.
 
-For MLE, this value is either the number of observation for ML. For REML, this
+For MLE, this value is the number of observation for ML. For REML, this
```
palday, line 818:
```suggestion
For MLE, this value is the number of observations. For REML, this
```

dmbates: Thank you @likanzhan for your careful reading of the comments and documentation.  I think the sentence could be shortened to 
```suggestion
For MLE, this value is the number of observations. For REML, this
```

dmbates (approved):

src/linearmixedmodel.jl
palday: @dmbates we were reviewing at the same time -- I fixed that too.

@likanzhan thank you from me too -- you've caught a lot of issues in the documentation, thanks!

PR #395, opened 2020-10-01
AbstractReTerm

### LINKED ISSUE #293: Question on ZeroCorr term / AbstractRandomEffectsTerm
I have the following problem:
I have a multiple dispatch function with a term-type signature

`term::TimeExpandedTerm{<:RandomEffectsTerm}}`

But `y~1+zerocorr(A|B)`  has the following type:

```julia
unfold.TimeExpandedTerm{MixedModels.ZeroCorr}
```
without any access to the underlying subtype.

Is it possible to change

```julia
struct ZeroCorr <: AbstractTerm
    term::RandomEffectsTerm
end
```
to

```julia
struct ZeroCorr{T<:RandomEffectsTerm} <: AbstractTerm 
    term::T
end
```
Then I think I could do a dispatch like this (I think):

`term::TimeExpandedTerm{<:Union(RandomEffectsTerm,AbstractTerm{<:RandomEffectsTerm}}`

**Alternative**, is it possible to introduce an AbstractRandomEffectsTerm, where the Term ZeroCorr could inherit from? 

**Maybe** it is also possible to put all ZeroCorr definition in one place so that it is easier to understand what is internally happening?

tagging @kleinschmidt

palday:
Closes #293. Not sure if this is a good or bad idea. Thoughts, @kleinschmidt ?


dmbates:
Should `FullDummy` also be an `AbstractReTerm`?

palday:
@dmbates Good question. I'm hoping @kleinschmidt has a strong feeling on that, but I'll give it a think. The only downside I see is that StatsModels also has `fulldummy` for any categorical predictor and `fulldummy` is defined at the level of a single categorical experimental variable, while `zerocorr` is approximately at the level of a grouping variable.

kleinschmidt (approved):
If this is just for dispatch then it seems harmless to me.  There's no struct to represent full dummy coding (it just creates a new `CategoricalTerm` using `FullDummyCoding`) so that's not an issue.  `RandomEffectsTerm` doesn't seem to be used elsewhere in the code (e.g. to restrict what can be stored as the `reterms` in the model struct) so no objections there either.

NEWS.md
```diff
@@ -82,3 +84,4 @@ Package dependencies
 [#382]: https://github.com/JuliaStats/MixedModels.jl/issues/382
 [#384]: https://github.com/JuliaStats/MixedModels.jl/issues/384
 [#390]: https://github.com/JuliaStats/MixedModels.jl/issues/390
+[#395]: https://github.com/JuliaStats/MixedModels.jl/issues/390
```
dmbates, line 87: Shouldn't the URL end in 395?
```suggestion
[#395]: https://github.com/JuliaStats/MixedModels.jl/issues/395
```

dmbates (approved):
Other than the URL change, it looks good.

There is a script in the `docs/` directory to create the footnotes automatically.

PR #397, opened 2020-10-02
fix propogation of defaults in non-column table contructor

palday:
This is a nasty, subtle bug about where you set the defaults.


ararslan (approved):
Yikes, good catch!

PR #399, opened 2020-10-02
Bugfixes and last set of backports for 2.x 

### LINKED ISSUE #360: Backports for 2.x
- [x] Grouping for 2.x (#353) 
- [x] use weights when fitting LMM (#334) (#357)
- [x] fix propogation of defaults in non-column table contructor (#398) 
- [x] check GLMM stderr (#308)
- [x] document problems with GLMM families with dispersion parameter (#373)
- [x] change GLM column labels (#335)
- [x] `ranef method for GLMM (#336)
- [ ] ~~make (1|x) + (y|x) equivalent to (1|x) + (0+y|x) (#338)~~ (too much effort, potentially can be interpreted as breaking, and 3.0 is coming out soon)
- [x]  use `wts` argument in GLMM constructor when conversion to column table is necessary (#397, #398)
- [x] fix `ranef` docstring (#355)

(Since we depend on GLM for other things, maybe it's possible to extract these automatically from GLM? Perhaps we need an upstream change to have their col names accessible as a constant?)

palday:
Closes #360 

With 3.0 basically a RC at this point, this will be the last set of backports I do, but some of these are quite serious bugs, so it makes sense to provide these as a non breaking update before the breaking release.


dmbates:
Looks good to me.  Thanks for doing this.

PR #401, opened 2020-10-02
PosDefException in rank deficiency docs

palday:
Hopefully this will return in a search result if users do a quick search before asking questions.


palday:
@dmbates As always, feel free to suggest/make changes if my simplified explanation is infelicitous.

dmbates (approved):
Looks good.  The wording is fine, I think.

Issue #402, opened 2020-10-05
Failure on julia-nightly


dmbates:
When running the tests by hand on
```julia
julia> versioninfo()
Julia Version 1.6.0-DEV.1128
Commit 8519538bb7 (2020-10-05 13:05 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-10.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = "/usr/share/code/code"
  JULIA_NUM_THREADS = 4
```
I got a failure on 
```julia
julia> pca = models(:kb07)[3].PCA.item

Principal components based on correlation matrix
Error showing value of type MixedModels.PCA{Float64}:
ERROR: MethodError: no method matching zero(::String)
Closest candidates are:
  zero(::T) where T<:Dates.TimeType at /home/bates/git/julia/usr/share/julia/stdlib/v1.6/Dates/src/types.jl:423
  zero(::Diagonal{T, StaticArrays.SVector{N, T}}) where {N, T} at /home/bates/.julia/packages/StaticArrays/l7lu2/src/SDiagonal.jl:43
  zero(::FillArrays.Zeros{T, N, Axes} where Axes) where {T, N} at /home/bates/.julia/packages/FillArrays/RM6r2/src/FillArrays.jl:517
  ...
Stacktrace:
  [1] iszero(x::String)
    @ Base ./number.jl:40
  [2] fzeropreserving(bc::Base.Broadcast.Broadcasted{LinearAlgebra.StructuredMatrixStyle{LowerTriangular{Float64, Matrix{Float64}}}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(string), Tuple{LowerTriangular{Float64, Matrix{Float64}}}})
    @ LinearAlgebra ~/git/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/structuredbroadcast.jl:107
[...]

PR #403, opened 2020-10-05
PCA show method fixes for Julia 1.6

palday:
Partly addresses #402.


palday:
Closed in favor of #404

PR #405, opened 2020-10-05
GLMM refit!

### LINKED ISSUE #400: refit! for GLMM?
Should we add in a `refit!` method for GLMM? @dmbates

palday:
Closes #400

### DIFF
--- a/src/generalizedlinearmixedmodel.jl
+++ b/src/generalizedlinearmixedmodel.jl
@@ -557,8 +557,8 @@ function refit!(m::GeneralizedLinearMixedModel{T};
     optsum = m.LMM.optsum
     # we need to reset optsum so that it
     # plays nice with the modifications fit!() does 
-    optsum.lowerbd = foldl(vcat, lowerbd(c) for c in reterms)
-    optsum.initial = foldl(vcat, getθ(c) for c in reterms)
+    optsum.lowerbd = mapfoldl(lowerbd, vcat, reterms)
+    optsum.initial = mapfoldl(getθ, vcat, reterms)
     optsum.final = copy(optsum.initial)
     optsum.xtol_abs = fill!(copy(optsum.initial), 1.0e-10)
     optsum.initial_step = T[]
--- a/src/generalizedlinearmixedmodel.jl
+++ b/src/generalizedlinearmixedmodel.jl
@@ -530,6 +530,36 @@ ranef(m::GeneralizedLinearMixedModel; uscale::Bool=false) = ranef(m.LMM, uscale=
 LinearAlgebra.rank(m::GeneralizedLinearMixedModel) = first(m.LMM.feterms).rank
 
 """
+    refit!(m::LinearMixedModel[, y::Vector];
+          fast::Bool = (length(m.θ) == length(m.optsum.final)), 
+          nAGQ::Integer = m.optsum.nAGQ))
+
+Refit the model `m` after installing response `y`.
+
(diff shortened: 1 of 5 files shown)


src/generalizedlinearmixedmodel.jl
```diff
@@ -529,6 +534,52 @@ ranef(m::GeneralizedLinearMixedModel; uscale::Bool=false) = ranef(m.LMM, uscale=
...
+                nAGQ::Integer = m.optsum.nAGQ)  where T
+    
+    deviance!(m, 1)
+    reevaluateAend!(m.LMM)
+    
+    reterms = m.LMM.reterms
+    optsum = m.LMM.optsum
+    # we need to reset optsum so that it
+    # plays nice with the modifications fit!() does 
+    optsum.lowerbd = foldl(vcat, lowerbd(c) for c in reterms)
```
ararslan, line 560:
```suggestion
    optsum.lowerbd = mapfoldl(lowerbd, vcat, reterms)
```
Can you write it this way instead of using a generator?

src/linearmixedmodel.jl
```diff
@@ -309,6 +309,10 @@ objective and the parameters are printed on stdout at each function evaluation.
 """
 function fit!(m::LinearMixedModel{T}; verbose::Bool = false, REML::Bool = false) where {T}
     optsum = m.optsum
+    # this doesn't matter for LMM, but it does for GLMM, so let's be
```
palday, line 312:
```suggestion
    # this doesn't matter for LMM, but it does for GLMM, so let's be consistent
```

src/linearmixedmodel.jl
```diff
@@ -698,19 +702,22 @@ function reevaluateAend!(m::LinearMixedModel)
 end
 
 """
-    refit!(m::LinearMixedModel[, y::Vector])
+    refit!(m::LinearMixedModel[, y::Vector]; REML=m.optsum.reml)
```
palday, line 705:
```suggestion
    refit!(m::LinearMixedModel[, y::Vector]; REML=m.optsum.REML)
```

src/generalizedlinearmixedmodel.jl
```diff
@@ -529,6 +534,52 @@ ranef(m::GeneralizedLinearMixedModel; uscale::Bool=false) = ranef(m.LMM, uscale=
...
+    optsum.initial_step = T[]
+    optsum.feval = -1
+
+    fit!(m; fast=fast, nAGQ=nAGQ)
+end
+
+function refit!(m::GeneralizedLinearMixedModel{T}, y; 
+                fast::Bool = (length(m.θ) == length(m.optsum.final)), 
+                nAGQ::Integer = m.optsum.nAGQ) where T
+    resp = m.y
```
dmbates, line 573: I think this is a bit confusing to call this `resp` as `m.y` actually evaluates to `m.resp.y`.  It seems that there are too many `resp`s in play here.  Perhaps `mrespy` or `m_resp_y`?

I would also make the rhs more explicit as `m.resp.y`.  I had forgotten about `getproperty` and initially looked for a field named `y` in `GeneralizedLinearMixedModel`.

dmbates (approved):
Thanks for adding all the tests too.

src/generalizedlinearmixedmodel.jl
palday: Ah, good points all. I'll modify and update.

PR #408, opened 2020-10-06
Update documentation

### LINKED ISSUE #407: Document that model matrix is always at least double precision / Float64
Just marking this from our internal todo list. I think this is the last MixedModels 3.0 issue without an active PR. :smile:

dmbates:
- Update documentation to explain that only `Float64` is allowed as an element type when optimizing with `NLopt`
- Correct the model formula for an example with `cask` within `batch`
- Closes #407

### DIFF
--- a/docs/src/constructors.md
+++ b/docs/src/constructors.md
@@ -8,8 +8,9 @@ LinearMixedModel
 
 ## Examples of linear mixed-effects model fits
 
-For illustration, several data sets from the *lme4* package for *R* are made available in `.feather` format in this package.
-These include the `Dyestuff` and `Dyestuff2` data sets.
+For illustration, several data sets from the *lme4* package for *R* are made available in `.arrow` format in this package.
+Often, for convenience, we will convert these to `DataFrame`s.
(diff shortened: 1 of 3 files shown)


docs/src/constructors.md
```diff
@@ -67,6 +68,18 @@ fm1reml = fit(MixedModel, fm, dyestuff, REML=true)
...
+### Float-point type in the model
+
+The type of `fm1`
+```@example Main
+typeof(fm1)
+```
+includes the floating point type used internally for the various matrices, vectors, etc. that represent the model.
+At present, this will always be `Float64` because the parameter estimates are optimized using the [`NLopt` package](https://github.com/JuliaOpt/NLopt.jl) which calls compiled C code that only allows for optimization with respect to a `Float64` parameter vector.
+
+So in theory other floating point types, such as `BigFloat` or `Float32`, can be used to define a model but in practice only `Float64` works at present.
```
palday, line 80: A lot of our documentation has something like "in theory, everything is easy, but in practice, double precision leads makes life *interesting*".

palday (approved):
I'm happy when you and the CIs are happy

PR #414, opened 2020-10-08
Restore cached fits to previous state when modifying in testsuite

palday:
~Still WIP~

If you're going to modify something that's cached, you should restore the original value.


palday:
Superseded by #415.

PR #416, opened 2020-10-11
Update benchmarks to TestData.

dmbates:
- update the benchmarks to the current TestData contents
- should add benchmarks of GLMMs
- the `nested` example is still way off.  Somehow there must be compilation time in there.
- may need to allow more time per run if the longer benchmarks (`mrk17_exp1:[12]`, `ml1m:1`) are going to be useful.  These are only timed for 1 run at present.  Alternatively, just use some smaller, faster examples.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,171 +1,110 @@
-using BenchmarkTools, DataFrames, MixedModels, RData, Tables
+using BenchmarkTools, MixedModels, StatsModels
+using MixedModels: dataset
 
 const SUITE = BenchmarkGroup()
 
-const dat = Dict(Symbol(k) => v for (k, v) in load(joinpath(
-    dirname(pathof(MixedModels)),
-    "..",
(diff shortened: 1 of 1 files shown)


palday:
> * the `nested` example is still way off.  Somehow there must be compilation time in there.
Without taking a look at the specifics, the syntax for nesting involves a very different codepath through the StatsModels machinery and the StatsModels machinery can be quite hard on the compiler. I'll take a look when I get a chance, but @kleinschmidt will know a lot more than my vague, uninformed intution.

 
> * may need to allow more time per run if the longer benchmarks (`mrk17_exp1:[12]`, `ml1m:1`) are going to be useful.  These are only timed for 1 run at present.  Alternatively, just use some smaller, faster examples.

I would be keen to have one or two big models in the benchmarks because that will really help us to detect whether every scales as we would expect. Some of the smaller examples might even fit in the CPU cache on some machines, but those definitely won't.

PR #418, opened 2020-10-13
simulate! and thus parametricbootstrap for GLMM

### LINKED ISSUE #245: simulate! methods for GLMM
GLMMs are not yet supported in `parametricbootstrap` largely because `simulate!` doesn't have appropriate methods for GLMM.

At least for Bernoulli models (the most common GLMM in my experience), this should be a relatively straightforward to add in. We may want to wait until the API for the linear case has stabilized a bit.

palday:
Closes #245.

I had to add in a few accessor-to-preallocated-buffer methods for GLMM elsewhere (`fixef!`, etc.). 
~`stderror!` actually does a hidden allocation, but I'll fix that as soon as I've looked up `StatsBase.stderror`.~

~There's also still some debugging code in there, but it should otherwise be in a good state to check for conceptual errors.~
```
julia> using DataFrames

julia> using MixedModels

julia> using StableRNGs

julia> contra = MixedModels.dataset(:contra)
Arrow.Table: (dist = ["D01", "D01", "D01", "D01", "D01", "D01", "D01", "D01", "D01", "D01"  …  "D61", "D61", "D61", "D61", "D61", "D61", "D61", "D61", "D61", "D61"], urban = ["Y", "Y", "Y", "Y", "Y", "Y", "Y", "Y", "Y", "Y"  …  "N", "N", "N", "N", "N", "N", "N", "N", "N", "N"], livch = ["3+", "0", "2", "3+", "0", "0", "3+", "3+", "1", "3+"  …  "1", "3+", "3+", "2", "2", "3+", "2", "3+", "0", "3+"], age = [18.44, -5.56, 1.44, 8.44, -13.56, -11.56, 18.44, -3.56, -5.56, 1.44  …  -5.56, 14.44, 19.44, -9.56, -2.56, 14.44, -4.56, 14.44, -13.56, 10.44], use = ["N", "N", "N", "N", "N", "N", "N", "N", "N", "N"  …  "N", "N", "N", "Y", "N", "N", "N", "N", "N", "N"])

julia> gm0 = fit(MixedModel, @formula(use ~ 1+age+abs2(age)+urban+livch+(1|urban&dist)), contra, Bernoulli(), fast=true);

julia> bs = parametricbootstrap(StableRNG(42), 1000, deepcopy(gm0));
Progress: 100%|███████████████████████████████████████ Time: 0:00:14

julia> combine(groupby(DataFrame(bs.β), :coefname), :β => first ∘ shortestcovint => :lower, :β => last ∘ shortestcovint => :upper)
7×3 DataFrame
│ Row │ coefname    │ lower       │ upper       │
│     │ Symbol      │ Float64     │ Float64     │
├─────┼─────────────┼─────────────┼─────────────┤
│ 1   │ (Intercept) │ -1.36078    │ -0.671704   │
│ 2   │ age         │ -0.01329    │ 0.0218545   │
│ 3   │ abs2(age)   │ -0.00565161 │ -0.00294269 │
│ 4   │ urban: Y    │ 0.439763    │ 1.09529     │
│ 5   │ livch: 1    │ 0.495756    │ 1.11646     │
│ 6   │ livch: 2    │ 0.493062    │ 1.23956     │
│ 7   │ livch: 3+   │ 0.549297    │ 1.23275     │

# CIs from the normal Wald approximation for comparison
julia> DataFrame(coef=fixefnames(gm0), lower=fixef(gm0) .- 2 .* stderror(gm0), upper=fixef(gm0) .+ 2 .* stderror(gm0))
7×3 DataFrame
│ Row │ coef        │ lower       │ upper       │
│     │ String      │ Float64     │ Float64     │
├─────┼─────────────┼─────────────┼─────────────┤
│ 1   │ (Intercept) │ -1.39311    │ -0.667259   │
│ 2   │ age         │ -0.0153779  │ 0.0218421   │
│ 3   │ abs2(age)   │ -0.00581783 │ -0.00289668 │
│ 4   │ urban: Y    │ 0.407463    │ 1.09011     │
│ 5   │ livch: 1    │ 0.486554    │ 1.14029     │
│ 6   │ livch: 2    │ 0.516467    │ 1.26265     │
│ 7   │ livch: 3+   │ 0.527929    │ 1.27843     │
```

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -44,8 +44,8 @@ The default random number generator is `Random.GLOBAL_RNG`.
 # Named Arguments
 
 `β`, `σ`, and `θ` are the values of `m`'s parameters for simulating the responses.
-`σ` is only valid for `LinearMixedModel` and `GeneralizedLinearMixedModel` when
-the family that has a dispersion parameter.
+`σ` is only valid for `LinearMixedModel` and `GeneralizedLinearMixedModel` for
+families with a dispersion parameter.
 `use_threads` determines whether or not to use thread-based parallelism.
 
 Note that `use_threads=true` may not offer a performance boost and may even
@@ -63,7 +63,10 @@ function parametricbootstrap(
     θ::AbstractVector=morig.θ,
     use_threads::Bool=false,
 ) where {T}
-    β, σ, θ = convert(Vector{T}, β), T(σ), convert(Vector{T}, θ)
+    if σ !== missing
+        σ = T(σ)
+    end
+    β, θ = convert(Vector{T}, β), convert(Vector{T}, θ)
     βsc, θsc, p, k, m = similar(β), similar(θ), length(β), length(θ), deepcopy(morig)
 
     β_names = (Symbol.(fixefnames(morig))..., )
@@ -138,7 +141,7 @@ function allpars(bsamp::MixedModelBootstrap{T}) where {T}
     )
     nrmdr = Vector{T}[]  # normalized rows of λ
     for (i, r) in enumerate(bstr)
-        σ = r.σ
+        σ = r.σ === missing ? 1 : r.σ
         for (nm, v) in pairs(r.β)
             push!.(cols, (i, "β", missing, String(nm), v))
         end
@@ -161,7 +164,7 @@ function allpars(bsamp::MixedModelBootstrap{T}) where {T}
                 end
             end
         end
-        push!.(cols, (i, "σ", "residual", missing, σ))
+        r.σ === missing || push!.(cols, (i, "σ", "residual", missing, r.σ))
     end
     (
         iter=cols[1],
@@ -301,7 +304,7 @@ function tidyσs(bsamp::MixedModelBootstrap{T}) where {T}
     )
     for (iter, r) in enumerate(bstr)
         setθ!(bsamp, iter)    # install r.θ in λ
-        σ = r.σ
+        σ = r.σ === missing ? 1 : r.σ
         for (grp, ll) in zip(keys(fcnames), λ)
             for (cn, col) in zip(getproperty(fcnames, grp), eachrow(ll))
                 push!(result, NamedTuple{colnms}((iter, grp, Symbol(cn), σ * norm(col))))
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -44,18 +44,20 @@ The default random number generator is `Random.GLOBAL_RNG`.
 # Named Arguments
 
 `β`, `σ`, and `θ` are the values of `m`'s parameters for simulating the responses.
+`σ` is only valid for `LinearMixedModel` and `GeneralizedLinearMixedModel` when
+the family that has a dispersion parameter.
 `use_threads` determines whether or not to use thread-based parallelism.
 
-Note that `use_threads=true` may not offer a performance boost and may even 
+Note that `use_threads=true` may not offer a performance boost and may even
 decrease peformance if multithreaded linear algebra (BLAS) routines are available.
-In this case, threads at the level of the linear algebra may already occupy all 
+In this case, threads at the level of the linear algebra may already occupy all
 processors/processor cores. There are plans to provide better support in coordinating
-Julia- and BLAS-level threads in the future. 
+Julia- and BLAS-level threads in the future.
 """
 function parametricbootstrap(
     rng::AbstractRNG,
     n::Integer,
-    morig::LinearMixedModel{T};
+    morig::MixedModel{T};
     β::AbstractVector=coef(morig),
     σ=morig.σ,
     θ::AbstractVector=morig.θ,
@@ -108,7 +110,7 @@ end
 
 function parametricbootstrap(
     nsamp::Integer,
-    m::LinearMixedModel;
+    m::MixedModel;
     β = m.β,
     σ = m.σ,
     θ = m.θ,
@@ -129,7 +131,7 @@ function allpars(bsamp::MixedModelBootstrap{T}) where {T}
     nresrow = length(bstr) * npars
     cols = (
         sizehint!(Int[], nresrow),
-        sizehint!(String[], nresrow), 
+        sizehint!(String[], nresrow),
         sizehint!(Union{Missing,String}[], nresrow),
         sizehint!(Union{Missing,String}[], nresrow),
         sizehint!(T[], nresrow),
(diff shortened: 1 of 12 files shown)


palday:
I need to fix the Gaussian with non-identity link pathway.

test/bootstrap.jl
```diff
@@ -65,4 +98,26 @@ end
...
     end
+
+
+    @testset "Bernoulli simulate! and GLMM boostrap" begin
+        contra = dataset(:contra)
+        gm0 = fit(MixedModel, only(gfms[:contra]), contra, Bernoulli(), fast=true)
+        bs = parametricbootstrap(StableRNG(42), 100, gm0)
+        bsci = combine(groupby(DataFrame(bs.β), :coefname),
+                       :β => first ∘ shortestcovint => :lower,
+                       :β => last ∘ shortestcovint => :upper)
```
dmbates, line 109: Its a very minor gripe but `shortestcovint` is being called twice on the same bootstrap results and it is comparatively expensive (extract the data, sort it, calculate the end points, run the loop to determine the shortest interval, etc.). I would just store the intervals and do the calls to `first` and `last` later when I need to extract the endpoints.

src/simulate.jl
```diff
@@ -37,7 +37,98 @@ function simulate!(
...
+from the `GlmResp.d` field. This isn't vectorized nicely because
+for distributions where the scale/dispersion is dependent on the
+location (e.g. Bernoulli, Binomial, Poisson), it's not really
+possible to avoid creating multiple `Distribution` objects.
+
+Note that `n` is the `n` parameter for the Binomial distribution,
+*not* the number of draws from the RNG. It is then used to change the
+random draw (an integer in [0, n]) into a probability (a float in [0,1]).
+"""
+function _rand(rng::AbstractRNG, d::Distribution, location, scale=NaN, n=1)
```
dmbates, line 117: Maybe use `missing` instead of `NaN` to indicate that `scale` is not defined?

src/simulate.jl
```diff
@@ -37,7 +37,98 @@ function simulate!(
...
+    # note that these m.resp.y and m.LMM.y will later be sychronized in (re)fit!()
+    # but for now we use them as distinct scratch buffers to avoid allocations
+
+    # the noise term is actually in the GLM and not the LMM part so no noise
+    # at the LMM level
+    η = fill!(m.LMM.y, zero(T))
+    y = m.resp.y
+
+    # assemble the linear predictor
+    @inbounds for trm in m.reterms             # add the unscaled random effects
```
dmbates, line 83: May need to flag this for consideration when we decide what to do about scale parameters.  The `unscaledre!` call is based on the assumption that any scaling factors have been transformed away.

src/simulate.jl
```diff
@@ -37,7 +37,98 @@ function simulate!(
...
+        unscaledre!(rng, η, trm)
+    end
+
+    # scale by lm.σ and add fixed-effects contribution
+    BLAS.gemv!('N', one(T), lm.X, β, lm.σ, η)
```
dmbates, line 88: Are you sure that `lm.σ` is the correct multiple here?  It just calls `sdest(lm)` which calls `varest(m)` which is based on `pwrss(m)`.  And that value is not necessarily related to any scale parameter.

dmbates (commented):
I like the idea of using the `StableRNGs` package in the tests.

I think it will be necessary to consider the scaling in building `η` in the `simulate!` method.  I put a comment there.  I'm pretty sure that using `sdest` applied to the penalized weighted least squares structure for the working response is not correct.

test/bootstrap.jl
palday: Fair point and something I meant to clean up before pushing, but then forgot.

src/simulate.jl
palday: Right now, `GeneralizedLinearMixedModel.sigma` yields `sdest(::GeneralizedLinearMixedModel)` which returns `NaN` for models without a dispersion parameter. I have no objections to changing this -- it's in a strict sense breaking, but

1. Users really shouldn't be invoking `sdest`, `varest` or `dispersion` on models without a dispersion parameter. 
2. We already warn that the behavior for models with dispersion parameter isn't correct.
3. I would view fixing this behavior as a bugfix, which are allowed to be breaking to fix incorrect behavior.

My deep dive into the GLMM code this last week has highlighted to me that these definitions are inconsistent. Currently, For models  without a dispersion parameter:
1. `dispersion` is defined to be 1. This follows the behavior of `GLM.jl` for Bernoulli, Binomial and Poisson.
2. `varest` is `NaN`. This is a MixedModels function, not defined in `StatsBase` nor `GLM`.
3. `sdest` is `NaN`. This is a MixedModels function, not defined in `StatsBase` nor `GLM`.

I think it makes sense to have `varest` and `sdest` return `missing` for Bernoulli, Binomial and Poisson and document that this differs from the behavior of `dispersion` (much the same way that `coef` and `fixef` do similar but not identical things). Thoughts?

src/simulate.jl
palday: Yep, my deep dive into the GLMM code has highlighted this to me. It 'works' here because sigma is usually close to 1 for families without a dispersion parameter and I think unit scaling is correct.

palday:
The failure on nightly looks unrelated:
```julia
   MethodError: propertynames(::LinearMixedModel{Float64}, ::Bool) is ambiguous. Candidates:
    propertynames(m::LinearMixedModel, private) in MixedModels at /home/runner/work/MixedModels.jl/MixedModels.jl/src/linearmixedmodel.jl:564
    propertynames(x, private::Bool) in Base at reflection.jl:1530
```

src/simulate.jl
palday: I've now implemented this suggestion.

palday:
@dmbates Care to take a second look?

src/simulate.jl
dmbates: Thanks for implementing this and sorry that I haven't responded earlier.  Living in a swing state at this point in the election is nerve wracking and my productivity has plummeted.

src/bootstrap.jl
```diff
@@ -290,7 +304,7 @@ function tidyσs(bsamp::MixedModelBootstrap{T}) where {T}
     )
     for (iter, r) in enumerate(bstr)
         setθ!(bsamp, iter)    # install r.θ in λ
-        σ = r.σ
+        σ = r.σ === missing ? 1 : r.σ
```
dmbates, line 307: This could be written as
```
σ = coalesce(r.σ, one(T))
```
to keep type consistency and more in the spirit of the `missing` design.

dmbates (approved):
Just one minor suggestion of using `coalesce` instead of `x === missing ? ...`

src/bootstrap.jl
palday: Nice, thanks for the tip!

PR #419, opened 2020-10-14
Better loglikelihood for GLMM 

palday:
Amongst other things, this makes it so that the `show(::GeneralizedLinearMixedModel)` method works for models with a dispersion parameter.

(~If it's not obvious from the commit history,~ this is me stripping off part of #291 so that at least some of those improvements make it in sooner rather than later.)


dmbates (approved):
Sorry not to review this previously.  I just noticed now that you asked for my review.

At some point we should revisit the decision to use `NaN` for the value of `sdest` when there is no dispersion parameter.  It seems that in modern Julia we should consider `missing` so we can extract a value of type `T` with `coalesce(sdest(m), one(T))`.

Thanks for straightening out the use of ϕ in `loglikelihood`.

palday:
So we had actually revisited that decision in a previous PR (#418, on GLMM parametric bootstrap) -- `sdest` and `varest` return `missing` for families without a dispersion parameter, but `dispersion` returns `1` for compatibility with `GLM.jl`. Over- and underdispersion are things I would like to think about there as well, but that comes after the dispersion parameter fits even work consistently.....

 I've now merged those changes back into this branch and am re-running the tests.

PR #420, opened 2020-10-21
Run CI on tag push, not on release creation

ararslan:
I think this is what Documenter needs in order to deploy documentation for tagged versions properly. Will hopefully fix stable docs issues.


palday (approved):

palday:
Thanks!

PR #421, opened 2020-10-21
Fix `propertynames` method ambiguity on Julia nightly

ararslan:
The ambiguity is between `propertynames(::Any, ::Bool)` and the overloads in this package `propertynames(::T, ::Any)` for various types `T`. Since the second argument should always be a `Bool` anyway, we can simply add the type annotation to the function signature to fix the ambiguity.


palday (approved):

palday:
As soon as the CIs pass, I'll merge. Thanks!

palday (approved):

PR #422, opened 2020-10-21
Create manualdocs.yml

palday:
Attempt at manual docs deployment. @ararslan

### DIFF
--- a/.github/workflows/manualdocs.yml
+++ b/.github/workflows/manualdocs.yml
@@ -1,0 +1,21 @@
+# This is a basic workflow that is manually triggered
+
+name: Manual Documenter
+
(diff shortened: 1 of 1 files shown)


ararslan (approved):
So you then just go to the Actions tab for the repo, select this workflow, and trigger it manually whenever documentation is being a pain? Pretty cool.

mortenpi:
Just a drive-by comment which might be useful: for missing release docs (because this would just work for `dev/` I think?), you can tag pseudo-releases. See "Fixing broken release deployments" here: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#Documentation-Versions

palday:
@mortenpi  Thanks for the tip! That'll be useful later on. :) I suspect the problem is secret/token management.

PR #423, opened 2020-10-22
Fix docs deployment

palday:
I think the fundamental issue was TagBot's secrets, but I'm not 100% sure.

I've also re-arranged a few Actions and fixed a broken doctest.

@ararslan The Zenodo link is busted, but I don't think I have the GitHub Organizations priveleges to manage the Zenodo link.


DilumAluthge:
I don't think this PR will actually fix your docs. I notice you have a Documentation job, but it doesn't actually do the right thing. The documentation job needs to, among other things, explicitly run the `docs/make.jl` file.

palday:
The documentation job is directly copied and pasted from the https://github.com/julia-actions/julia-docdeploy -- if that doesn't work, then we should probably file a PR upstream as well.

DilumAluthge:
I've never actually used that action (https://github.com/julia-actions/julia-docdeploy). I don't know how it works.

DilumAluthge:
We might need to just deprecate that action.

DilumAluthge:
The method I've always used is to use the Documentation job in the workflow files generated by PkgTemplates.jl.

palday:
It at least partly works -- it generates and deploys the `dev` docs.

DilumAluthge:
That's good!

palday:
I think the issue with triggering on releases is related to the [TagBot token access issue](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions)

DilumAluthge:
That's easily enough fixed by giving TagBot the documenter SSH key.

palday:
Already did that ;)

palday:
(That was the original change in this PR, but I've been checking our doctests while I was on this bughunt.)

DilumAluthge:
Make sure you also give CompatHelper the documenter SSH key, or your tests won't run on the PRs that CompatHelper opens.

DilumAluthge:
> (That was the original change in this PR, but I've been checking our doctests while I was on this bughunt.)

Yeah in #424 I commented out the doctest section because doctests were failing.

DilumAluthge:
Also, what is the difference between "Future" and "Tier 1"?

It will probably be easier for you to put all the CI in a single workflow file like #424 does.

palday:
Thanks for all the help, @DilumAluthge . This has been lots of painful details....

DilumAluthge:
Yeah it looks like you discovered all of these details on your own manually, which sounds quite painful.

Chris has built all this stuff into PkgTemplates by default. If you do this:
```julia
using PkgTemplates

t = Template(; 
    user="my-username",
    dir="~/code",
    authors="Acme Corp",
    julia=v"1.1",
    plugins=[
        License(; name="MPL"),
        Git(; manifest=true, ssh=true),
        GitHubActions(; x86=true),
        Codecov(),
        Documenter{GitHubActions}(),
        Develop(),
    ],
)

generate(t, "MixedModels") # or whatever package name
```

And then just copy the three workflow files into your package:
- `.github/workflow/CompatHelper.yml`
- `.github/workflow/TagBot.yml`
- `.github/workflow/ci.yml`

Then you don't have to worry about the details.

(Taken from here: https://invenia.github.io/PkgTemplates.jl/dev/user/#A-More-Complicated-Example-1)

DilumAluthge:
Also, since your package is public, you don't need to use a token when uploading to Codecov.

For simplicity, I would remove that from your workflow file.

DilumAluthge:
And also, delete the `CODECOV_TOKEN` GitHub secret from this repository, since you don't need it.

DilumAluthge:
https://blog.codecov.io/2020/03/09/tokenless-uploads-for-github-actions/

palday:
Future and Tier 1 differ in what we support. Future is Julia nightly, so we like to know if things are working there, but we don't stress if the tests fail. Tier 1 has to work. Tier 1 will be moved to Tier 2 (nice to work, but can fail with good reason) when the next Julia LTS is released and made Tier 1.

DilumAluthge:
Hmmm. The Julia version of each job is already included in the job name, so you can see it in the list. Is that sufficient?

(I'm just trying to think of ways to simplify your config - it'll be less complicated if you only have a single CI config.)

DilumAluthge:
Also, another way to potentially simplify config: note that `'1'` automatically expands to the latest stable 1.x release of Julia. So instead of writing `'1.5'`, you can just write `'1'` - and when 1.6 is released, `'1'` will automatically start expanding to `'1.6'`.

palday:
The Tiers are also for the badges in the README.

(I'm not the original author on any of the CI stuff, so I'm also thinking hard about what the original motivations were.)

DilumAluthge:
> The Tiers are also for the badges in the README.

Yep in that case you'll need to keep them in separate jobs.

Nosferican:
Battle-tested docs template I use:
```
jobs:
  ci:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        julia-version: ['1.5']
        julia-arch: [x64]
        os: [ubuntu-20.04]
    steps:
      - uses: actions/checkout@v2
      - uses: julia-actions/setup-julia@v1
        with:
          version: ${{ matrix.julia-version }}
      - name: Cache artifacts
        uses: actions/cache@v1
        env:
          cache-name: cache-artifacts
        with:
          path: ~/.julia/artifacts 
          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
          restore-keys: |
            ${{ runner.os }}-test-${{ env.cache-name }}-
            ${{ runner.os }}-test-
            ${{ runner.os }}-
      - uses: julia-actions/julia-buildpkg@v0.1
      - uses: julia-actions/julia-runtest@v0.1
      - uses: julia-actions/julia-processcoverage@v1
      - uses: codecov/codecov-action@v1
        with:
          file: lcov.info
      - uses: julia-actions/julia-docdeploy@releases/v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
which I then customize as needed. Should be quite current.

DilumAluthge:
This won't trigger TagBot though. To trigger TagBot you need to use an SSH key, e.g. DOCUMENTER_KEY

palday:
TagBot runs as is right now, just doesn't deploy the release docs.

DilumAluthge:
Whoops I misspoke. There is a different problem with @Nosferican's template. If you use the GITHUB_TOKEN to push, GitHub Pages won't build your site.

DilumAluthge:
https://github.com/JuliaDocs/Documenter.jl/issues/1177

DilumAluthge:
It may have been partially fixed by GitHub: https://github.com/JuliaDocs/Documenter.jl/issues/1177#issuecomment-590138234

That being said I always recommend just using the DOCUMENTER_KEY when pushing docs, and also when running TagBot

DilumAluthge:
And also for running CompatHelper

Nosferican:
Using the `GITHUB_TOKEN` will push fine it just needs a previous commit to the `gh-pages` branch. After that condition has been satisfied it can be used just fine.

For the `TagBot.yaml` as mentioned above it is a different workflow with different conditions so it doesn't rely on Action-chaining so it is fine to use the official template which runs as a daily CRON.

DilumAluthge:
> Using the `GITHUB_TOKEN` will push fine it just needs a previous commit to the `gh-pages` branch. After that condition has been satisfied it can be used just fine.

Personally I still consider that to be a bug. But if it works for people( and they don't mind making sure that they have already committed to the branch) then that's good!

> For the `TagBot.yaml` as mentioned above it is a different workflow with different conditions so it doesn't rely on Action-chaining so it is fine to use the official template which runs as a daily CRON.

The official TagBot template (https://github.com/JuliaRegistries/TagBot/blob/master/README.md#setup) will not trigger Documenter to build on tags. You need to have the DOCUMENTER_KEY.

ararslan (approved):

PR #427, opened 2020-11-04
Simulation and prediction methods

### LINKED ISSUE #175: No predict method for GeneralizedLinearMixedModels
I am estimating a glm similar to the VerbAgg example in the documentation. The predict method doesn't support GeneralizedLinearMixedModel. I was able to replicate the error using the VerbAgg example.

```
using DataFrames, MixedModels, RData

const dat = Dict(Symbol(k)=>v for (k,v) in 
    load(joinpath(dirname(pathof(MixedModels)), "..", "test", "dat.rda")));

verbaggform = @formula(r2 ~ 1 + a + g + b + s + m + (1|id) + (1|item));

gm1 = fit(GeneralizedLinearMixedModel, verbaggform, dat[:VerbAgg], Bernoulli())

predict(gm1, dat[:VerbAgg])
```

And I get the following error:

```
MethodError: no method matching predict(::GeneralizedLinearMixedModel{Float64}, ::DataFrame)
Closest candidates are:
  predict(!Matched::StatsModels.DataFrameRegressionModel{T,T} where T, ::AbstractDataFrame; kwargs...) where T at /home/ec2-user/.julia/packages/StatsModels/pBxdt/src/statsmodel.jl:118
  predict(!Matched::StatsModels.DataFrameRegressionModel, ::Any...; kwargs...) at /home/ec2-user/.julia/packages/StatsModels/pBxdt/src/statsmodel.jl:28
  predict(::GeneralizedLinearMixedModel) at /home/ec2-user/.julia/packages/MixedModels/YIBOE/src/PIRLS.jl:134
  ...

Stacktrace:
 [1] top-level scope at In[8]:1
```

palday:
Closes #175.

Note: this won't do prediction intervals. We only support confidence intervals via bootstrapping, so I think it's fair to kick the can down the road on prediction intervals as well. If a user really needs prediction intervals, they can use these methods with the parametric bootstrap to achieve that goal.

(I hope to be able to open source some code for "effects plots" soon. That code could potentially be enhanced by adding in the conditional variance on the BLUPs for RE-level effects, which would take it a bit beyond the FE-level implemented in the R package of the same name. This is mostly a note to self for later.)

### To do:

- [x] add docs page showing the difference between `predict` and `simulate` (e.g. that `predict` assumes essentially fixed RE levels no no observation-level variance, but `simulate` uses random draws for both all levels of variation)
- [x] move all `simulate` code to `simulate.jl`
- [x] move `predict` to end of testsuite
- [x] re-organize simulate methods for GLMM (maybe allow for some duplicated code if this avoids methods that are too abstract)

### Tests (and potentially code to write):

- [x] LMM `predict` tabular with `new_re_levels=:missing`
- [x] LMM `predict` tabular with `new_re_levels=:population`
- [x] LMM `predict` tabular with `new_re_levels=:error`
- ~LMM `predict` tabular with `new_re_levels=:simulate`~
- [X] ~LMM `predict` matrix using RE~
- [X] ~LMM `predict` matrix using only FE~
- [x] GLMM `predict` tabular with `new_re_levels=:missing`
- [x] GLMM `predict` tabular with `new_re_levels=:population`
- [x] GLMM `predict` tabular with `new_re_levels=:error`
- ~GLMM `predict` tabular with `new_re_levels=:simulate`~
- [x] ~GLMM `predict` matrix using RE~
- [x] ~GLMM `predict` matrix using only FE~
- [x] LMM `simulate` tabular 
- [x] ~LMM `simulate` matrix~
- [x] GLMM `simulate` tabular 
- [x] ~GLMM `simulate` matrix~


src/predict.jl
```diff
@@ -0,0 +1,495 @@
...
+        # because we're using ranef(), this actually
+        # adds the *scaled* random effects
+        for (rt, bb) in zip(m.reterms, ranef(m))
+            unscaledre!(y, rt, bb)
+        end
+    end
+
+    type == :linpred && return y
+
+    @inbounds for (idx, val) in enumerate(y)
```
dmbates, line 71: Would it be cleaner to write
```suggestion
    y .= linkinv.(Link(m.resp), y)
```

dmbates: I meant that to replace the whole loop.

src/predict.jl
```diff
@@ -0,0 +1,495 @@
...
+The keyword argument `new_re_levels` specifies how previously unobserved
+values of the grouping variable are handled. Possible values are
+`:population` (return population values, i.e. fixed-effects only),
+`:missing` (return `missing`), `:error` (error on this condition;
+the error type is an implementation detail), `:simulate` (simulate new values).
+For `:simulate`, the values are determined by solving for their values by
+using the existing model's estimates for the new data. (These are in general
+*not* the same values as the estimates computed on the new data.)
+
+Predictions based purely on the fixed effects can be obtained with by
```
dmbates, line 133:
```suggestion
Predictions based purely on the fixed effects can be obtained by
```

src/predict.jl
```diff
@@ -0,0 +1,495 @@
...
+    The `predict` and `predict!` methods with `newX` as a fixed effects matrix
+    differ from the methods with `newdata` in tabular format. The matrix
+    methods are more efficient than the tabular methods. The tabular methods
+    can accomodate new values of the grouping variable(s), but the matrix methods
+    cannot.
+"""
+function StatsBase.predict(m::LinearMixedModel{T}, newdata::Tables.ColumnTable;
+                           new_re_levels=:population) where T
+
+    new_re_levels in (:population, :missing, :)
```
dmbates, line 158: Is there a missing `|| throw(...` here?

palday: yes -- looks like a chunk got chopped off at some point.

src/predict.jl
palday: yes.

I have no idea why I wrote it as a loop.

docs/src/prediction.md
```diff
@@ -0,0 +1,167 @@
...
+fit!(msim)
+DisplayAs.Text(ans) # hide
+```
+
+For simulating from generalized linear mixed models, there is no `type` option because the observation-level always occurs at the level of the response and not of the linear predictor.
+
+## API Reference
+
+Note all the methods that take new data as a table construct an additional `MixedModel` behind the scenes, even when the new data is exactly the same as the data that the model was fitted to.
+For the simulation methods in particular, these thus from a convenience wrapper for constructing a new model and
```
dmbates, line 155:
```suggestion
For the simulation methods in particular, these thus form a convenience wrapper for constructing a new model and
```

src/linearmixedmodel.jl
```diff
@@ -644,8 +644,6 @@ function objective(m::LinearMixedModel{T}) where {T}
     isempty(wts) ? val : val - T(2.0) * sum(log, wts)
 end
 
-StatsBase.predict(m::LinearMixedModel) = fitted(m)
```
dmbates, line 647: Could this method and the corresponding method for `GeneralizedLinearMixedModel` be combined as
```
StatsBase.predict(m::MixedModel) = fitted(m)
```

src/predict.jl
```diff
@@ -0,0 +1,292 @@
...
+    StatsBase.predict(m::LinearMixedModel{T},
+                    newX::AbstractMatrix{T}=m.X; use_re=true)
+    StatsBase.predict!(y::AbstractVector{T}, m::GeneralizedLinearMixedModel{T},
+                    newX::AbstractMatrix{T}=m.X; use_re=true, type=:response)
+    StatsBase.predict(m::GeneralizedLinearMixedModel{T},
+                    newX::AbstractMatrix{T}=m.X; use_re=true, type=:response)
+
+Predict response for new values of the fixed-effects matrix X.
+
+The modifying methods overwrite `y` in place, while the non modifying methods
```
dmbates, line 13: Would it be better to say `mutating` rather than `modifying`?  I think the term `mutating` is used more frequently.

src/predict.jl
```diff
@@ -0,0 +1,292 @@
...
+
+    mul!(y, newX, m.β)
+
+    use_re || return y
+
+    # because we're using ranef(), this actually
+    # adds the *scaled* random effects
+    for (rt, bb) in zip(m.reterms, ranef(m))
+        unscaledre!(y, rt, bb)
+    end
```
dmbates, line 46: Do we know that the `reterms` will line up with the `newX`?  It seems that if `size(newX, 1) == length(y)` and `size(newX, 2) == length(m.β)` then the `use_re = false` version works but that doesn't mean that the random-effects terms will match with `length(y)`.

src/predict.jl
```diff
@@ -0,0 +1,292 @@
...
+    y
+end
+
+function StatsBase.predict(m::LinearMixedModel{T}, newX::AbstractMatrix{T}=m.X;
+                          use_re=true) where T
+    # this is called `use_re` in case we later decide to support prediction
+    # with only a subset of the RE
+
+    y = zeros(T, nobs(m))
+    predict!(y, m, newX; use_re=use_re)
```
dmbates, line 83: Here there seems to be an assumption that `size(newX, 1) == nobs(m)`.  There is no need to check it because the `mul!` call in `predict!` will fail if that is not the case.  But should it be `zeros(T, size(newX, 1))` instead so that it will work, at least in the `use_re = false` case, if `newX` does not conform to the number of observations in the original model?

src/predict.jl
```diff
@@ -0,0 +1,292 @@
...
+
+function StatsBase.predict(m::GeneralizedLinearMixedModel{T}, newX::AbstractMatrix{T}=m.X;
+                           use_re=true, type=:response) where T
+    # this is called `use_re` in case we later decide to support prediction
+    # with only a subset of the RE
+
+    type in (:linpred, :response) || throw(ArgumentError("Invalid value for type: $(type)"))
+
+    if use_re && newX === m.X
+        # should we add a kwarg to fitted to allow returning eta?
```
dmbates, line 94: Seems reasonable.

src/predict.jl
```diff
@@ -0,0 +1,292 @@
...
+        type == :response && return fitted(m)
+        return m.resp.eta
```
dmbates, line 96: I would probably write those two lines as 
```suggestion
        return type == : response ? fitted(m) : m.resp.eta
```

src/simulate.jl
```diff
@@ -1,8 +1,46 @@
...
+See [`simulate!`](@ref)
+"""
+function simulate end
+
+function simulate(rng::AbstractRNG, m::MixedModel{T}, newX::AbstractMatrix = m.X;
+                  kwargs...) where {T}
+    size(newX, 1) == nobs(m) ||
+        throw(DimensionMismatch("New fixed-effect model matrix must have the same number of observations as the original."))
+    size(newX, 2) == size(m.X, 2) ||
+    throw(DimensionMismatch("New fixed-effect model matrix must have the same predictors as the original."))
```
dmbates, line 11: why not just check that `size(newX) == size(m.X)`?

dmbates: Also, the indentation of the `throw` lines is inconsistent.

src/simulate.jl
```diff
@@ -1,8 +1,46 @@
...
+    y = zeros(T, nobs(m))
+    simulate!(rng, y, m, newX; kwargs...)
```
dmbates, line 13: I think the next line, `y`, is redundant as `simulate!` returns the modified `y` argument.

src/simulate.jl
```diff
@@ -151,7 +332,7 @@ vector is generated as `randn(rng, size(M, 2))`
 """
 function unscaledre! end
```
dmbates, line 333: Perhaps we should change this name for some of the methods.  I kept trying to remember why it was called `unscaled` and realized that this only applies to the case where the random effects are being simulated so they get multiplied by λ but not by σ.  But the name doesn't apply to the cases shown here where `b` is being passed in.

dmbates (commented):
The write-up in the documentation is very good.  

I will need to study src/simulate.jl in more detail if that is okay.  There is a lot going on in there.

palday:
Please take your time. I can't allow myself to get nerd-sniped by this week anyway. :smile: Also, if there are obvious thinkos/typos/organizational issues and you think you can make the change faster than explaining it, feel free to push. (And sorry that the order of the `simulate` methods doesn't seem to have any underlying logic -- it's relatively clear that I did it piecewise over the course of several months....)

src/simulate.jl
dmbates: I realized what was bothering me about this.  The methods such as
```julia
unscaledre!(y, A, b)
```
could be written as 5-argument `mul!` methods.  Essentially the operation is
```julia
mul!(y, A, b, one(T), one(T))
```

I plan to make a modification where the `unscaledre!` methods are only those that simulate the data and the 5-argument `mul!` methods are in `src/remat.jl`.  Should I create a new branch from this branch so that you can look at it before merging back into this branch?

palday: That works for me, thanks.

dmbates: Resolved by #526

dmbates:
@palday Where are we on this PR?  Is there something I can do - other than a `git rebase` which I view as the single most dangerous activity known to man.

src/predict.jl
palday: It took me a minute to understand the pathology you're thinking of .... and I don't know. Looking at these now, I'm wondering whether the methods with `newX::Matrix` should just be internal. They seem like a nasty foot gun for novice users (because there's no checking that the new fixed effects matrix has columns that correspond to the coefficients in the model and you have to construct that matrix somehow) and I think it's less useful overall, except for their internal use. (Looking at the code, they are clearly internal methods that I exposed for reasons lost to me now.) In their internal use, the error checking for dimensionality matching happens before the methods are called / happens via construction of the new model, so then the problem goes away. 

The only use case this doesn't cover is predictions based purely on the fixed effects, but this is trivial for someone with tabular new data to do -- they just use the tabular methods, set `new_re_levels=:population` and use new levels of the grouping variable.

palday: Actually having said that, I think we can drop `predict(..., newX::Matrix)` completely -- looks like they're not even used internally in a meaningful way and can just be replaced with `mul!(y, X, coef(m))` in the appropriate place.

src/predict.jl
palday: yep, definitely better :)

src/predict.jl
```diff
@@ -0,0 +1,179 @@
...
+    end
+
+    y
+end
+
+function StatsBase.predict(m::GeneralizedLinearMixedModel{T}, newdata::Tables.ColumnTable;
+                           new_re_levels=:population, type=:response) where T
+    type in (:linpred, :response) || throw(ArgumentError("Invalid value for type: $(type)"))
+
+    y = predict(m.LMM, newdata; new_re_levels)
```
palday, line 168: @dmbates  I need to refactor this a bit so that I can share the BLUP extraction logic between both LMM and GLMM, but this isn't the correct way to do it.

palday:
@dmbates I've cut a lot of methods that I think aren't useful for most users. We can revisit them if there turns out to be an actual need for them. 

The biggest thing is that `predict` for GLMM doesn't quite work -- my late night shortcut wasn't a good idea and I haven't had a chance to look at it again in daylight hours. 

After that's done, I need to take another pass on editing the docs and making sure the test coverage, but that's quick and easy once everything else is stable and done.

src/predict.jl
palday: I think I've got it figured out, testing locally now

dmbates: Okay.  I'm just starting to read through all this again.  If you have suggestions of where I should start please let me know.

palday: I think I've got it refactored appropriately now and am tracking down some floating-point tolerance issues for `fast=false` fits.

I've pushed the changes so far.

src/predict.jl
dmbates: Seems like a good approach.  I am inclined to avoid passing around model matrices in general as there are just too many ways to botch the process of generating them.

docs/src/prediction.md
```diff
@@ -0,0 +1,172 @@
...
+
+!!! note
+    Currently, we do not support predicting based on a subset of the random effects.
+
+
+!!! note
+    `predict` is deterministic (within the constraints of floating point) and never adds noise to the result.
+    If you want to construct prediction intervals, then `simulate` will generate new data with noise (including new values of the random effects).
+
+For generalized linear mixed models, there is an additional keyword argument to `predict`: `type` specifies whether the predictions are returned on the scale of the linear predictor (`:linpred`) or a the level of the response `(:response)` (i.e. the level at which the values were originally observed).
```
dmbates, line 79:
```suggestion
For generalized linear mixed models, there is an additional keyword argument to `predict`: `type` specifies whether the predictions are returned on the scale of the linear predictor (`:linpred`) or on the level of the response `(:response)` (i.e. the level at which the values were originally observed).
```

docs/src/prediction.md
```diff
@@ -0,0 +1,172 @@
...
+```@example Main
+logit(x) = log(x / (1 - x))
+predict(gm, cbpp; type=:linpred) ≈ logit.(fitted(gm))
+```
+
+## Simulation
+
+In contrast to `predict`, `simulate` and `simulate!` introduce randomness.
+This randomness occurs both at the level of the observation-level (residual) variance and at the level of the random effects, where new conditional modes are sampled based on the specified covariance parameter (θ; see [Details of the parameter estimation](@ref)), which defaults to the estimated value of the model.
+For reproducibility, we specify a psuedorandom generator here; if non is provided, the global PRNG is taken as the default.
```
dmbates, line 96:
```suggestion
For reproducibility, we specify a pseudorandom generator here; if non is provided, the global PRNG is taken as the default.
```

docs/src/prediction.md
```diff
@@ -0,0 +1,172 @@
...
+
+The simplest example of `simulate` takes a fitted model and generates a new response vector based on the existing model matrices combined with noise.
+
+```@example Main
+using Random
+ynew = simulate(MersenneTwister(42), slpm)
+```
+
+The simulated response can also placed in a pre-allocated vector:
```
dmbates, line 105:
```suggestion
The simulated response can also be placed in a pre-allocated vector:
```

docs/src/prediction.md
```diff
@@ -0,0 +1,172 @@
...
+```@example Main
+logit(x) = log(x / (1 - x))
+predict(gm, cbpp; type=:linpred) ≈ logit.(fitted(gm))
+```
+
+## Simulation
+
+In contrast to `predict`, `simulate` and `simulate!` introduce randomness.
+This randomness occurs both at the level of the observation-level (residual) variance and at the level of the random effects, where new conditional modes are sampled based on the specified covariance parameter (θ; see [Details of the parameter estimation](@ref)), which defaults to the estimated value of the model.
+For reproducibility, we specify a pseudorandom generator here; if non is provided, the global PRNG is taken as the default.
```
dmbates, line 96:
```suggestion
For reproducibility, we specify a pseudorandom generator here; if none is provided, the global PRNG is taken as the default.
```

dmbates (approved):
- Not related to this particular PR but we should consider running the code through `JuliaFormatter.format`, probably with `BlueStyle()` but I am willing to hear other options, before releasing v4.0.0

- This looks great.  Thanks for all the work on this.

palday:
:smile: I still need to write a NEWS entry and do a few small docs generation tweaks.

dmbates:
I'm glad you remembered the News update.  I seem to have difficulty remembering it.

PR #428, opened 2020-11-05
Check BLAS vendor, fix test warning, rephrase warn

dmbates:
- Tests were failing under MKL because `openblas_get_config()` is not defined.
- Rephrase a warning message
- Move definition of `io` inside a `testset` in `test/utilities.jl` to avoid later warning on redefinition in `test/pls.jl`.

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -6,7 +6,6 @@ using Test
 
 using MixedModels: allequal, average, densify, dataset
 
-const io = IOBuffer()
 include("modelcache.jl")
 
 @testset "average" begin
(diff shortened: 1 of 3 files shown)


test/utilities.jl
```diff
@@ -60,6 +59,7 @@ end
 end
 
 @testset "PCA" begin
+	io = IOBuffer()
```
palday, line 62: I have a different fix for this in #427, but let's go with this version for now. :smile:

palday (approved):

palday:
Good for squash and merge as soon as the CIs finish.

test/utilities.jl
dmbates: I just saw that when I was looking through #427

PR #430, opened 2020-11-10
loosen floating point equality constraint

palday:
There is a very weird race condition that seems to arise in Julia 1.5 and which I can't consistently replicate.  I'm pretty sure the locking is correct. I suspect it's actually a floating point issue interacting in a fun way with threading. If it were only Julia threads there, then you would have the exact same floating point operations for each iteration, but the iterations potentially out of order (hence the sorting). But I think some of layers of optimization are breaking that assumption and that test should be made approximate but with a tight tolerance.

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -45,7 +45,7 @@ end
 		r = only(randn(rng, 1));
 	end
 
-	@test all(sort!(single_thread) .== sort!(multi_thread))
+	@test all(sort!(single_thread) .≈ sort!(multi_thread))
 end
 
 @testset "datasets" begin


test/utilities.jl
```diff
@@ -45,7 +45,7 @@ end
 		r = only(randn(rng, 1));
 	end
 
-	@test all(sort!(single_thread) .== sort!(multi_thread))
+	@test all(sort!(single_thread) .≈ sort!(multi_thread))
```
ararslan, line 48: Seems good to me. Worth specifying a tolerance explicitly do you think, or is the default sufficient? Also, you can do this if you feel so inclined:
```suggestion
	@test sort!(single_thread) ≈ sort!(multi_thread)
```
since there is an `isapprox` for arrays, though it uses `norm` rather than elementwise approximate equality, IIRC.

palday: I thought about that and think I want the elementwise comparison because our guarantee is on the elementwise equality.

I suspect the default tolerance is good enough -- it should reflect what floating-point equality can reasonably be expected to be, right? I definitely don't want a looser tolerance, but I have no idea how tight we can go -- 1 ups? 2 ups? 1e-16? I have no idea.

ararslan (approved):

palday:
Failure on Future is unrelated.

PR #431, opened 2020-11-11
Swap adjoint and triangular wrappers

dkarrasch:
Should fix the failure on nightly.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -41,52 +41,104 @@ LinearAlgebra.mul!(
     β::Number,
 ) where {T} = mul!(C, adjA.parent.cscmat', B, α, β)
 
-function LinearAlgebra.ldiv!(
-    adjA::Adjoint{T,<:LowerTriangular{T,UniformBlockDiagonal{T}}},
-    B::StridedVector{T},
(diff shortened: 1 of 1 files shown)


dkarrasch:
I understand it's a bit ugly with the version-dependent branches. But in fact, one can remove those branches and just have all those methods side by side, since they don't overwrite each other due to different signatures. Instead, one could have a comment saying that some (and which) methods can be removed once the lower bound on julia is v1.6. Would you prefer that? Sorry again for the inconvenience!

dmbates (approved):
I am fine with this fix.  Thanks @dkarrasch 

I agree that if we want to avoid conditioning on `VERSION` we could just define all the methods because they do not conflict.  However, I think it is better to leave the condition in there because it will remind us if we specify a dependence on Julia 1.6 or later to remove the old definitions.

palday:
@dkarrasch Thanks for addressing this so quickly. I'm being a big grump about all of it, but I appreciate the speed and effort you've put into getting us a fix. :smile: 

I was torn on the issue of the branch, so I'll go with @dmbates on it.

PR #432, opened 2020-11-13
Force BlasInt type for piv

dmbates:
When using [`MKL`](https://github.com/JuliaComputing/MKL.jl) on a 64-bit Ubuntu system I found that I needed to set
```
ENV["USE_BLAS64"]=true
```
before building `MKL` or there would be a missing method error in the call to `QRPivoted`.  I believe that, by default, MKL uses 32-bit integers but OpenBLAS uses 64-bit integers.  The `piv` vector was being constructed as `collect(1:n)` hence it would default to 64-bit integers.

I think it is because the second external constructor for `QRPivoted` in the `LinearAlgebra` package
```
function QRPivoted{T}(factors::AbstractMatrix, τ::AbstractVector, jpvt::AbstractVector) where {T}
    QRPivoted(convert(AbstractMatrix{T}, factors),
              convert(Vector{T}, τ),
              convert(Vector{BlasInt}, jpvt))
end
```
which would change the eltype of `jpvt` isn't being called in this case.  IIUC it would need to be called with an explicit type `T`, as in
```
QRPivoted{Float64}(...)
```
because `T` can't be inferred from the arguments to the function.

@andreasnoack Does this make sense?

I threw in a second change related to MKL.  I thought that the symbol returned from `BLAS.vendor()` was `:openblas` but, in fact, it is `:openblas64` or `:openblas32`.  Should the symbol returned for MKL also be changed to `:mkl32` or `:mkl64`?

### DIFF
--- a/src/linalg/pivot.jl
+++ b/src/linalg/pivot.jl
@@ -10,7 +10,7 @@ function statscholesky(xtx::Symmetric{T}, tol::Real = -1) where {T<:AbstractFloa
     chpiv = cholesky(xtx, Val(true), tol = T(tol), check = false)
     chunp = cholesky(xtx, check = false);
 
-    piv = [1:n;]
+    piv = LinearAlgebra.BlasInt.(collect(1:n))
     r = chpiv.rank
 
     if r < n
@@ -58,7 +58,7 @@ function statsqr(x::Matrix{T}; ranktol=1e-8) where {T<:AbstractFloat}
     if rank < n
         piv = qrpiv.p
     else
-        piv = collect(1:n)
+        piv = LinearAlgebra.BlasInt.(collect(1:n))
     end
 
     if rank < n && piv[1] ≠ 1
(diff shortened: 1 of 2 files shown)


palday:
I think this make sense -- should we add a MKL CI?

palday:
(where we = me)

dmbates:
> I think this make sense -- should we add a MKL CI?

How would that work?  I don't think the standard CI bots have a built-in Julia version with MKL and it takes a long time to install MKL because any packages that use Lapack/BLAS have to be recompiled.

palday:
See #433. :smile: 

It's not a quick run at all, but the trick is to install MKL before installing any dependencies.  I think it might be one to only run on demand (i.e. manual triggering, not for every PR). It's already highlighting how fragile all the pivoting stuff still is, unfortunately.

palday (approved):

dmbates:
Should this PR be merged separately or as part of #433 ?

palday:
I would merge separately.

PR #433, opened 2020-11-13
CI using MKL

palday:
I never get CI stuff to work the first time around, so very WIP.

Note that this only uses Julia 1.5 on Ubuntu 18.04 x86-64, but testing the full combinatorial explosion of MKL possibilities is just too much compute time/effort.

### DIFF
--- a/.github/workflows/MKL.yml
+++ b/.github/workflows/MKL.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
       - uses: julia-actions/julia-buildpkg@v0.1
-      - run:  julia --project -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
+      - run:  julia -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
       - uses: julia-actions/julia-runtest@v0.1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/MKL.yml
+++ b/.github/workflows/MKL.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
       - uses: julia-actions/julia-buildpkg@v0.1
-      - run:  julia --project='.' -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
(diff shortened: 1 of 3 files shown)


.github/workflows/MKL.yml
```diff
@@ -0,0 +1,27 @@
...
+        julia-version: [1.5]
+        julia-arch: [x64]
+        os: [ubuntu-18.04]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.julia-version }}
+      - uses: julia-actions/julia-buildpkg@v0.1
+      - run:  julia --project='.' -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
```
palday, line 24:
```suggestion
      - run:  julia --project -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
```

.github/workflows/MKL.yml
```diff
@@ -0,0 +1,27 @@
...
+        julia-version: [1.5]
+        julia-arch: [x64]
+        os: [ubuntu-18.04]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.julia-version }}
+      - uses: julia-actions/julia-buildpkg@v0.1
+      - run:  julia --project -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
```
palday, line 24:
```suggestion
      - run:  julia -e 'using Pkg; Pkg.add("MKL"); Pkg.build("MKL"; verbose=true)'
```

palday:
I've made the MKL CI run only when explicitly triggered, but for an example of how a run looks, see [here](https://github.com/JuliaStats/MixedModels.jl/runs/1396985288?check_suite_focus=true).

dmbates (approved):
Looks good, thanks.  Do you know what the cause of the CI failure is?

palday:
It's the pivoting again:
```
 fullrankcategorical |    2      2
dependentcolumn: Test Failed at /home/runner/work/MixedModels.jl/MixedModels.jl/test/pivot.jl:37
  Expression: st.p == perm
   Evaluated: Int32[1, 3, 4, 2] == [1, 2, 4, 3]
```

I think I know how to fix this one case, but my preference would be to ignore it for now. I'm not sure I can deal with the frustrations of numerical rank deficiency right now ....

PR #441, opened 2020-11-27
Allow v1.0.0 of StaticArrays and Arrow

dmbates:
Probably best to hold off on merging this until BlockArrays.jl and Distributions.jl catch up to StaticArrays.jl v1.0.0


dmbates:
CI failure on "future" is due to the switch from Pkg.Artifacts to LazyArtifacts in v1.6.0.  I have a question on discourse about the recommended way to handle it.

palday:
I saw that failure come up in a recent PR for a doc change. 

All this looks good to me.

palday (approved):

palday:
It looks like StaticArrays 1.0.0 was used inte CI, so I'm good to merge when you are.

PR #444, opened 2020-12-02
Eagerness to avoid LazyArtifacts 

palday:
It seems challenging to have lazy artifact downloading be compatible with Julia < 1.6 and 1.6+, see recent test failures and [Discourse](https://discourse.julialang.org/t/recommended-transition-to-lazyartifacts/). However, we use the `TestData` artifact in `__init__` so they're downloaded with the very first `using MixedModels`, which we can assume most users will want to do not too much after `]add MixedModels`. So we can avoid all the laziness issues and just be eager. Slightly longer download times upon initial install, but since the artifacts system has its own caching, there shouldn't be any real performance issues as long as we don't change `TestData` too often.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -10,7 +10,7 @@
 # println("git-tree-sha1: ", Tar.tree_hash(IOBuffer(inflate_gzip(filename))))
 # from https://julialang.github.io/Pkg.jl/dev/artifacts/
 git-tree-sha1 = "75260d131b693f26e5834adf855f4c35d627348d"
-lazy = true
+lazy = false
 
     [[TestData.download]]
     # this is the SHA from https://osf.io/djaqb/download?version=4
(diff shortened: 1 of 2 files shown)


palday:
Also, this makes me think we should slightly change our support policy:

Tier 1: Last LTS release
Tier 2: Current release

All of this subject to the condition that there aren't critical new features in `LinearAlgebra` and `Distributed` or `Threads` in a non LTS release.

Currently, our plan is to take current Tier 1 (1.4, 1.5) and make it Tier 2 upon the 1.6 LTS release and make the LTS Tier 1.  This is changing plans a bit and going back on a vague support committment, but we also haven't really defined what Tier 1 and Tier 2 support mean anyway...

dmbates (approved):
Thanks for picking up on the use of `lazy` in the `Artifacts.toml` file.  If I understand correctly, it would not have any effect to change it back to `lazy=true` for julia v1.6.0+ because the artifact is used in the `__init__` function.  Would the only difference be that the artifact is downloaded at installation time rather than when the package is first used?  If so, it doesn't seem very important to make it lazy.

palday:
@dmbates Your assessment is exactly correct.

Issue #445, opened 2020-12-03
Nested models


Jejulia:
Hi:
I was fitting models with nested grouping factors, like `y ~ x + (1|A) + (1|A&B)`, but errors occurred:
```
ERROR: LoadError: MethodError: no method matching strides(::Diagonal{Float64,Array{Float64,1}})  
...
Stacktrace:
 [1] stride(::Diagonal{Float64,Array{Float64,1}}, ::Int64) at .\abstractarray.jl:396
 [2] stride1(::Diagonal{Float64,Array{Float64,1}}) at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\LinearAlgebra.jl:197
 [3] _chkstride1 at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\LinearAlgebra.jl:203 [inlined]
 [4] chkstride1 at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\LinearAlgebra.jl:201 [inlined]
 [5] syrk!(::Char, ::Char, ::Float64, ::Array{Float64,2}, ::Float64, ::Diagonal{Float64,Array{Float64,1}}) at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\blas.jl:1594
 [6] rankUpdate!(::Hermitian{Float64,Diagonal{Float64,Array{Float64,1}}}, ::Array{Float64,2}, ::Float64, ::Float64) at C:\Users\sciph\.julia\packages\MixedModels\36RJL\src\linalg\rankUpdate.jl:24
...
```
I've tried different datasets (which worked in `lme4` or `nlme`), but same errors kept occuring. 

It seems that the problem is using `BLAS.syrk!` which will eventually stride the diagonal matrix.
Is there an alternative way to avoid `Diagonal` or other suggestions? Thanks!

palday:
If the problem is fundamentally at `strides`, then it's a `LinearAlgebra` and not `MixedModels` issue. But first we need more information to see if something is going wrong before then:

- What Julia version? (looks like 1.5 from the stack trace; which patch version?)
- What MixedModels version?
- Can you share a minimum working example or (even privately) your data so that we can test and debug this issue?

Jejulia:
I am using Julia 1.5.3 and `MixedModels@3.1.1`
[...]

PR #447, opened 2020-12-04
rankUpdate for BlockDiagonal by Dense

palday:
Another one from #445.

### DIFF
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -64,10 +64,9 @@ end
 function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T}, α, β) where {T,S}
     Cdat = C.data.data
     isone(β) || (Cdat .*= β)
-    nblks = size(Cdat, 3)
     blksize = size(Cdat, 1)
 
-    for k in 1:nblks
+    for k in axes(Cdat, 3)
         ioffset = (k - 1) * blksize
         joffset = (k - 1) * blksize
         for i in 1:blksize, j in 1:i
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -63,7 +63,7 @@ end
 
 function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T}, α, β) where {T,S}
     Cdat = C.data.data
-    @. Cdat = β * Cdat
+    isone(β) || (Cdat .*= β)
     nblks = size(Cdat, 3)
     blksize = size(Cdat, 1)
 
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -70,7 +70,7 @@ function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T
     for k in 1:nblks
         ioffset = (k - 1) * blksize
         joffset = (k - 1) * blksize
-        for i in 1:blksize, j in 1:blksize
+        for i in 1:blksize, j in 1:i
             iind = ioffset + i
             jind = joffset + j
             Arow = view(A, iind, :)
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -61,6 +61,27 @@ function rankUpdate!(C::HermOrSym{T,Diagonal{T,Vector{T}}}, A::StridedMatrix{T},
     C
 end
 
+function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T}, α, β) where {T,S}
+    Cdat = C.data.data
+    @. Cdat = β * Cdat
+    nblks = size(Cdat, 3)
+    blksize = size(Cdat, 1)
+
+    for k in 1:nblks
+        ioffset = (k - 1) * blksize
+        joffset = (k - 1) * blksize
+        for i in 1:blksize, j in 1:blksize
+            iind = ioffset + i
+            jind = joffset + j
+            Arow = view(A, iind, :)
+            Acol = view(A, jind, :) # because the column comes from A'
+            Cdat[i,j,k] = Cdat[i,j,k] + α * Arow'Acol
+        end
+    end
+
+    C
+end
+
 function rankUpdate!(
     C::HermOrSym{T,Diagonal{T,Vector{T}}},
     A::SparseMatrixCSC{T},
--- a/src/linalg/rankUpdate.jl
+++ b/src/linalg/rankUpdate.jl
@@ -72,9 +72,12 @@ function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T
         for i in 1:blksize, j in 1:i
             iind = ioffset + i
             jind = joffset + j
-            Arow = view(A, iind, :)
-            Acol = view(A, jind, :) # because the column comes from A'
-            Cdat[i,j,k] = Cdat[i,j,k] + α * Arow'Acol
+            AtAij = 0
+            for idx in axes(A, 2)
+                # because the second is actually A', we swap index orders
+                AtAij += A[iind,idx] * A[jind,idx]
+            end
+            Cdat[i,j,k] += α * AtAij
         end
     end
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 MixedModels v3.1.3 Release Notes
 ========================
 * Compatibility updates
-* `rankUpdate!` method for `BlockDiagonal` by `Dense` [#447]
+* `rankUpdate!` method for `UniformBlockDiagonal` by `Dense` [#447]
 
 MixedModels v3.1.2 Release Notes
 ========================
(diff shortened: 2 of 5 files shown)


palday:
@kleinschmidt While working on this, I discovered a second issue.

These two formulae *should* give the same model:
```julia
julia> fnested = @formula(Thickness ~ 1 + Source + (1+Source|Lot/Wafer))
FormulaTerm
Response:
  Thickness(unknown)
Predictors:
  1
  Source(unknown)
  (Source,Lot,Wafer)->(1 + Source) | Lot / Wafer

julia> f = @formula(Thickness ~ 1 + Source + (1+Source|Lot) + (1+Source|Lot&Wafer))
FormulaTerm
Response:
  Thickness(unknown)
Predictors:
  1
  Source(unknown)
  (Source,Lot)->(1 + Source) | Lot
  (Source,Lot,Wafer)->(1 + Source) | Lot & Wafer
```

But `fnested` fails in construction due to schema problems. I don't think we ever tested the nested RE stuff on a model with a random slope:
```julia
julia> fit(MixedModel, fnested, MixedModels.dataset(:oxide))
ERROR: MethodError: no method matching +(::Tuple{InterceptTerm{true},CategoricalTerm{DummyCoding,String,1}}, ::Tuple{RandomEffectsTerm,RandomEffectsTerm})
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:538
  +(::MutableArithmetics.Zero, ::Any) at /home/XXX/.julia/packages/MutableArithmetics/0tlz5/src/rewrite.jl:52
  +(::Tuple{Vararg{Union{AbstractTerm, Tuple{Vararg{AbstractTerm,N}} where N},N}} where N, ::AbstractTerm) at /home/XXX/.julia/packages/StatsModels/cpRh6/src/terms.jl:404
  ...
Stacktrace:
 [1] +(::InterceptTerm{true}, ::CategoricalTerm{DummyCoding,String,1}, ::Tuple{RandomEffectsTerm,RandomEffectsTerm}) at ./operators.jl:538
 [2] sum(::Tuple{InterceptTerm{true},CategoricalTerm{DummyCoding,String,1},Tuple{RandomEffectsTerm,RandomEffectsTerm}}) at ./tuple.jl:396
 [3] apply_schema(::Tuple{ConstantTerm{Int64},Term,FunctionTerm{typeof(|),var"#107#109",(:Source, :Lot, :Wafer)}}, ::MixedModels.MultiSchema{StatsModels.FullRank}, ::Type{T} where T) at /home/XXX/Work/MixedModels.jl/src/schema.jl:30
 [4] apply_schema(::FormulaTerm{Term,Tuple{ConstantTerm{Int64},Term,FunctionTerm{typeof(|),var"#107#109",(:Source, :Lot, :Wafer)}}}, ::StatsModels.Schema, ::Type{LinearMixedModel}) at /home/XXX/Work/MixedModels.jl/src/schema.jl:53
 [5] LinearMixedModel(::FormulaTerm{Term,Tuple{ConstantTerm{Int64},Term,FunctionTerm{typeof(|),var"#107#109",(:Source, :Lot, :Wafer)}}}, ::NamedTuple{(:Source, :Lot, :Wafer, :Site, :Thickness),Tuple{Arrow.DictEncoded{String,UInt32,SentinelArrays.ChainedVector{String,Arrow.List{String,Int32,Array{UInt8,1}}}},Arrow.DictEncoded{String,UInt32,SentinelArrays.ChainedVector{String,Arrow.List{String,Int32,Array{UInt8,1}}}},Arrow.DictEncoded{String,UInt32,SentinelArrays.ChainedVector{String,Arrow.List{String,Int32,Array{UInt8,1}}}},Arrow.DictEncoded{String,UInt32,SentinelArrays.ChainedVector{String,Arrow.List{String,Int32,Array{UInt8,1}}}},Arrow.Primitive{Float64,Array{Float64,1}}}}; contrasts::Dict{Symbol,Any}, wts::Array{Any,1}) at /home/XXX/Work/MixedModels.jl/src/linearmixedmodel.jl:67
```

src/linalg/rankUpdate.jl
```diff
@@ -61,6 +61,27 @@ function rankUpdate!(C::HermOrSym{T,Diagonal{T,Vector{T}}}, A::StridedMatrix{T},
     C
 end
 
+function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T}, α, β) where {T,S}
+    Cdat = C.data.data
+    @. Cdat = β * Cdat
+    nblks = size(Cdat, 3)
+    blksize = size(Cdat, 1)
+
+    for k in 1:nblks
```
dmbates, line 70: Prefer 
```suggestion
    for k in axes(Cdat, 3)
```

src/linalg/rankUpdate.jl
```diff
@@ -61,6 +61,27 @@ function rankUpdate!(C::HermOrSym{T,Diagonal{T,Vector{T}}}, A::StridedMatrix{T},
     C
 end
 
+function rankUpdate!(C::HermOrSym{T,UniformBlockDiagonal{T}}, A::StridedMatrix{T}, α, β) where {T,S}
+    Cdat = C.data.data
+    @. Cdat = β * Cdat
```
dmbates, line 66:
```suggestion
    isone(β) || (Cdat .*= β)
```

src/linalg/rankUpdate.jl
```diff
@@ -61,6 +61,27 @@ function rankUpdate!(C::HermOrSym{T,Diagonal{T,Vector{T}}}, A::StridedMatrix{T},
...
+    nblks = size(Cdat, 3)
+    blksize = size(Cdat, 1)
+
+    for k in 1:nblks
+        ioffset = (k - 1) * blksize
+        joffset = (k - 1) * blksize
+        for i in 1:blksize, j in 1:blksize
```
dmbates, line 73:
```suggestion
        for i in 1:blksize, j in 1:i
```
Only need to update the lower triangle of `C`.

src/linalg/rankUpdate.jl
```diff
@@ -61,6 +61,27 @@ function rankUpdate!(C::HermOrSym{T,Diagonal{T,Vector{T}}}, A::StridedMatrix{T},
...
+            iind = ioffset + i
+            jind = joffset + j
+            Arow = view(A, iind, :)
```
dmbates, line 76: I would probably write this as an explicit loop over `axes(A, 2)` and avoid the construction of views.

dmbates (approved):
A few stylistic comments, none of which are terribly important.  Do with them what you will.

src/linalg/rankUpdate.jl
palday: Oh, good trick. I wasn't even thinking that it's not just `UniformBlockDiagonal`, but also `Symmetric`.

src/linalg/rankUpdate.jl
palday: So something like this:
```julia
AtAij = 0
for idx in axes(A, 2)
    # because the second is actually A', we swap index orders
    AtAij += A[iind,idx] * A[jind,idx]
end
Cdat[i,j,k] = Cdat[i,j,k] + α * AtAij
```
? 
I also thought about writing this as
```julia
AtAij = sum(A[iind,idx] * A[jind,idx] for idx in axes(A,2))
```
but then I realized the construction of the generator is probably no better (and maybe worse) than the construction of views.

dmbates: Actually what I had in mind is something like
```
for idx in axes(A, 2)
    Cdat[i,j,k] += α * A[iind, idx] * A[jind, idx]
end
```
I'm pretty sure the compiler will detect that the location `Cdat[i,j,k]` is a loop invariant.

palday: Ah, interesting. 

That might be a decent research project if we ever get an intern -- taking a look at `@code_llvm` and `@code_native` for some of our loops and seeing where the Julia compiler and LLVM are doing nice vectorization and other loop transformations for.

dmbates: An alternative is to resize `Cdat` to be a matrix with `blksize` rows.  Then the columns match up with the columns of `C`.  Also there is an intriguing function called `fldmod1` which makes it easy to step over the blocks but I have forgotten the details.  I'll take a look.

palday: If we're going to resize in that way -- would it make sense to change the internal storage used by `UniformBlockDiagonal` to always store in that format?

kleinschmidt:
That's really strange, I don't see why that wouldn't work.  It looks like it's not failing at the the random slope point, but rather at the point where it's trying to combine the two RE terms with the FE terms...but that path should already be hit by the existing tests, so I'm confused.  I'll try to take a look when I get a chance.

palday:
@kleinschmidt Thanks! Glad to know at least the problem isn't something obvious I should have been able to fix in 10 minutes...

src/linalg/rankUpdate.jl
dmbates: There are times that you want the diagonal blocks to be accessed as k by k matrices.  The most important part is in evaluating `Lambda'Z'ZLambda + I`.  Also, the solving of multiple triangular systems to produce the `[i,1]` blocks of L for `i > 1`.  I think it is easier to resize from 3 dimensions to 2 when we need it, rather than the other way around.

PR #449, opened 2020-12-09
Modularize constructors

palday:
This splits our main (non default) constructor into several stages. The idea is that packages like [`unfold.jl`](https://github.com/unfoldtoolbox/unfold.jl/blob/fcbaa8e0b67fada2b3b3d745d114acbd7a207ffe/src/fit.jl#L199-L229) that do a lot of the model matrix construction can still take advantage of the rest of our machinery and not duplicate effort in a brittle way.

@behinger

### DIFF
--- a/.github/workflows/Future.yml
+++ b/.github/workflows/Future.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   ci:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


palday:
I think this should be a patch release because these changes are officially marked as being "internal", even those some of our collaborators are going to use these internal changes as a versioning requirement.

dmbates (approved):
Looks good.  Thanks.

PR #456, opened 2021-01-16
Use vector of matrices for A and L in LinearMixedModel

dmbates:
- `BlockArray`s were only being used to store and access different matrices that formed the blocks of the `A` and `L` components of a `LinearMixedModel` object.
- Replace the `BlockArray`s with  `Vector{AbstractMatrix{T}}` and functions to convert an `[i,j]` index in the lower triangle to an index into the vector
- Within the vector, the blocks in the lower triangle are stored in _row-major_ order because this simplifies the conversion of row-column to linear index
- Because this change will require a new major release number will consider also changing to a single `Xy` block instead of separate blocks.  Nothing in the update of the Cholesky factor requires separate handling of the blocks associated with `X` and with `y`.

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -546,7 +546,7 @@ function StatsBase.leverage(m::LinearMixedModel{T}) where {T}
         fill!(m.y, zero(T))
         m.y[i] = one(T)
         updateL!(reevaluateAend!(m))
-        sum(j -> sum(abs2, view(m.L, Block(l, j))), 1:(l-1))
+        sum(j -> sum(abs2, m.L[packedlowertri(l, j)]), 1:(l-1))
     end
     copyto!(m.y, yorig)
     updateL!(reevaluateAend!(m))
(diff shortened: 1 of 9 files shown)


src/linearmixedmodel.jl
```diff
@@ -314,7 +315,8 @@ function createAL(allterms::Vector{Union{AbstractReMat{T},FeMat{T}}}) where {T}
             cj = allterms[j]
             if !isnested(cj, ci)
                 for l = i:k
-                    L[Block(l, i)] = Matrix(getblock(L, l, i))
+                    ind = packedlowertri(l, i)
+                    L[ind] = Matrix(L[ind])
```
palday, line 319: Should we add a `getblock` method or something similar that just wraps the use of `packedlowertri`?

src/linearmixedmodel.jl
```diff
@@ -425,13 +426,13 @@ the length of `v` can be the rank of `X` or the number of columns of `X`.  In th
 case the calculated coefficients are padded with -0.0 out to the number of columns.
 """
 function fixef!(v::AbstractVector{T}, m::LinearMixedModel{T}) where {T}
-    Xtrm = first(m.feterms)
+    Xtrm = m.allterms[m.dims.nretrms + 1]
```
palday, line 429: why change this now?

(just curious)

src/linearmixedmodel.jl
dmbates: That probably would have been the simplest approach conceptually but I was in the middle of changing all those `getblock` calls to indexing with `packedlowertri` before it occurred to me.  Maybe it would be best to wait to see if and how the Xy amalgamation works out before deciding on a `getblock` strategy.

src/linearmixedmodel.jl
dmbates: Again, I am thinking in terms of removing the `reterms` and `feterms` fields of `LinearMixedModel` entirely in favor of the principal that the `allterms` field always consists of a vector of `AbstractReMat{T}` terms followed by exactly one `XyMat` term.  There may be a separate field with a single FeMat like the current type (w/o the `wtx` field) to be able to handle rank-deficient cases.

So here I was planning that the expression would eventually be replaced by `last(m.allterms)`.

dmbates:
@palday Feel free to review this for merging into master/main .  If you want to use an alternative to `packedlowertri` please go ahead and do so.

palday:
@dmbates I think I would just add change `packedlowertri` to `block` just for brevity -- I can do the find and replace if the new name works for you.

dmbates:
Sounds good - go for it.

palday (approved):

dmbates:
@palday Now that this has been merged should we change the version number to 4.0.0 so that anyone who is working off of the master branch will be aware that there are changes afoot.

PR #464, opened 2021-02-13
Xy block

dmbates:
- Replace the last two blocks, `X` and `y`, in the terms field with a single `femat` matrix that includes both `X` and `y`
- This cuts down on the number of calls to multiplication, etc. in `updateL!` and allows multi-threaded BLAS to be more effective
- This PR is incomplete


src/linearmixedmodel.jl
```diff
@@ -731,16 +735,20 @@ end
 """
     reevaluateAend!(m::LinearMixedModel)
 
-Reevaluate the last column of `m.A` from `m.feterms`.  This function should be called
-after updating the response, `m.feterms[end]`.
+Reevaluate the last column of `m.A` from `m.feterm`.  This function should be called
+after updating the response, `m.fe[end]`.
```
palday, line 739: There isn't a field `fe`, which ties into another thing I was thinking. Namely, that we should think about making the documentation for the `femat` vs `feterm` field very clear (and also clarify why we need both).

dmbates: Yes, absolutely.  I would still keep open the possibility of not using this change.  I think it will be helpful but it may cause too much pain with the sparse model matrices.  We'll see.  I just wanted to get it to a place where you can look at it.

dmbates:
- Changes in the GLMM code to accommodate the new structures.
- Several changes are because the GLM package specializes some methods to `Vector{T}`.  If those were changed to `AbstractVector{T}` the changes would not be necessary.
- Some of the changes of arguments like `foo.bar.baz'` to `adjoint(foo.bar.baz)` were simply to get the linter to shut up.  They were being flagged as syntax errors despite the fact that they were not syntax errors.

palday:
Thanks for the explanatory comments. :) Should we open a PR against GLM for changing loosening the type requirements?

test/pls.jl
```diff
@@ -92,7 +92,7 @@ include("modelcache.jl")
 
     @test logdet(fm1) ≈ 8.06014522999825 atol=0.001
     @test varest(fm1) ≈ 2451.2501089607676 atol=0.001
-    @test pwrss(fm1) ≈ 73537.50152584909 atol=0.001
+    @test pwrss(fm1) ≈ 73537.50152584909 atol=0.01 # this quantity is not precisely estimated
```
palday, line 95: thanks for adding the comment with the change!

dmbates:
@palday I am done tinkering.  Now if you could just do the heavy lifting of resolving conflicts, etc.

src/linalg.jl
```diff
@@ -31,15 +31,15 @@ LinearAlgebra.mul!(
     adjB::Adjoint{T,<:BlockedSparse{T}},
     α::Number,
     β::Number,
-) where {T} = mul!(C, A, adjB.parent.cscmat', α, β)
+) where {T} = mul!(C, A, adjoint(adjB.parent.cscmat), α, β)
```
palday, line 34: was this just because of the linter complaining?

dmbates: Yes.  I think they may have even fixed the issue in the linter used with VS Code now.  Your call on whether you want to revert this change or not.

src/Xymat.jl
```diff
@@ -0,0 +1,130 @@
+"""
+    FeTerm{T,S}
+
+Term with an explicit, constant matrix representation
+
+# Fields
```
palday, line 6:
```suggestion
Typically, an `FeTerm` represents the design matrix for the fixed effects. 

!!! note
    `FeTerm` is not the same as [`FeMat`](@ref)!

# Fields
```

I don't know if `FeTerm` is the best name because it doesn't work like a `StatsModels.AbstractTerm`. Maybe make this `FeMat` and `FeMat` into `FeResponseMat` or `XYMat` or `XyMat`?

src/Xymat.jl
```diff
@@ -0,0 +1,130 @@
...
+Does `A` have full column rank?
+"""
+isfullrank(A::FeTerm) = A.rank == length(A.piv)
+
+"""
+    FeMat{T,S}
+
+A matrix and a (possibly) weighted copy of itself.
+
+# Fields
```
palday, line 82:
```suggestion

Typically, an `FeMat` represents the fixed-effects design matrix with the response (`y`) concatenated as a final column.

!!! note
    `FeMat` is not the same as [`FeTerm`](@ref).

# Fields
```

src/Xymat.jl
palday: Having now also talked to @kleinschmidt , I think we should change at least one of the names. 

More generally, I'm not sure `FeTerm` and `ReTerm` are ideal, but we're internally consistent in our use of those even though they're not `Term`s in the StatsModel sense, but that latter bit is for a different PR. We could change those names or make them more term like, but again, that's something for a later day.

src/Xymat.jl
```diff
@@ -0,0 +1,130 @@
...
+end
+
+LinearAlgebra.rank(A::FeTerm) = A.rank
+
+"""
+    isfullrank(A::FeTerm)
+
+Does `A` have full column rank?
+"""
+isfullrank(A::FeTerm) = A.rank == length(A.piv)
```
palday, line 75: The pivot vector will always be the same length as the dimension, so this reduces to checking the rank against the dimension.

src/Xymat.jl
```diff
@@ -0,0 +1,130 @@
...
+    FeMat{T,typeof(xy)}(xy, xy)
+end
+
+Base.adjoint(A::FeMat) = Adjoint(A)
+
+Base.eltype(::FeMat{T}) where {T} = T
+
+Base.getindex(A::FeMat, i, j) = getindex(A.xy, i, j)
+
+Base.length(A::FeMat) = length(A.wtxy)
```
palday, line 104:
```suggestion
Base.length(A::FeMat) = length(A.xy)
```

More consistent if weight-independent ops always operate on the nonweighted field.

src/Xymat.jl
```diff
@@ -0,0 +1,130 @@
...
+            A.wtxy = similar(A.xy)
+        end
+        mul!(A.wtxy, Diagonal(sqrtwts), A.xy)
+    end
+    A
+end
+
+Base.size(A::FeMat) = size(A.xy)
+
+Base.size(A::FeMat, i::Integer) = size(A.xy, i)
```
palday, line 130: We have a few of these `Base` methods we're extending just by delegating -- should we put them in a loop and delegate on the symbols? 

No strong preference either way on my end.

src/Xymat.jl
```diff
@@ -0,0 +1,130 @@
...
+isfullrank(A::FeTerm) = A.rank == length(A.piv)
+
+"""
+    FeMat{T,S}
+
+A matrix and a (possibly) weighted copy of itself.
+
+# Fields
+- `xy`: original matrix, called `xy` b/c in practice this is `hcat(fullrank(X), y)`
+- `wtxy`: (possibly) weighted copy of `xy`
```
palday, line 84:
```suggestion
- `wtxy`: (possibly) weighted copy of `xy` (shares storage with `xy` until weights are applied)
```

src/linearmixedmodel.jl
```diff
@@ -6,9 +6,9 @@ Linear mixed-effects model representation
 ## Fields
 
 * `formula`: the formula for the model
-* `allterms`: a vector of random-effects terms, the fixed-effects terms and the response
 * `reterms`: a `Vector{AbstractReMat{T}}` of random-effects terms.
-* `feterms`: a `Vector{FeMat{T}}` of the fixed-effects model matrix and the response
+* `Xymat`: horizontal concatenation of a full-rank fixed-effects model matrix `X` and response `y` as an `FeMat{T}`
+* `feterm`: the fixed-effects model matrix as an `FeMat{T}`
```
palday, line 11:
```suggestion
* `feterm`: the fixed-effects model matrix as an `FeTerm{T}`
```

src/linearmixedmodel.jl
```diff
@@ -296,23 +291,28 @@ function condVar(m::LinearMixedModel{T}) where {T}
...
-    for i = 1:k
-        for j = 1:i
-            Lij = densify(allterms[i]' * allterms[j])
-            push!(L, Lij)
-            push!(A, deepcopy(isa(Lij, BlockedSparse) ? Lij.cscmat : Lij))
+function pushALblock!(A, L, blk)
+    push!(L, blk)
+    push!(A, deepcopy(isa(blk, BlockedSparse) ? blk.cscmat : blk))
+end
+function createAL(reterms::Vector{AbstractReMat{T}}, Xy::FeMat{T}) where {T}
```
palday, line 298:
```suggestion
end

function createAL(reterms::Vector{AbstractReMat{T}}, Xy::FeMat{T}) where {T}
```

src/linearmixedmodel.jl
```diff
@@ -296,23 +291,28 @@ function condVar(m::LinearMixedModel{T}) where {T}
...
-    A = AbstractMatrix{T}[]
-    L = AbstractMatrix{T}[]
-    for i = 1:k
-        for j = 1:i
-            Lij = densify(allterms[i]' * allterms[j])
-            push!(L, Lij)
-            push!(A, deepcopy(isa(Lij, BlockedSparse) ? Lij.cscmat : Lij))
+function pushALblock!(A, L, blk)
+    push!(L, blk)
+    push!(A, deepcopy(isa(blk, BlockedSparse) ? blk.cscmat : blk))
```
palday, line 296: Since all the calls to `pushALblock!` have the `blk` wrapped in `densify`, I wonder if it might make sense to have a `copy` option in that instead of `deepcopy` here. For matrices that are made dense, we're already dealing with a deep copy and so we add an extra copy here. For non dense matrices, we still need the copy of course.

src/linearmixedmodel.jl
```diff
@@ -296,23 +291,28 @@ function condVar(m::LinearMixedModel{T}) where {T}
...
     end
-    nretrm = sum(Base.Fix2(isa, ReMat), allterms)
-    for i = 2:nretrm      # check for fill-in due to non-nested grouping factors
-        ci = allterms[i]
-        for j = 1:(i-1)
-            cj = allterms[j]
+    for j in eachindex(reterms)   # can't fold this into the previous loop b/c block order
+        pushALblock!(A, L, densify(Xy' * reterms[j]))
+    end
+    pushALblock!(A, L, densify(Xy'Xy))
```
palday, line 311: I'm very curious whether we any sparse FE design matrix actually gets stored as sparse after the product here....

src/linearmixedmodel.jl
```diff
@@ -350,7 +350,10 @@ Return the lower Cholesky factor for the fixed-effects parameters, as an `LowerT
 `p × p` matrix.
 """
 function feL(m::LinearMixedModel)
-    LowerTriangular(m.L[kp1choose2(m.dims.nretrms + 1)])
+    XyL = m.L[end]
+    k = size(XyL, 1)
+    inds = Base.OneTo(k - 1)
+    LowerTriangular(view(XyL, inds, inds))
```
palday, line 356: more general question (i.e. not a change for this PR): I wonder if we should better document when our accessor methods are returning views / shallow copies and when they're returning deep copies.

src/linearmixedmodel.jl
```diff
@@ -6,9 +6,9 @@ Linear mixed-effects model representation
 ## Fields
 
 * `formula`: the formula for the model
-* `allterms`: a vector of random-effects terms, the fixed-effects terms and the response
```
palday, line 9: with the new Xy storage, `allterms` no longer provides a speed boost?

src/linearmixedmodel.jl
```diff
@@ -574,14 +590,7 @@ function mkparmap(reterms::Vector{AbstractReMat{T}}) where {T}
...
 
-function StatsBase.modelmatrix(m::LinearMixedModel)
-    fe = first(m.feterms)
-    if fe.rank == size(fe, 2)
-        fe.x
-    else
-        fe.x[:, invperm(fe.piv)]
-    end
-end
+StatsBase.modelmatrix(m::LinearMixedModel) = m.feterm.x
```
palday, line 593: I've lost the thread a bit here: but the matrix is no longer stored permuted and is instead dynamically permuted as a view, correct?

src/linearmixedmodel.jl
```diff
@@ -957,29 +963,29 @@ Update the blocked lower Cholesky factor, `m.L`, from `m.A` and `m.reterms` (use
...
+    k = length(reterms)
     for (l, a) in zip(L, A) # copy each element of A to corresponding element of L
         copyto!(l, a)       # For some reason copyto!.(L, A) allocates a lot of memory
     end
-    for (j, cj) in enumerate(m.reterms)  # pre- and post-multiply by Λ, add I to diagonal
+    for j in eachindex(reterms) # pre- and post-multiply by Λ, add I to diagonal
+        cj = reterms[j]
         scaleinflate!(L[kp1choose2(j)], cj)
-        for i = (j+1):k         # postmultiply column by Λ
+        for i = (j+1):(k+1)     # postmultiply column by Λ
```
palday, line 974: `k+1` because `k` is now just the number of `reterms` and we need to add one for the `feterm`, right?

test/matrixterm.jl
```diff
@@ -2,74 +2,70 @@ using LinearAlgebra, MixedModels, StableRNGs, Test, SparseArrays
...
-    @test length(trm) == 0
-    @test size(trm') == (0, 10)
-    @test eltype(trm) == Float64
-    @test trm.rank == 0
+    trm = MixedModels.FeTerm(ones(10,0), String[])
+    #@test size(trm) == (10, 0)
+    #@test length(trm) == 0
+    #@test size(trm') == (0, 10)
+    #@test eltype(trm) == Float64
+    #@test trm.rank == 0
```
palday, line 28: should we re-enable these tests?

test/matrixterm.jl
```diff
@@ -2,74 +2,70 @@ using LinearAlgebra, MixedModels, StableRNGs, Test, SparseArrays
...
-        dense_feval = m.optsum.feval
-        m.feterms[1] = X    
-        refit!(m)
-
-        @test dense_θ ≈ m.θ
+        m1 = fit!(LinearMixedModel(collect(m.y), X, deepcopy(m.reterms), m.formula))
+        @test isapprox(m1.θ, m.θ, rtol = 1.0e-5)
     end
 
     @testset "rank defiency in sparse FeMat" begin
```
palday, line 49:
```suggestion
    @testset "rank defiency in sparse FeTerm" begin
```

(contingent on any name changing)

test/matrixterm.jl
```diff
@@ -2,74 +2,70 @@ using LinearAlgebra, MixedModels, StableRNGs, Test, SparseArrays
...
-        @test prd == [30.0 135.0 270.0; 135.0 855.0 1710.0; 270.0 1710.0 3420][piv, piv]
+        @test prd[ipiv, ipiv] == [30.0 135.0 270.0; 135.0 855.0 1710.0; 270.0 1710.0 3420.0]
         wts = rand(StableRNG(123454321), 30)
-        MixedModels.reweight!(trm, wts)
-        @test mul!(prd, trm', trm)[ipiv[1], ipiv[1]] ≈ sum(abs2, wts)
+        MixedModels.reweight!(mat, wts)
+        @test mul!(prd, mat', mat)[ipiv[1], ipiv[1]] ≈ sum(abs2, wts)
     end
     
 end
```
palday, line 71:
```suggestion
end

```

(the diff highlighted the missing terminal newline)

test/pls.jl
```diff
@@ -31,7 +31,7 @@ include("modelcache.jl")
     @test_logs (:warn, "Model has not been fit") show(fm1)
 
     @test objective(updateL!(setθ!(fm1, [0.713]))) ≈ 327.34216280955366
-    @test_deprecated MixedModels.describeblocks(IOBuffer(), fm1)
+    #@test_deprecated MixedModels.describeblocks(IOBuffer(), fm1)
```
palday, line 34: If you want to remove the `describeblocks` deprecation machinery as well, go for it! This will be the second major release after that deprecation.

src/linalg.jl
palday: I think readability is about the same either way, so I can't be bothered to change it back. :smile:

palday:
The failure on `Future` is something I've fixed as part of my changes in #474

src/Xymat.jl
dmbates: I'm a bit of a purist about the terms `design matrix` and `model matrix`.  To me the term `design matrix` refers to the experimental design, which is to say the levels of the factors in the data table.  The matrix derived from the data table and the formula is the model matrix.  So I would change that to `model matrix`.

palday: Easy! I actually originally wrote "model matrix" then remembered that the last time I wrote that, you wanted "design matrix". But now I understand the distinction you're drawing here.

src/linalg/logdet.jl
```diff
@@ -31,7 +31,9 @@ function LinearAlgebra.logdet(m::LinearMixedModel{T}) where {T}
         s += LD(L[kp1choose2(j)])
     end
     if m.optsum.REML
-        s += LD(L[kp1choose2(nre + 1)])
+        lastL = last(L)
+        s += LD(lastL)  # this adds the log of sqrtpwrss
+        s -= log(last(lastL))
```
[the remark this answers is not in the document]
dmbates: I don't think it should be too bad here because it is the log of the square root of the penalized residual sum of squares.  It shouldn't be especially large or small.  Also, this block is only used for REML estimation and the default is ML.

An alternative would be to define a function like `LD` that iterates over `diagind(size(d, 1), size(d, 2) - 1)` instead of `diagind(d)`.

src/Xymat.jl
dmbates: I have this feeling that it is easier to check the length of a vector than to get the number of columns in a matrix, although it probably involves a few nanoseconds at most.

src/Xymat.jl
dmbates: Okay, makes sense.

src/Xymat.jl
dmbates: Perhaps.  I'm also in the "no strong preference" camp.

palday: Then let's leave it as is because that's less effort.

dmbates:
In `v"1.7.0-DEV.620"` and later, which use libblastrampoline, the first use of `BLAS.vendor()` produces a deprecation warning.  I can commit a change for that if it seems worthwhile.

test/pls.jl
dmbates: Okay, let's nuke `describeblocks` then.  Do you want to do it or shall I?

palday:
Go for it -- we've already got a `@static` guard there noting that I would still like to have tests notify us of platform specifics (since those have been the source of problems in the past).

test/pls.jl
palday: I'll let you do it.

src/linearmixedmodel.jl
dmbates: Yes, I believe you are correct.  Good that you caught that.

test/pls.jl
dmbates: Done in https://github.com/JuliaStats/MixedModels.jl/pull/464/commits/3ee812d63fa74da7a9fe8f3b3150b2794801d4c7

dmbates:
I think we should look at pivotedQR versus pivotedCholesky for determining the rank of X after we merge this.  So probably don't make a new release immediately.

palday:
I also think there's a few things to think about before we put a bow on 4.0.

palday:
There's also something I'm currently tracking down with the new `show` methods. In my local build of the docs, the markdown is used and renders correctly (not what we want for showing users how to step through the REPL, but also at least understandable). But the GHA builds all have the markdown output but not rendered into HTML. I'm working on figuring that out.

src/linalg/logdet.jl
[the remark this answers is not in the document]
palday: I think this is fine for now, but if we ever run into weird trouble on REML fits, it'll be the first place I look.

dmbates:
@palday Let me know if I should be looking at this PR or if you are ready to do the heavy git lifting to make it compatible with current master.

palday:
I already did the lifting. :smile:

dmbates:
I think this is ready to go.  If you agree @palday please go ahead and merge.

palday (approved):

PR #465, opened 2021-02-15
introduce an abstract type for collections of fits

### LINKED ISSUE #325: Abstract type for collection of mixed models
I saw that `MixedModelBootstrap` implements a memory-efficient collection of many mixed models, by saving only the necessary parameters. For unfold.jl it would be nice to be able to have an AbstractType to tune it to my application (which is many MixedModels, but not many Bootstraps).

What I am doing is running many mixed models, for instance one per electrode/voxel (could be thousands of voxels) and one per timepoint (could be thousands of timepoints as well). Saving only the necessary info seems important there.

palday:
Closes #325.

This is a very naive initial proposal. Given the way Julia does typing, I think it might also make sense to discuss the traits I think `MixedModelFitCollection` should have. (Also, I'm willing to add an `Abstract` somewhere in that name.)  I think `MixedModelBootstrap` is actually very close to what we want for the abstract type, but maybe a few changes:

- [x] ~remove `objective` from the inner table and let users compute that on the fly using the stored theta value? (we would need to add an appropriate method)~ This would require keeping the data and/or a lot of PRNG state around.
- [x] rename `bstr` to `fits` and set a minimum table schema that derived types can of course add to. For all the current bootstrap methods that only require that the table, broaden them.
- [x] keep the lambda field
- [x] keep the `lowerbd` field
- [x] `inds` and `fcnames`  might require method specialization to work under a permutation regime.

This change could be made non breaking and introduced into the 3.x series by adding a property accessor/manipulator that returns `fits` when `bstr` is used.

Mostly looking for feedback about whether I missed anything. For the permutation tests over in [MixedModelsPermutations.jl](https://github.com/palday/MixeModelsPermutations.jl), we could of course consider adding in the permutation matrix as a subtype specific struct and then providing property accessors for `inds` and `fcnames` that permute those.

cc: @jaromilfrossard @behinger @dmbates


dmbates:
- I am not clear on the bullet points that you have in the original comment.  The λ, inds, lowerbd and fcnames fields are part of the current struct.  Are you just reiterating here that those should be retained in a more general type.
- I agree with renaming `bstr` to `fits`.  Sometimes I am lazy about naming things.
- I think it will be advisable to keep the `objective`.  As you say, trying to reconstruct it will require considerably more state to be retained and I don't think it would be work the extra cost in storage and complexity.

palday:
>     * I am not clear on the bullet points that you have in the original comment.  The λ, inds, lowerbd and fcnames fields are part of the current struct.  Are you just reiterating here that those should be retained in a more general type.

Yes, exactly.

dmbates (approved):

PR #468, opened 2021-02-17
simulate! marks models as unfitted

### LINKED ISSUE #417: Should we mark a model as unfitted after simulate! ?
Technically the model is no longer fitted to the values contained within.

palday:
Closes #417 .

I've also done some re-arranging of utility functions.


dmbates (approved):
Looks good.  Thanks for splitting out the code from utilities.jl into separate source files.

PR #469, opened 2021-02-17
remove (unused) `named` kwarg from ranef

### LINKED ISSUE #467: Drop kwarg named from ranef

palday:
Closes #467

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -686,14 +686,16 @@ end
 ranef!(v::Vector, m::LinearMixedModel, uscale::Bool) = ranef!(v, m, fixef(m), uscale)
 
 """
-    ranef(m::LinearMixedModel; uscale=false, named=false)
+    ranef(m::LinearMixedModel; uscale=false)
 
 Return, as a `Vector{Matrix{T}}`, the conditional modes of the random effects in model `m`.
 
 If `uscale` is `true` the random effects are on the spherical (i.e. `u`) scale, otherwise on
 the original scale.
+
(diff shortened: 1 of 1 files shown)


src/linearmixedmodel.jl
```diff
@@ -686,14 +686,16 @@ end
...
 
 Return, as a `Vector{Matrix{T}}`, the conditional modes of the random effects in model `m`.
 
 If `uscale` is `true` the random effects are on the spherical (i.e. `u`) scale, otherwise on
 the original scale.
+
+For a named variant, see [`@raneftables`](@ref).
 """
-function ranef(m::LinearMixedModel{T}; uscale = false, named = false) where {T}
+function ranef(m::LinearMixedModel{T}; uscale = false) where {T}
```
ararslan, line 698:
```suggestion
function ranef(m::LinearMixedModel{T}; uscale = false, named = nothing) where {T}
    if named !== nothing
        Base.depwarn("the `named` keyword argument is deprecated; it has no effect", :ranef)
    end
```
This avoids making a breaking change, since user code currently calling `ranef(...; named=true)` will error if the argument is removed.

palday: This is queued for 4.0, which has other breaking changes.... but this would provide increased backwards compatibility (as most of the breaking changes are in the internal representation that the average user won't touch)

ararslan: Yeah if it's queued for a breaking release, might as well break it :smile:

ararslan (approved):

PR #470, opened 2021-02-20
Runtime random effect terms (and zerocorr)

### LINKED ISSUE #462: Programmatically Creating RE Terms
This is at least partly a StatsModels.jl issue, partly a mixedmodels issue, and partly a julia/macro issue.

I think what you're trying to do is construct a formula from a list of variable names.  For more information on that, see the docs for statsmodels on constructing a formula at run-time: https://juliastats.org/StatsModels.jl/stable/formula/#Constructing-a-formula-programmatically-1.  You can do the fixed effects part easily:

```julia
julia> fix = term(1) + sum(term.("x$n" for n in 1:3))
1
x1(unknown)
x2(unknown)
x3(unknown)
```

The mixedmodels issue is that constructing a random-effects term at run-time using that technique is tricky, but it can be done.  The reason is that you have to construct the special `RandomEffectsTerm` used by mixedmodels, but that's assumed to happen after the schema has already been extracted (see this section of the statsmodels docs: https://juliastats.org/StatsModels.jl/stable/internals/#The-lifecycle-of-a-@formula-1, and this code from mixedmodels on how the random effects syntax with `|` is implemented: https://github.com/JuliaStats/MixedModels.jl/blob/master/src/randomeffectsterm.jl; note especially how the constructor for `RandomEffectsTerm` checks whether the grouping is an interaction or categorical term, which is only known after the schema is applied).  This is to my mind a legitimate quality of life issue.  At the moment you'll have to construct the concrete versions of all the terms and assemble the RandomEffectsTerm manually.  While this technically works (in that it will give you a formula you can fit with mixedmodels) it isn't going to give you any of the bumpers that mixedmodels normally provides (e.g., if you create multiple random effects terms this way, then you might end up with ... "interesting" (mis-specified) models), so it's definitely a "here be dragons" kinda thing that we should fix.  So with the caveat that I'm handing you a big foot-gun here, this is how I'd do it:

```julia
julia> ran = sum(term.("z$n" for n in 1:2))
z1(unknown)
z2(unknown)

julia> ran = apply_schema(sum(term.("z$n" for n in 1:2)), schema(dat), MixedModel)
z1(continuous)
z2(continuous)

julia> grouping = StatsModels.concrete_term(term(:g), dat, Dict(:g => Grouping()))
g(Grouping:5→5)

julia> reterm = MixedModels.RandomEffectsTerm(ran, grouping)
(z1 + z2 | g)

julia> formula = term(:y) ~ term(1) + fix + reterm
FormulaTerm
Response:
  y(unknown)
Predictors:
  1
  x1(unknown)
  x2(unknown)
  x3(unknown)
  (z1 + z2 | g)
```

This gets us most of the way there but it still doesn't work unfortunately.  The issue is that MixedModels tries to extract a schema and doesn't recognize that `RandomEffectsTerms` don't need them...we can duck patch that by defining `StatsModels.needs_schema(::RandomEffectsTerm) = false` but that's better fixed here or in statsmodels IMO.

In the interest of fixing this the right way, here are the missing pieces in MixedModels at the moment

- [ ] make `apply_schema(::FunctionTerm{typeof(|)}, ...)` create an "non-concrete" `RandomEffectsTerm`
- [ ] add additional `apply_schema(::RandomEffectsTerm, ...)` which makes it concrete, does the validity checks done in teh inner constructor right now
- [ ] add run-time methods for `Base.|(::AbstractTerm, ::AbstractTerm)` to create a `RandomEffectsTerm`, or if we don't want to do type piracy, use a different syntax (like `re` or `reterm` or something...)

Finally, the most superficial reason why your attempt doesn't work is a Julia/macro issue: `@formula(f)` only sees one thing as its argument, the Symbol `:f`.  Macros in julia can only do transformations of the surface syntax (symbols, exprs, strings, and literal numbers).  Reading the docs on macros in julia will be helpful, and people on slack/discourse/zulip can answer other questions in that direction.  If you're going to go this way (which is not recommended), then you need to use `eval` and `$f`.

_Originally posted by @kleinschmidt in https://github.com/JuliaStats/MixedModels.jl/issues/426#issuecomment-720564742_

kleinschmidt:
This fixes #462 and more or less follows the plan outlined there: it adds a
method for `Base.:|` for two terms which returns a `RandomEffectsTerm`, and adds
an additional layer of indirection for `apply_schema` to operate on the
"untyped" `RandomEffectsTerm`s created by that process.  I've also made the
analogous changes to `zerocorr` (although that runtime method was actually
already in place, it didn't, you know, work...)

To my mind, the one big wart here is that we're "claiming"
`Base.:|(::TermOrTerms, ::TermOrTerms)` which wouldn't play nicely with another
package that also wants to provide a different set of semantics for `|` (at
run-time).  I first thought that we could do the same trick that
FilePathsBase.jl does with `/` (creating a new `/` function that falls back to
`Base./` for arguments of types it doesn't own) but that didn't seem to work (I
got an "ignoring conflicting import" when doing `using MixedModels: |` with that
strategy).

### DIFF
--- a/docs/src/constructors.md
+++ b/docs/src/constructors.md
@@ -27,12 +27,33 @@ dyestuff = DataFrame(MixedModels.dataset(:dyestuff))
 describe(dyestuff)
 ```
 
-### Models with simple, scalar random effects
+### The `@formula` language in Julia
 
-The formula language in *Julia* is similar to that in *R* which is based on (Wilkinson and Rogers [1973](https://dx.doi.org/10.2307/2346786)). In Julia a formula must be enclosed in a call to the `@formula` macro.
+MixedModels.jl builds on the the *Julia* formula language provided by [StatsModels.jl](https://juliastats.org/StatsModels.jl/stable/formula/), which is similar to the formula language in *R* and is also based on (Wilkinson and Rogers [1973](https://dx.doi.org/10.2307/2346786)). There are two ways to construct a formula in *Julia*.  The first way is to enclose the formula expression in the `@formula` macro:
 ```@docs
 @formula
 ```
+
+The second way is to combine `Term`s with operators like `+`, `&`, `~`, and others at "run time".  This is especially useful if you wish to create a formula from a list a variable names.  For instance, the following are equivalent:
+
+```@example Main
+@formula(y ~ 1 + a + b + a & b) == term(:y) ~ term(1) + term(:a) + term(:b) + term(:a) & term(:b)
+```
+
+MixedModels.jl provides additional formula syntax for representing *random-effects terms*.  Most importantly, `|` separates random effects and their grouping factors (as in the formula extension used by the *R* package [`lme4`](https://cran.r-project.org/web/packages/lme4/index.html).  Much like with the base formula language, `|` can be used within the `@formula` macro and to construct a formula programmatically:
+
+```@example Main
+@formula(y ~ 1 + a + b + (1 + a + b | g))
+```
+
+```@example Main
+terms = sum(term(t) for t in [1, :a, :b])
+group = term(:g)
+response = term(:y)
+response ~ terms + (terms | group)
+```
+
+### Models with simple, scalar random effects
 
 A basic model with simple, scalar random effects for the levels of `batch` (the batch of an intermediate product, in this case) is declared and fit as
 
--- a/docs/src/constructors.md
+++ b/docs/src/constructors.md
@@ -29,7 +29,7 @@ describe(dyestuff)
 
 ### The `@formula` language in Julia
 
-MixedModels.jl builds on the the *Julia* formula language provided by [StatsModels.jl](https://juliastats.org/StatsModels.jl/stable/formula/), which is similar to the formula language in *R* and is also based on (Wilkinson and Rogers [1973](https://dx.doi.org/10.2307/2346786)). There are two ways to construct a formula in *Julia*.  The first way is to enclose the formula expression in the `@formula` macro:
+MixedModels.jl builds on the the *Julia* formula language provided by [StatsModels.jl](https://juliastats.org/StatsModels.jl/stable/formula/), which is similar to the formula language in *R* and is also based on the notation from Wilkinson and Rogers ([1973](https://dx.doi.org/10.2307/2346786)). There are two ways to construct a formula in Julia.  The first way is to enclose the formula expression in the `@formula` macro:
 ```@docs
 @formula
 ```
@@ -556,4 +556,3 @@ fm4r = fit(MixedModel, @formula(diameter ~ 1+(1|plate)+(1|sample)),
 ```@example Main
 sum(leverage(fm4r))
 ```
-
--- a/src/randomeffectsterm.jl
+++ b/src/randomeffectsterm.jl
@@ -36,6 +36,12 @@ function StatsModels.apply_schema(
     apply_schema(RandomEffectsTerm(lhs, rhs), schema, Mod)
 end
 
+# allowed types (or tuple thereof) for blocking variables (RHS of |):
+const GROUPING_TYPE = Union{<:CategoricalTerm, <:InteractionTerm{<:NTuple{N,CategoricalTerm} where {N}}}
+check_re_group_type(term::GROUPING_TYPE) = true
+check_re_group_type(terms::Tuple{Vararg{<:GROUPING_TYPE}}) = true
+check_re_group_type(x) = false
+
 # make a potentially untyped RandomEffectsTerm concrete
 function StatsModels.apply_schema(
     t::RandomEffectsTerm,
@@ -60,8 +66,7 @@ function StatsModels.apply_schema(
     lhs, rhs = apply_schema.((lhs, rhs), Ref(schema), Mod)
 
     # check whether grouping terms are categorical or interaction of categorical
-    rhs isa CategoricalTerm ||
-        rhs isa InteractionTerm{<:NTuple{N,CategoricalTerm} where {N}} ||
+    check_re_group_type(rhs) ||
         throw(ArgumentError("blocking variables (those behind |) must be Categorical ($(rhs) is not)"))
     
     RandomEffectsTerm(MatrixTerm(lhs), rhs)
--- a/src/randomeffectsterm.jl
+++ b/src/randomeffectsterm.jl
@@ -22,6 +22,8 @@ function StatsModels.termvars(t::RandomEffectsTerm)
     vcat(StatsModels.termvars(t.lhs), StatsModels.termvars(t.rhs))
 end
 
+StatsModels.terms(t::RandomEffectsTerm) = union(StatsModels.terms(t.lhs), StatsModels.terms(t.rhs))
+
 # | in MixedModel formula -> RandomEffectsTerm
 function StatsModels.apply_schema(
     t::FunctionTerm{typeof(|)},
--- a/src/randomeffectsterm.jl
+++ b/src/randomeffectsterm.jl
@@ -194,12 +194,24 @@ Remove correlations between random effects in `term`.
 """
 zerocorr(x) = ZeroCorr(x)
 
+# for schema extraction (from runtime-created zerocorr)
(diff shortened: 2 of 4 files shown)


kleinschmidt:
Also added docs for run-time construction (not sure it worked?)

palday:
> Also added docs for run-time construction (not sure it worked?)

It did: https://juliastats.org/MixedModels.jl/previews/PR470/constructors/#The-@formula-language-in-Julia

docs/src/constructors.md
```diff
@@ -27,13 +27,34 @@ dyestuff = DataFrame(MixedModels.dataset(:dyestuff))
 describe(dyestuff)
 ```
 
-### Models with simple, scalar random effects
+### The `@formula` language in Julia
 
-The formula language in *Julia* is similar to that in *R* which is based on (Wilkinson and Rogers [1973](https://dx.doi.org/10.2307/2346786)). In Julia a formula must be enclosed in a call to the `@formula` macro.
+MixedModels.jl builds on the the *Julia* formula language provided by [StatsModels.jl](https://juliastats.org/StatsModels.jl/stable/formula/), which is similar to the formula language in *R* and is also based on (Wilkinson and Rogers [1973](https://dx.doi.org/10.2307/2346786)). There are two ways to construct a formula in *Julia*.  The first way is to enclose the formula expression in the `@formula` macro:
```
palday, line 32:
```suggestion
MixedModels.jl builds on the the *Julia* formula language provided by [StatsModels.jl](https://juliastats.org/StatsModels.jl/stable/formula/), which is similar to the formula language in *R* and is also based on the notation from Wilkinson and Rogers ([1973](https://dx.doi.org/10.2307/2346786)). There are two ways to construct a formula in Julia.  The first way is to enclose the formula expression in the `@formula` macro:
```

docs/src/constructors.md
```diff
@@ -27,13 +27,34 @@ dyestuff = DataFrame(MixedModels.dataset(:dyestuff))
...
+
+```@example Main
+@formula(y ~ 1 + a + b + (1 + a + b | g))
+```
+
+```@example Main
+terms = sum(term(t) for t in [1, :a, :b])
+group = term(:g)
+response = term(:y)
+response ~ terms + (terms | group)
```
palday, line 53: This is the coolest -- you can guarantee that your RE includes the same things as you're FE.

Just don't use this for ~evil~ maximal models.

src/randomeffectsterm.jl
```diff
@@ -49,15 +22,40 @@ function StatsModels.termvars(t::RandomEffectsTerm)
...
+        throw(ArgumentError("Same variable appears on both sides of |"))
+
+    apply_schema(RandomEffectsTerm(lhs, rhs), schema, Mod)
+end
+
+# allowed types (or tuple thereof) for blocking variables (RHS of |):
+const GROUPING_TYPE = Union{<:CategoricalTerm, <:InteractionTerm{<:NTuple{N,CategoricalTerm} where {N}}}
+check_re_group_type(term::GROUPING_TYPE) = true
+check_re_group_type(terms::Tuple{Vararg{<:GROUPING_TYPE}}) = true
+check_re_group_type(x) = false
```
palday, line 45: should we explicitly mark these as private/internal with `_`?

src/randomeffectsterm.jl
```diff
@@ -3,43 +3,16 @@ abstract type AbstractReTerm <: AbstractTerm end
...
+# imported with (a la FilePathsBase.:/)
+# using MixedModels: |
+# to avoid conflicts with definitions in other packages...
+Base.:|(a::StatsModels.TermOrTerms, b::StatsModels.TermOrTerms) = RandomEffectsTerm(a, b)
 
-    return first + fulldummy(first) & second
-end
-
-RandomEffectsTerm(lhs, rhs::NTuple{2,AbstractTerm}) =
+# expand (lhs | a + b) to (lhs | a) + (lhs | b)
```
palday, line 14: what happens with the nesting `/` syntax when constructed programmatically?

palday: (I mostly mean in the grouping/block vars, i.e. RHS of `|`, but would also be curious about LHS)

src/randomeffectsterm.jl
```diff
@@ -169,6 +194,10 @@ Remove correlations between random effects in `term`.
 """
 zerocorr(x) = ZeroCorr(x)
 
+# for schema extraction (from runtime-created zerocorr)
+StatsModels.terms(t::ZeroCorr) = StatsModels.terms(t.term)
+StatsModels.termvars(t::ZeroCorr) = StatsModels.termvars(t.term)
```
palday, line 199: I kept meaning to add in these methods, thanks!

test/FactorReTerm.jl
```diff
@@ -99,6 +99,55 @@ end
...
+        
+        # sleep study data:
+        r, d, s, one = term.((:reaction, :days, :subj, 1))
+
+        f1 = @formula(reaction ~ 1 + (1 + days | subj))
+        f2 = r ~ one + (one + d | s)
+        @test f2.rhs[end] isa RandomEffectsTerm
+        ff1 = apply_schema(f1, schema(slp), LMM)
+        ff2 = apply_schema(f2, schema(slp), LMM)
+        # equality of RE terms not defined so check that they generate same modelcols
```
palday, line 117: should we define equality? I have no idea if it's useful beyond tests....

palday (approved):
LGTM, mostly small stuff, so gonna go ahead and approve. If you change something with `/`, request another review please.

docs/src/constructors.md
kleinschmidt: you know people are gonna use this for maximal models

src/randomeffectsterm.jl
kleinschmidt: Nothing right now (it won't work AFAIK) since we don't define methods for `Base.:/(a::AbstractTerm, b::AbstractTerm)`

kleinschmidt: but I guess that's now the odd-one-out so maybe worth it?

palday: If you can do it in a few lines, please. :smile: We have a certain user who's a big fan. :wink:

kleinschmidt: actually this is a bit trickier than I thought at first...the run-time stuff has to be able to work before we have an applied schema, so we'd have to either be able to lazily convert to full-dummy, or lazily expand the `/`.

OR, actually, maybe it would be okay since the full dummy thing only applies to fixed effects anyway and this is exactly what the `FullRank` schema is intended to handle (since if you have `1 + a + a & b`, there's no `b`, so `a` should get promoted).

The other possibly slightly tricky thing would be deciding whether to apply the "distributive rule" for `|` at run-time or leave that until schema time; but I think that's handled by a constructor right now so should also be okay.

palday: I'll leave it up to you. If you think it's not worthwhile, I would add one line to the docs and move on.

test/FactorReTerm.jl
kleinschmidt: equality of terms generally is kind of a mess so I'm happy to leave it as is for now.

src/randomeffectsterm.jl
kleinschmidt: they're not exported so I dunno, up to you I guess.

kleinschmidt:
Okay I've also added a method for `Base.:/` for run-time nesting.  I don't think that should live here and it's technically type piracy but I made a note of it and we can see if it causes anyone problems...

The one slightly tricky thing is that we can't check that the nesting variable is categorical at run-time (since in general the schema has not been applied yet).  I don't think that's SUCH a big deal but it is a bit of a gotcha I guess.

palday (approved):

src/randomeffectsterm.jl
palday: Hmmm.... Not exported, not advertised and lacking a docstring. People should get the hint. :smile:

palday:
Oh no the bots aren't happy

palday:
Can you add a `NEWS.md` entry? I nearly forgot

PR #472, opened 2021-02-24
use fname instead of term.sym directly

### LINKED ISSUE #471: isnested assumes that grouping terms have single blocking vars
Checking `isnested` for two models where at least one of them has an interaction
blocking variable (e.g., `(1 | batch & cask)`) fails because the interaction
term does not have a `.sym` field.  Should use the `fname` function for this
instead (I'll file a fix).

kleinschmidt:
Fixes #471


kleinschmidt:
might make sense to backport this to a 3.x branch (if that's a thing)

palday:
As soon as the bots are happy, I'm happy

palday (approved):

PR #479, opened 2021-03-02
QR-Based Rank Determination and Pivoting

dmbates:
- add `statsrank` to compute the rank and pivot vector for a model matrix using a pivoted QR decomposition
- modify tests, etc.
- remove `statsqr` and `statscholesky` functions

### DIFF
--- a/src/linalg/pivot.jl
+++ b/src/linalg/pivot.jl
@@ -1,35 +1,36 @@
 """
-    statsqr(x::Matrix{T}, ranktol::Real=1e-8) where {T<:AbstractFloat}
+    statsrank(x::Matrix{T}, ranktol::Real=1e-8) where {T<:AbstractFloat}
 
-Return the numerical column rank and, for a rank-deficient X, a pivot vector
+Return the numerical column rank and a pivot vector.
+
+The rank is determined from the absolute values of the diagonal of R from
+a pivoted QR decomposition, relative to the first (and, hence, largest) 
+element of this vector.
+
+In the full-rank case the pivot vector is `collect(axes(x, 2))`.
 """
-function statsqr(x::Matrix{T}; ranktol=1e-8) where {T<:AbstractFloat}
-    n = size(x, 2)
+function statsrank(x::AbstractMatrix{T}; ranktol=1e-8) where {T<:AbstractFloat}
+    m, n = size(x)
 
     qrpiv = qr(x, Val(true))
-    rank =  searchsortedlast(abs.(diag(qrpiv.R)), ranktol, rev=true);
-    if rank < n
-        piv = qrpiv.p
-    else
-        piv = LinearAlgebra.BlasInt.(collect(1:n))
+    dvec = abs.(diag(qrpiv.R))
+    fdv = first(dvec)
+    if (last(dvec) / fdv) > ranktol
+        return (rank = n, piv = collect(axes(x, 2)))
     end
-
-    if rank < n && piv[1] ≠ 1
-        # make sure the first column isn't moved
-        # this is usually the intercept and it's desirable to avoid
-        # pivoting the intercept out. note that this inflation may
-        # not always be enough, but if it isn't, then you probably
-        # have other problems
-
-        x = deepcopy(x) # this is horrible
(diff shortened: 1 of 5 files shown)


palday:
@dmbates While I'm reviewing, can you go ahead and write a NEWS entry for this?

src/linalg/pivot.jl
```diff
@@ -1,82 +1,36 @@
...
-        # make sure the first column isn't moved
-        # this is usually the intercept and it's desirable to avoid
-        # pivoting the intercept out. note that this inflation may
-        # not always be enough, but if it isn't, then you probably
-        # have other problems
-
-        x = deepcopy(x) # this is horrible
-        x[:, 1] .*= 1e6
+    rank = searchsortedlast(dvec ./ fdv, ranktol, rev=true)
+    @assert rank < n
```
palday, line 22: I like having this defense here (as a reminder to us as well), but note that `@assert` may disappear at some optimization levels.

src/linalg/pivot.jl
```diff
@@ -1,82 +1,36 @@
...
+    piv = qrpiv.p
+    v1 = first(eachcol(x))
+    if all(isone, v1) && first(piv) ≠ 1
+        # make sure the first column isn't moved by inflating v1
+        v1 .*= (fdv + true) / sqrt(m)
         qrpiv = qr(x, Val(true))
         piv = qrpiv.p
+        fill!(v1, true)    # restore the contents of the first column
```
palday, line 30:
```suggestion
        fill!(v1, one(T))    # restore the contents of the first column
```

src/linalg/pivot.jl
```diff
@@ -1,82 +1,36 @@
...
-        # not always be enough, but if it isn't, then you probably
-        # have other problems
-
-        x = deepcopy(x) # this is horrible
-        x[:, 1] .*= 1e6
+    rank = searchsortedlast(dvec ./ fdv, ranktol, rev=true)
+    @assert rank < n
+    piv = qrpiv.p
+    v1 = first(eachcol(x))
+    if all(isone, v1) && first(piv) ≠ 1
```
palday, line 25: I like that this only prevents moving the first column if it looks like an intercept

src/linalg/pivot.jl
```diff
@@ -1,82 +1,36 @@
...
         piv = qrpiv.p
+        fill!(v1, true)    # restore the contents of the first column
     end
 
     # preserve as much of the original column ordering as possible
-    piv = [sort!(piv[1:rank]); sort!(piv[rank+1:n])];
-    qrunp = qr(x[:, piv], Val(false));
-
-    QRPivoted(qrunp.factors, qrpiv.τ, piv)
+    sort!(view(piv, 1:rank))
```
palday, line 34:
```suggestion
    sort!(view(piv, 1:rank)) # maintain the original column order for the linearly independent columns
```

It took me a minute to realize what was going on here and I always assume Future Me is worse off than Present Me when it comes to understanding clever code. :smile:

palday (approved):
I think this looks much cleaner. Should I try an MKL test run on the CIs?

palday:
I've kicked off an [MKL run](https://github.com/JuliaStats/MixedModels.jl/runs/2017195243)

dmbates:
> I've kicked off an [MKL run](https://github.com/JuliaStats/MixedModels.jl/runs/2017195243)

I think it would be better to only do the MKL run on the Future  platform.  As you can see, it takes a long time to run on 1.5

src/linalg/pivot.jl
dmbates: Good suggestion.  I had forgotten that `T` is defined in this method

palday:
@dmbates The MKL run is now done on [nightly](https://github.com/JuliaStats/MixedModels.jl/runs/2017558628). 

When 1.7 is released, we can revisit making MKL part of the standard test suite instead of purely optional like it is now.

dmbates:
> @dmbates While I'm reviewing, can you go ahead and write a NEWS entry for this?

Done.

palday (approved):

PR #480, opened 2021-03-03
Show methods for HTML and LaTeX

palday:
Closes #273

PR #481, opened 2021-03-04
BACKPORT: LaTeX and HTML show

palday:
This brings the methods from #480 back to the v3.x series.


dmbates:
Good to merge I say.

PR #482, opened 2021-03-04
Allow for offset in GLMMs and simplify wts

### LINKED ISSUE #453: Offset for Poisson models
Hi,

I might be missing something, but there doesn't seem to be any way to provide per-observation constant offsets for the `GeneralizedLinearMixedModel`?

The `fit` function supports an argument `offset`, but it doesn't appear to do anything. (Fitting with and without setting this gives the same result).

In the `GLM.jl` package the `fit` function for `GeneralizedLinearModel` has an `offset` argument that works.

How can I add an offset to a Poisson GLMM in MixedModels.jl?

Best,
/Valentine

dmbates:
- closes #453 
- working on an example based on one in lme4


dmbates:
@palday Should we go ahead and merge this, perhaps after turning off the Future ci.  I would say that this is a bug fix.  (I suppose now you want me to put it in News.md, don't you?)

palday:
@dmbates you predicted my request. :smile: Can you put it under `3.4.1` since it's technically a bug fix?

As a sidebar: should we support offsets in LMM (in a separate PR)?

dmbates:
This branch is currently based on the `master` branch because I was intending it to check off one of the milestones for v4.0.0,  I know it is a bug fix and could be incorporated as v3.4.1 but I am not sure my git skills are up to backing out the other changes in the master branch.

palday:
Put the NEWS entry in the right spot, we'll merge into master and I'll git-magic the backport aftwards. :smile:

dmbates:
> As a sidebar: should we support offsets in LMM (in a separate PR)?

The uses of an offset for a model with identity link have always seemed rather contrived to me.  You might as well just shift the response by the offset and proceed normally.

palday:
> The uses of an offset for a model with identity link have always seemed rather contrived to me. You might as well just shift the response by the offset and proceed normally.

My thoughts exactly. :smile: Should we error if an offset is specified for an LMM then?

dmbates:
> > The uses of an offset for a model with identity link have always seemed rather contrived to me. You might as well just shift the response by the offset and proceed normally.
> 
> My thoughts exactly. smile Should we error if an offset is specified for an LMM then?

Yes, please do.

palday:
The constructor clearly will error, but for the `fit(MixedModel, formula, data; offset=data.o)` I don't know what will happen.

I'll add it later today and then get this merged in. :smile:

palday (approved):

PR #484, opened 2021-03-08
shortestcovint for MixedModelBootstrap objects

palday:
Doing this is as a 3.x release, but I'll of course port this into the development version. I think it's useful enough for an immediate release.

This also contains a few other fixes:
- the doc fix from #483 
- the TagBot release notes now strongly suggest that users look at NEWS rather than the closed issues/PRs because those can include changes to other parallel development streams.


dmbates (approved):
Looks good,  Thanks for thinking of doing this.  It helps to have fewer situations where we require Dataframes.

PR #486, opened 2021-03-11
bring back LazyArtifacts

palday:
This takes advantage of the compatibility package `LazyArtifacts` available in the general registry.

For users who never use the test datasets, this will speed up install time and avoid a few MB of downloads.


palday:
We seem to be triggering a failure on Julia nightly in the `ReMat eltype` test -- a pretty big stack overflow somehow.

src/MixedModels.jl
```diff
@@ -4,11 +4,11 @@ using Arrow
 using DataAPI
 using Distributions
 using GLM
+using LazyArtifacts
 using LinearAlgebra
 using Markdown
 using NLopt
 using Random
-using Pkg.Artifacts
```
ararslan, line 11: Can you drop the dependency on Pkg from Project.toml with this change?

palday: I think so, let's see if the bots agree. I also added a `compat` entry for LazyArtifacts

ararslan (approved):
I assume that using LazyArtifacts in this way is compatible between having it be or not be a stdlib. I guess we'll find out with nightly CI if the package can load test data properly across all versions.

palday:
I've been testing on 1.6rc1 locally and it works fine. I think that's kinda the point of `LazyArtifacts` as a registry package. :smile:

PR #489, opened 2021-03-18
Diagonalize ZeroCorr

### LINKED ISSUE #487: Be Diagonal where you can
We should allow the λ field in an `ReMat` in the `MixedModels` package to be `LowerTriangular{T,Union{Diagonal{T},Matrix{T}}}` and check if the `inds` make the result diagonal.  For example, `zerocorr` applied to an `ReTerm` results in a diagonal λ.  A surprisingly large amount of the time in `updateL!` is spent in the mulitplications by Λ, especially for a `UniformBlockDiagonal` (1, 1) block.

_Originally posted by @dmbates in https://github.com/RePsychLing/MixedModelsSim.jl/pull/42#issuecomment-801161864_

palday:
Closes #487 .

I had to loosen some type constraints in various places to make this work and I didn't add in any additional `mul!`  methods, so I don't know if we're fully taking advantages of this. @dmbates Feel free to add those in if you want, but those wouldn't be breaking changes.

I also changed `rownormalize!` to be non mutating. This made specializing for `Diagonal` easier and actually cut out an extra allocation.

### DIFF
--- a/.github/workflows/Future.yml
+++ b/.github/workflows/Future.yml
@@ -1,12 +1,13 @@
 name: Future
 on:
-  pull_request:
-    branches:
-      - master
-      - v3.x
-    paths-ignore:
-      - 'LICENSE.md'
-      - 'README.md'
+  # pull_request:
+  #   branches:
+  #     - master
+  #     - v3.x
+  #   paths-ignore:
+  #     - 'LICENSE.md'
+  #     - 'README.md'
+  workflow_dispatch:
 jobs:
   ci:
     runs-on: ${{ matrix.os }}
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,9 +14,8 @@ MixedModels v4.0.0 Release Notes
   rank of `X` by `statsrank`. [#479]
 * Artifacts are now loaded lazily: the test data loaded via `dataset` is
   downloaded on first use [#486]
-* `ReMat` and `PCA` now support covariance factors (`λ`) originating from
-  `<:AbstractMatrix`. Building on this, the covariance factor for random
-  effects without correlation parameters are now stored as `LowerTriangular{T, Diagonal{T, Vector{}}}`. This representation is both more memory efficient and
+* `ReMat` and `PCA` now support covariance factors (`λ`) that are `LowerTriangular`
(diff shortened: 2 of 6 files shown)


dmbates:
Thanks.  Taking a look now.

dmbates:
It is not great that I am just remembering this but I think that taking advantage of this change in the multiplications will either involve type piracy or a PR on the `LinearAlgebra` package.  The reason is that methods like
```
function lmulΛ!(adjA::Adjoint{T,ReMat{T,S}}, B::VecOrMat{T}) where {T,S}
    lmul!(adjoint(adjA.parent.λ), reshape(B, S, :))
    B
end
```
get to the point of calling `lmul!` before the type of λ is known.   So we would have to override methods of `lmul!`.

In retrospect I think the best approach is to define the type of λ to be `Union{Diagonal{T,Vector{T}}, LowerTriangular{T,Matrix{T}}}`.  I make those modifications if you agree @palday

palday:
> In retrospect I think the best approach is to define the type of λ to be `Union{Diagonal{T,Vector{T}}, LowerTriangular{T,Matrix{T}}}`. I make those modifications if you agree @palday

If we ever implement other covariance structures, are there special matrix forms that occur there? In that case, I would make  λ to be `<:AbstractMatrix{T}` and then remove the `LowerTriangular` wrapper around the `Diagonal` things I added (so that we could add in specialization for those structures without needing additional change to the `ReMat` type). If not, then your approach is simpler.

Go with whatever you think is best!

dmbates:
@palday For the time being I think I will stay with `Union{LowerTriangular{T,Matrix{T}},Diagonal{T,Vector{T}}}` for the type.  I believe that the compiler can handle small unions better than more general supertypes.

How would I go about turning off the `Future.yml` spec for the CI runs?  Right now it needs to be stopped manually.

palday:
I've disabled Future and I think you're right on the compiler stuff.

palday (commented):
It's my own pull request, so I can't approve it, but it looks good to me!

dmbates (approved):

dmbates:
@palday Would you take another look at this and merge if it looks okay to you?

PR #492, opened 2021-03-23
Add sparseL extractor and use it in condVar

dmbates:
- `sparseL` extracts the `L` matrix as a lower-triangular `SparseMatrixCSC`
- rewrite `condVar` to solve the lower-triangular system using `L` and a block of columns from `sΛ'`,  The X'X product of this solution is a block of the `condVar` array.
- needs tests but currently fails b/c the lower-triangular solution in `SparseArrays` does not handle integer types other than `Int` correctly.  It's a trivial fix but still needs to be done.


src/linearmixedmodel.jl
```diff
@@ -285,15 +285,28 @@ diagonal blocks from the conditional variance-covariance matrix,
...
 """
 function condVar(m::LinearMixedModel{T}) where {T}
-    retrms = m.reterms
-    t1 = first(retrms)
-    L11 = first(m.L)
-    if !isone(length(retrms)) || !isa(L11, Diagonal{T,Vector{T}})
-        throw(ArgumentError("code for multiple or vector-valued r.e. not yet written"))
+    L = m.L
+    s = sdest(m)
+    spL = LowerTriangular(sparseL(m))
```
palday, line 290:
```suggestion
    # when https://github.com/JuliaLang/julia/pull/40174 lands,
    # we can update the version bound
    @static if VERSION < v"2"
        spL = LowerTriangular(SparseMatrixCSC{T, Int}(sparseL(m)))
    else 
        spL = LowerTriangular(sparseL(m))
    end
```

It adds an extra step, but at a point where it should be relatively cheap, and the type change across versions isn't exposed to the user / doesn't cause type instability in the final result.

dmbates: Do we have a version number to fill in here?  Yesterday it was `v"1.7.0-DEV.761"` which is only three commits after your modification.

palday: I think we can use that number. The change is also marked to be backported to 1.6.1 when it comes out.

dmbates:
For testing on v"1.7.0-DEV" I tried to wrap any fitting of GLMMs in conditionals on VERSION but it turned into a game of whack-a-mole.  I did check that the tests on `condVar` in `pls.jl` succeeded.

dmbates:
I decided to play whack-a-mole and found a case which went into the infinite inference loop without trying to fit a GLMM.
```julia
julia> using MixedModels, Random

julia> fm = fit(MixedModel, @formula(yield ~ 1 + (1|batch)), MixedModels.dataset(:dyestuff));

julia> bsamp = parametricbootstrap(MersenneTwister(1234321), 100, fm);
```
At this point `bsamp.objective` causes the infinite loop for me.  Could others check it out under v"1.7.0-DEV.760" or later?

palday:
If `typeof(bsamp.fits)` is causing an infinite loop, I don't think there's much to do except keep bumping the [relevant Julia issue](https://github.com/JuliaLang/julia/issues/40048). If the `pls.jl` tests pass locally for you on 1.7-DEV, I'm fine with that and we can revisit the problem when the upstream bug is fixed.

palday:
Should we add something analogous to `raneftables` for `condVar`?

kliegl:
Yes, please.

dmbates:
Okay, I will add a function like `raneftables` for `condVar` later today.

dmbates:
Another enhancement I was thinking of is to specify which grouping factors you want the `condVar` array for.  The point is that it is much easier to evaluate them for the grouping factors with fewer random effects associated with each level.  I think @kliegl mentioned that he had hundreds of thousands of students but only hundreds of schools and doesn't need the conditional variances for the students.  They will take much longer to evaluate that will the conditional variances for the schools and districts.

dmbates:
In a function like `raneftables` for the conditional variances, how do you want to handle the covariance matrices?  Should it just return a table of covariance matrices or should it split them into standard deviations and correlations?  (Why do I think the answer is going to be, "why not both?"?)

palday:
> In a function like `raneftables` for the conditional variances, how do you want to handle the covariance matrices? Should it just return a table of covariance matrices or should it split them into standard deviations and correlations? 

Standard deviations and correlations would be my preference -- that's also what people are going to care about for making dot plots. Power users who need the covariance matrices should be able to deal with the less user friendly `condVar` directly. 

> (Why do I think the answer is going to be, "why not both?"?)

Type stability and simplicity of interface would be my argument for only doing one. :smile:

kliegl:
I agree with Phillip.

dmbates:
I was going to add more to this PR but I think it is best to get it into main now and worry about non-breaking enhancements later.

dmbates:
@palday Could you check on the conflict in `test/mime.jl`?

src/utilities.jl
```diff
@@ -25,7 +25,7 @@ end
 
 Return the average of `a` and `b`
 """
-average(a::T, b::T) where {T<:AbstractFloat} = (a + b) / 2
+average(a::T, b::T) where {T<:AbstractFloat} = (a + b) / T(2)
```
palday, line 28: Did you run into trouble with this somewhere?  As an integer `2` should be promoted to whatever floating point type `(a+b)` is.

test/bootstrap.jl
```diff
@@ -34,7 +35,7 @@ include("modelcache.jl")
         # restore the original state
         refit!(fm, vec(float.(ds.yield)))
     end
-
+@static if VERSION < v"1.7.0-DEV.700"
```
palday, line 38: can we remove these guards and/or open an issue reminding us to do so when the type inference issue is solved in Julia?

test/pls.jl
```diff
@@ -217,6 +229,10 @@ end
     @test fm1.optsum.initial == ones(3)
     @test lowerbd(fm1) == zeros(3)
 
+    spL = sparseL(fm1)
+    @test size(spL) == (4114, 4114)
+    @test 733090 < nnz(spL) < 733100
```
palday, line 234: are the bounds here just because of floating point and whether some values are rounded off to zero?

src/utilities.jl
dmbates: I thought there would be an issue is with `Float32` in that the literal `2` is an `Int` which will get promoted to `Float64` on a 64-bit system.  But I was wrong
```julia
julia> (1.0f0 + 2.0f0) / 2
1.5f0
```
so can go back to the original expression.

test/bootstrap.jl
palday: Please

test/pls.jl
dmbates: Yes.  I forget the details but I did encounter changes.  At one time I had a call to `dropzeros!` in the `sparseL` method but it is not certain it is a good idea to drop isolated zeros in the middle of a dense block so I allowed for that the be changed without affecting the test.

src/utilities.jl
dmbates: I thought I would leave that in as it doesn't seem to be harmful in any way and it could be protection against unforeseen circumstances.

palday (approved):

PR #496, opened 2021-03-26
update standard tests to include 1.6

palday:
I've removed 1.5 from the tests because we're already doing a lot of testing and if things work on 1.4 and 1.6, they probably work on 1.5. 

As soon as 1.6 is officially declared LTS and 1.7 is released, then we can test on `[1.6, 1]` for Tier 1 and `[1.4]` for Tier 2 (or whatever the last version of Julia with all the linear algebra features we need is).

### DIFF
--- a/.github/workflows/Tier1.yml
+++ b/.github/workflows/Tier1.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


dmbates (approved):

dmbates:
I approved too quickly.  There is still one place - the last line of `Tier1.yml` - where 1.5 should be changed to 1.6

PR #500, opened 2021-04-06
take advantage of multiple dispatch for GLMM comparability check

palday:


src/likelihoodratiotest.jl
```diff
@@ -166,9 +166,12 @@ function _iscomparable(m::LinearMixedModel...)
     true
 end
 
+_samefamily(::GeneralizedLinearMixedModel{T,S}...) where {T, S<:Distribution} = true
```
ararslan, line 169:
```suggestion
_samefamily(::GeneralizedLinearMixedModel{T, S}...) where {T, S} = true
```
The subtyping restriction is redundant here since a GLMM object can't have a `S` that isn't a `Distribution`.

src/likelihoodratiotest.jl
```diff
@@ -166,9 +166,12 @@ function _iscomparable(m::LinearMixedModel...)
...
+_samefamily(::GeneralizedLinearMixedModel{T, <:Distribution}...) where {T} = false
```
ararslan, line 170:
```suggestion
_samefamily(::GeneralizedLinearMixedModel...) = false
```
The `T` specification isn't useful in this case, and saying that the `S` is allowed to vary is covered by being disjoint from the above method.

ararslan (approved):

PR #501, opened 2021-04-09
update mime show for RE without FE

palday:
These models are well defined so we should `show` them correctly.


dmbates (approved):
Looks good.  Thanks.

PR #506, opened 2021-04-11
Define StructType for OptSummary

dmbates:
- add dependencies on `StructTypes` and `JSON3`
- define `StructType` and `excludes` for `OptSummary`
- still needs tests


src/optsummary.jl
```diff
@@ -110,3 +110,6 @@ function NLopt.Opt(optsum::OptSummary)
     end
     opt
 end
+
+StructTypes.StructType(::Type{<:OptSummary}) = StructTypes.Mutable()
```
palday, line 114: Since `OptSummary` is a concrete type, there can't be subtypes and the subtype operator can't really do anything.

palday (commented):
If we define a stable serialization format and a stable set of *properties*, we might consider making the set of *fields* an implementation detail, in which case a lot of the breaking optimizations we've done lately would no longer be breaking.

src/optsummary.jl
dmbates: The subtype operator is necessary because `OptSummary` is a parametric type.  The concrete types are `OptSummary{Float64}`, etc.  I had drafted an issue on whether it makes sense to continue to have `OptSummary` be a parametric type - it is closely tied to the `NLopt` package for which the only floating point type that can be used is `Float64` - but I didn't post it.

Anyway, the choices are to use the subtype operator here or to remove the parameter for `OptSummary` and I chose to keep the parameter.  It's a house of cards in a way - if we remove the parameter on `OptSummary` we might as well remove it everywhere.  I think it is better to retain the redundant parameter on `OptSummary` and modify that struct if we decide to incorporate other optimizers.

palday: Ah, yeah I had forgotten that `OptSummary` is parametric. I think it's worth keeping around the type parameter. On my long list of dream projects is changing the internals to take advantage of `MathOptInterface`, which I think NLopt finally supports. Then we could trivially support additional optimizers.

palday:
I don't have a better idea for names right now (and I think we should probably avoid using `save` or `restore` in exported functions), but maybe @ararslan  or @kleinschmidt do? They both seem to have a knack for detecting unfortunate naming conventions.

But: the restore function needs a bang, doesn't it (like `fit!`)?

dmbates:
@palday Agree that the name of the restore function should end with `!`.

dmbates:
I'm not sure that I am being sufficiently paranoid about the tests built into `restoreoptsum!` here but this should be a start.

src/linearmixedmodel.jl
```diff
@@ -782,6 +782,45 @@ StatsBase.residuals(m::LinearMixedModel) = response(m) .- fitted(m)
...
+function restoreoptsum!(m::LinearMixedModel, io::IO)
+    dict = JSON3.read(io)
+    ops = m.optsum
+    okay = (setdiff(propertynames(ops), keys(dict)) == [:lowerbd]) &&
+        isapprox(ops.initial, copy(dict.initial)) &&
+        isapprox(ops.xtol_abs, copy(dict.xtol_abs)) &&
+        ops.ftol_rel == dict.ftol_rel &&
+        ops.xtol_rel == dict.xtol_rel &&
+        ops.ftol_abs == dict.ftol_abs &&
+        ops.maxfeval == dict.maxfeval &&
```
palday, line 800: I'm wondering if we need this strict criterion. What happens if a user creates and fits a model, save its optsum, then modifies these parameters and then decides "nope, let's go back to the original fit"? 

Of course, not checking this would result in "clobbering" behavior.

test/pls.jl
```diff
@@ -401,6 +401,23 @@ end
...
+        saveoptsum(seekstart(io), fm)
+        m = LinearMixedModel(fm.formula, MixedModels.dataset(:sleepstudy))
+        restoreoptsum!(m, seekstart(io))
+        @test loglikelihood(fm) ≈ loglikelihood(m)
+        @test bic(fm) ≈ bic(m)
+        @test coef(fm) ≈ coef(m)
+            # using a temporary file for saving JSON
+        fnm = first(mktemp())
+        saveoptsum(fnm, fm)
+        restoreoptsum!(m, fnm)
```
palday, line 416:
```suggestion
        m = LinearMixedModel(fm.formula, MixedModels.dataset(:sleepstudy))        
        restoreoptsum!(m, fnm)
```

This guarantees that we're not just seeing any lingering state from the earlier tests.

palday (approved):
Looks good. Thoughts for future PRs:

1. GLMM. I think it actually should be pretty easy to do.
2. using the same logic here to update a model to have the fit from a particular bootstrap iteration
3. serializing bootstrap this way

(Future PRs because I suspect the devil is in the details for all of those.)

src/linearmixedmodel.jl
dmbates: On thinking this over I favor the "clobbering" behavior.  The assumption is that the saved values came from a bona fide optimization and it makes sense to then restore them all.  The values in the model before being fit are the defaults and not necessarily meaningful.

I will add code to check that `objective(updateL!(setθ!(m, dict.final)))` is approximately `dict.fmin` and also compare `dict.finitial` to the evaluation at `dict.initial`.  (I will loosen the type signature of the second argument to `setθ!` to allow for this.)  We can't guarantee that we have the same model fit until we preserve the formula and contrasts but at least this will mean that the model fits are consistent.

PR #508, opened 2021-04-12
LRT  for comparing non-mixed to mixed models

### LINKED ISSUE #491: Likelihood ratio test to compare `Linear[Mixed]Model`
Likelihood ratio tests can be used to assess the suitability of random effects terms, but it's currently not possible to directly compare a `LinearModel` and a `LinearMixedModel` to assess the suitability of no random effects vs. some random effects.  This would require adding methods to determine whether a `LinearModel` is nested within a `LinearMixedModel` and maybe a specialized test method to make sure you're getting the right likelihood value (e.g. can't use the deviance since the baseline is different for the two).

palday:
Closes #491.

This only allows one non-mixed model and it has to be the first argument. That's fairly restrictive, but generally speaking, I want to avoid testing a massive series of increasingly complicated fixed-effects models and then adding random-effects onto that series.


src/likelihoodratiotest.jl
```diff
@@ -46,12 +46,24 @@ Base.getindex(lrt::LikelihoodRatioTest, s::Symbol) = getfield(lrt,s)
...
+!!! note
+    The nesting of the models is not checked.  It is incumbent on the user
+    to check this. This differs from [`StatsModels.lrtest`](@ref) as nesting in
+    mixed models, especially in the random effects specification, may be non obvious.
+
+!!! note
+    For comparisons between mixed and non-mixed models, the deviance for the non-mixed
+    model is taken to be -2 log likelihood, i.e. omitting the additive constant for the
+    fully saturated model. This is in line with the computation of the deviance for mixed
+    models.
```
palday, line 66: @dmbates I can feel you looking over my shoulder and shaking your head when I write something like "the deviance  isn't really the deviance".  :slightly_smiling_face:  Do you have a better way of handling / labelling this?

src/likelihoodratiotest.jl
```diff
@@ -215,3 +255,49 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+`m1` can be obtained by constraining some parameters in `m2`.
+Both models must have been fitted on the same data. The mixed model must be
+fitted with maximum likelihood (i.e. **not** REML).
+"""
+function StatsModels.isnested(m1::TableRegressionModel{<:Union{LinearModel, GeneralizedLinearModel}},
+                              m2::MixedModel; atol::Real=0.0)
+    StatsModels.isnested(m1.model, m2) || return false
+
+    # check that the nested fixef are a subset of the outer
+    all(in.(coefnames(m1),  Ref(coefnames(m2)))) || return false
```
kleinschmidt, line 276: I think this has a curried form:
```suggestion
    all(in(coefnames(m2)), coefnames(m1)) || return false
```

src/likelihoodratiotest.jl
```diff
@@ -215,3 +255,49 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+    nobs(m1) == nobs(m2) || return false
+
+    Distribution(m1) == Distribution(m2) ||
+        throw(ArgumentError("Models must be fit to the same distribution"))
+
+    Link(m1) == Link(m2) ||
+        throw(ArgumentError("Models must have the same link function"))
+
+    return true
+end
```
kleinschmidt, line 303: For these could you check somehow the width of the design matrix (just the fixed part for the mixed models)

src/likelihoodratiotest.jl
```diff
@@ -215,3 +255,49 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+"""
+    isnested(m1::LinearModel, m2::LinearMixedModel; atol::Real=0.0)
+    isnested(m1::TableRegressioLinearModel{LinearModel}, m2::LinearMixedModel; atol::Real=0.0)
+    isnested(m1::GeneralizedLinearModel, m2::GeneralizedLinearMixedModel; atol::Real=0.0)
+    isnested(m1::TableRegressioLinearModel{GeneralizedLinearModel}, m2::GeneralizedLinearMixedModel; atol::Real=0.0)
+
+Indicate whether model `m1` is nested in model `m2`, i.e. whether
+`m1` can be obtained by constraining some parameters in `m2`.
+Both models must have been fitted on the same data. The mixed model must be
+fitted with maximum likelihood (i.e. **not** REML).
```
kleinschmidt, line 269: Add a note here too saying that the nesting isn't strictly checked (and in the case of the non-formula-bearing GLMs, isn't really checked at all)?

kleinschmidt (commented):
I think I understand why true nesting isn't checked (it's hard to do in general) but it feels like a potential footgun...I suggested a few places where you could be clearer in the docstrings and/or add some additional heuristics for non-formula-bearing GLMs.

src/likelihoodratiotest.jl
palday: Actually, this should just be a method of our internal `_is_comparable` because `lrtest` doesn't allow models of different types anyway, and the documentation for `likelihoodratiotest` already discusses the lack of checking.

src/likelihoodratiotest.jl
palday: Hmmm.... yeah. That sounds like a basic safety check.

src/likelihoodratiotest.jl
palday: Apparently, it does!

That saves a small allocation, but makes it less readable for me.  But good to know!

src/likelihoodratiotest.jl
dmbates: The important point is that the difference in -2 log likelihood is the same as the difference in deviances would be if they could be calculated.  Of course, now that I have written that I wonder if it makes sense when comparing a LinearModel and a LinearMixedModel.  We know what the saturated model for a LinearModel is but we don't for a LinearMixedModel.  My statement seems to indicate that the deviance for the saturated model in a LinearMixedModel would then have to be the same as that of a LinearModel but I am not sure that makes sense.  I am starting to confuse myself.

dmbates:
> I think I understand why true nesting isn't checked (it's hard to do in general) but it feels like a potential footgun...I suggested a few places where you could be clearer in the docstrings and/or add some additional heuristics for non-formula-bearing GLMs.

There is a way of checking nesting - approximately checking due to floating point, of course.  The columns of the model matrix for the linear model should be in the column space of `m.X` where `m` is the linear mixed model.  If you take a QR decomposition of `m.X` and multiply the linear model's model matrix by `Q'` then in each of the columns of the product, the squared length of the last `n - rank(m.X)` positions should be negligible compared to the squared length of the first `rank(m.X)` positions.

src/likelihoodratiotest.jl
palday: I'm glad I'm not the only one who struggles with some of this stuff. :smile: 

I think the relevant bit here is actually that the test statistic for LRT is -2 log(L_1/L2) which then is just the difference of -2 log likelihood. Using the true deviance was always just a convenience because the additive constants cancel out, when both models have the same constant.

All that said, my question is: should we change the label of that column and/or the internal fields? I would prefer to keep the internal fieldnames, but changing the `show` method's output is trivial.

palday:
> There is a way of checking nesting - approximately checking due to floating point, of course. The columns of the model matrix for the linear model should be in the column space of `m.X` where `m` is the linear mixed model. If you take a QR decomposition of `m.X` and multiply the linear model's model matrix by `Q'` then in each of the columns of the product, the squared length of the last `n - rank(m.X)` positions should be negligible compared to the squared length of the first `rank(m.X)` positions.

I like this idea (and that's what the tolerance kwarg is for, I suppose), but I think that should be a different PR because we can add that to the nesting checks for two `MixedModel`s.

src/likelihoodratiotest.jl
dmbates: Hmm. For a linear mixed model it would be more accurate to label the column as `-2 loglik`.  For a GLMM I think it should be `deviance`.  I believe that for a `Binomial` GLMM that is not also `Bernoulli` (e.g. `cbpp`) the quantity in that column will not be `-2 loglik`

palday: Ah, yes, that is true. And indeed the `loglikelihood` and `deviance` for GLMM are defined independently. 

Let me see what it would take to add a conditional to the `show` method --I think we'll need to add something to the `LikelihoodRatioTest` struct because right now there is no notion of which model it came from.

src/likelihoodratiotest.jl
```diff
@@ -218,3 +270,91 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+        throw(ArgumentError("REML-fitted models cannot be compared to linear models"))
+
+    return true
+end
+
+function _iscomparable(m1::GeneralizedLinearModel, m2::GeneralizedLinearMixedModel)
+    nobs(m1) == nobs(m2) || return false
+
+    # XXX This reaches into the internal structure of GLM
+    size(m1.pp.X, 2) <= size(m2.X, 2) || return false
```
kleinschmidt, line 306: Can use `StatsBase.modelmatrix` for both models to get them:
```suggestion
    size(modelmatrix(m1), 2) <= size(modelmatrix(m2), 2) || return false
```

src/likelihoodratiotest.jl
```diff
@@ -218,3 +270,91 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+
+"""
+    _isnested(x::AbstractMatrix, y::AbstractMatrix; atol::Real=0.0)
+
+Test whether the column span of `x` is a subspace of (nested within)
+the column span of y.
+
+The nesting of the column span of the fixed-effects model matrices is a necessary,
+but not sufficient condition for a linear model (whether mixed-effects or not)
+to be nested within a linear mixed-effects model.
```
kleinschmidt, line 327: What kind of situations would you have a nested fixed effects matrix but non-nested models?  (Assuming that the links/data are the same)

dmbates:
The CI failures on Julia v1.4 are because the newly added code calls `contains` which doesn't exist in v1.4.

Should we continue to test on 1.4?  Is it because that is currently the LTS release?

src/likelihoodratiotest.jl
```diff
@@ -218,3 +270,91 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+    compatible with the `atol` argument of `StatsModels.isnested`.
+"""
+function _isnested(x::AbstractMatrix, y::AbstractMatrix; rtol=1e-8, ranktol=1e-8)
+
+    # technically this can return false positives if x or y
+    # are rank deficient, but either they're rank deficient
+    # in the same way (b/c same data) and we don't care OR
+    # it's not the same data/fixef specification and we're
+    # extra conservative
+    # size(x, 2) <= size(y, 2) || return false
```
kleinschmidt, line 340: Why is this commented out?  Seems like you can avoid some potentially expensive computation if someone has done something dumb...

src/likelihoodratiotest.jl
```diff
@@ -218,3 +270,91 @@ function StatsModels.isnested(m1::MixedModel, m2::MixedModel; atol::Real=0.0)
...
+    return true
+end
+
+function _iscomparable(m1::GeneralizedLinearModel, m2::GeneralizedLinearMixedModel)
+    nobs(m1) == nobs(m2) || return false
+
+    # XXX This reaches into the internal structure of GLM
+    size(m1.pp.X, 2) <= size(m2.X, 2) || return false
+
+    _isnested(m1.pp.X, m2.X) || return false
```
kleinschmidt, line 308: same here
```suggestion
    _isnested(modelmatrix(m1), modelmatrix(m2)) || return false
```

kleinschmidt (approved):
I didn't look closely at the matrix checking stuff because it's a bit over my head but otherwise looks good.

src/likelihoodratiotest.jl
palday: Woops, I wanted to make sure the other stuff worked first :)

NEWS.md
```diff
@@ -36,6 +36,14 @@ Run-time formula syntax
 * Methods for `Base./(::AbstractTerm, ::AbstractTerm)` are added, allowing
   nesting syntax to be used with `Term`s at run-time as well [#470]
 
+MixedModels v3.6.0 Release Notes
+========================
+* Add `likelihoodratiotest` method for comparing non-mixed (generalized) linear models to (generalized) linear mixed models [#508].
+
+MixedModels v3.5.2 Release Notes
+========================
+* Explicitly deprecate non-functional `named` kwarg in `ranef` in favor of `raneftables` [#507].
```
dmbates, line 45:
```suggestion
* Explicitly deprecate vestigial `named` kwarg in `ranef` in favor of `raneftables` [#507].
```
I was reading `non-functional` as "not a function type of argument" instead of "no longer used argument"

src/likelihoodratiotest.jl
```diff
@@ -6,17 +6,20 @@ Results of MixedModels.likelihoodratiotest
...
 * `formulas`: Vector of model formulae
 * `models`: NamedTuple of the `dof` and `deviance` of the models
-* `tests`: NamedTuple of the sequential `dofdiff`, `deviancediff`, and resulting `pvalues`
+* `tests`: NamedTuple of the sequential `dofdiff`, `deviancediff`,
+           and resulting `pvalues`
 
 ## Properties
-* `deviance`
+* `deviance` : note that this is actually -2 log likelihood for linear models
+               (i.e. without the additive constant for a saturated model)
```
dmbates, line 14:
```suggestion
               (i.e. without subtracting the constant for a saturated model)
```
To me, an "additive constant" sounds like it is going to be added to the value, not subtracted from the value.

dmbates (approved):

src/likelihoodratiotest.jl
palday: I was thinking of the data being different when I wrote that. But I think I mostly wanted to highlight that that function only checks the fixed-effect model matrix and not e.g. the link.

NEWS.md
palday: Ah, yeah, I agree!

palday:
@dmbates The current LTS is 1.0! I'm happy to drop 1.4 and 1.5 with MixedModels 4.0, if for no other reason than CI speed. If/when 1.6 is declared LTS, my preference would be to have to LTS to change out tiered support architecture to "LTS" and "current release".

PR #515, opened 2021-04-29
Remove DataFramesMeta from the docs

### LINKED ISSUE #512: Remove DataFramesMeta from docs
All of the stuff we use it for can be done with just DataFrames and `filter`.

palday:
Closes #512


dmbates (approved):
Looks good.  Thanks.

Issue #514, opened 2021-04-26
StatsBase Modelling API


palday:
- [ ] dof_residual(model::RegressionModel)
- [x] fitted(model::RegressionModel)
- [x] modelmatrix(model::RegressionModel)
- [x] predict(model::RegressionModel)
- [x] residuals(model::RegressionModel)
- [x] response(model::RegressionModel)
- [ ] cooksdistance(model::RegressionModel)
- [ ] crossmodelmatrix(model::RegressionModel)
- [x] leverage(model::RegressionModel)
- [ ] meanresponse(model::RegressionModel)
- [ ] predict!(model::RegressionModel)
- [ ] responsename(model::RegressionModel) 

- [ ] ~adjr2(model::StatisticalModel)~ won't implement here
- [ ] ~adjr2(model::StatisticalModel, variant::Symbol)~ won't implement here
- [x] coef(model::StatisticalModel)
- [x] coefnames(model::StatisticalModel)
- [ ] coeftable(model::StatisticalModel)
- [ ] confint(model::StatisticalModel)
- [x] deviance(model::StatisticalModel)
- [x] dof(model::StatisticalModel)
- [x] fit(model::StatisticalModel, args...)
- [x] fit!(model::StatisticalModel, args...)
- [x] loglikelihood(model::StatisticalModel)
- [ ] model_response(obj::StatisticalModel)
- [ ] nobs(model::StatisticalModel)
- [ ] ~nulldeviance(model::StatisticalModel)~ not clearly defined (i.e. is "intercept-only" in just the fixed effects, just the random effects or both?)
- [ ] ~nullloglikelihood(model::StatisticalModel)~ not clearly defined
- [ ] ~r2(model::StatisticalModel)~
- [ ] ~r2(model::StatisticalModel, variant::Symbol)~
- [x] stderror(model::StatisticalModel)
- [x] vcov(model::StatisticalModel)
- [x] aic(model::StatisticalModel)
- [x] aicc(model::StatisticalModel)
- [x] bic(model::StatisticalModel)
- [ ] informationmatrix(model::StatisticalModel; expected)
- [ ] isfitted(model::StatisticalModel)
- [ ] islinear(model::StatisticalModel)
- [ ] mss(model::StatisticalModel)
- [ ] rss(model::StatisticalModel)
- [ ] score(model::StatisticalModel)
- [ ] weights(model::StatisticalModel)

dmbates:
I can work on some of these if that would help.

[...]

PR #520, opened 2021-05-10
Add methods for StatsBase API

dmbates:
Add some of the methods for #514

### DIFF
--- a/src/mixedmodel.jl
+++ b/src/mixedmodel.jl
@@ -55,6 +55,14 @@ StatsBase.isfitted(m::MixedModel) = m.optsum.feval > 0
 
 StatsBase.meanresponse(m::MixedModel) = mean(m.y)
 
+"""
+    modelmatrix(m::MixedModel)
+
+Returns the model matrix `X` for the fixed-effects parameters, as returned by [`coef`](@ref).
+
+This is always the full model matrix in the original column order and from a field in the model
+struct.  It should be copied if it is to be modified.
+"""
 StatsBase.modelmatrix(m::MixedModel) = m.feterm.x
 
 StatsBase.nobs(m::MixedModel) = length(m.y)
@@ -71,7 +79,7 @@ end
 """
     raneftables(m::MixedModel; uscale = false)
 
-Return the conditional means of the random effects as a NamedTuple of columntables
+Return the conditional means of the random effects as a `NamedTuple` of columntables
 """
 function raneftables(m::MixedModel{T}; uscale = false) where {T}
     NamedTuple{fnames(m)}((map(retbl, ranef(m, uscale=uscale), m.reterms)...,))
@@ -79,6 +87,15 @@ end
 
 StatsBase.residuals(m::MixedModel) = response(m) .- fitted(m)
 
+"""
+    response(m::MixedModel)
+
+Return the response vector for the model.
+
+For a linear mixed model this is a `view` of the last column of the `XyMat` field.
+For a generalized linear mixed model this is the `m.resp.y` field.
+In either case it should be copied if it is to be modified.
+""" 
 StatsBase.response(m::MixedModel) = m.y
 
 function StatsBase.responsename(m::MixedModel)


src/generalizedlinearmixedmodel.jl
```diff
@@ -445,6 +445,8 @@ end
 getθ(m::GeneralizedLinearMixedModel)  = copy(m.θ)
 getθ!(v::AbstractVector{T}, m::GeneralizedLinearMixedModel{T}) where {T} = copyto!(v, m.θ)
 
+StatsBase.islinear(m::GeneralizedLinearMixedModel) = isa(GLM.Link, GLM.IdentityLink)
```
dmbates, line 448: The documentation for `StatsBase,islinear` is vague.  I am interpreting it to mean that the linear predictor is the conditional mean.

src/generalizedlinearmixedmodel.jl
```diff
@@ -670,6 +659,7 @@ function Base.show(io::IO, ::MIME"text/plain", m::GeneralizedLinearMixedModel{T,
     println(io)
     print.(Ref(io), lpad.(nums, fieldwd))
     println(io)
+    println(io)
```
dmbates, line 662: This was just fixing up the vertical spacing in the show results.

dmbates:
@palday I think this is ready to go if the CI checks out.  Can you take a look at it and pull the trigger if you agree?  I can make the modifications to #514 after this lands.  It would probably be a good idea to make a separate section of the docs to describe methods for generics declared in `StatsBase`.  I can start on that if you agree.

src/mixedmodel.jl
```diff
@@ -6,6 +10,17 @@ Return a vector of condition numbers of the λ matrices for the random-effects t
 """
 LinearAlgebra.cond(m::MixedModel) = cond.(m.λ)
 
+function StatsBase.dof(m::MixedModel)
+    m.feterm.rank + length(m.parmap) + dispersion_parameter(m)
+end
+
+function StatsBase.dof_residual(m::MixedModel)
```
palday, line 17:
```suggestion
"""
    dof_residual(m::MixedModel)

  Return the residual degrees of freedom of the model.
    
!!! note
  The residual degrees of freedom for mixed-effects models is not clearly defined due to partial pooling. 
  The classical `nobs(m) - dof(m)` fails to capture the extra freedom granted by the random effects, but 
  `nobs(m) - nranef(m)` would overestimate the freedom granted by the random effects. `nobs(m) - sum(leverage(m))` 
  provides a nice balance based on the relative influence of each observation, but is computational 
  expensive for large models. This problem is also fundamentally related to [long-standing debates](https://bbolker.github.io/mixedmodels-misc/glmmFAQ.html#why-doesnt-lme4-display-denominator-degrees-of-freedomp-values-what-other-options-do-i-have) 
  about the appropriate treatment of the denominator degrees of freedom for ``F``-tests.
  In the future, MixedModels.jl may provide additional methods allowing the user to choose the computation 
  to use.

!!! warning
   Currently, the residual degrees of freedom is computed as `nobs(m) - dof(m)`, but this may change in 
   the future without being considered a breaking change because there is no canonical definition of the 
   residual degrees of freedom in a mixed-effects model.
"""
function StatsBase.dof_residual(m::MixedModel)
```

src/mixedmodel.jl
```diff
@@ -15,6 +30,19 @@ Equality comparisons are used b/c small non-negative θ values are replaced by 0
...
 issingular(m::MixedModel, θ=m.θ) = any(lowerbd(m) .== θ)
 
+# FIXME: better to base this on m.optsum.returnvalue
+StatsBase.isfitted(m::MixedModel) = m.optsum.feval > 0
+
+StatsBase.meanresponse(m::MixedModel) = mean(m.y)
+
+StatsBase.modelmatrix(m::MixedModel) = m.feterm.x
+
+StatsBase.model_response(m::MixedModel) = m.y
```
palday, line 40:
```suggestion
```

Looking at the StatsBase source, we can skip this if we have `reponse`:

```julia
@deprecate model_response(obj::StatisticalModel) response(obj)
```

test/pls.jl
```diff
@@ -34,11 +34,17 @@ end
     @test fm1.optsum.initial == ones(1)
     fm1.θ = ones(1)
     @test fm1.θ == ones(1)
+    @test islinear(fm1)
+    @test responsename(fm1) == "yield"
+    @test meanresponse(fm1) ≈ 1527.5
+    @test modelmatrix(fm1) == ones(30, 1)
+    @test weights(fm1) == ones(30)
```
palday, line 41: can you add a test for a linear model that has non unit weights?

src/mixedmodel.jl
```diff
@@ -15,6 +30,19 @@ Equality comparisons are used b/c small non-negative θ values are replaced by 0
 """
 issingular(m::MixedModel, θ=m.θ) = any(lowerbd(m) .== θ)
 
+# FIXME: better to base this on m.optsum.returnvalue
+StatsBase.isfitted(m::MixedModel) = m.optsum.feval > 0
+
+StatsBase.meanresponse(m::MixedModel) = mean(m.y)
+
+StatsBase.modelmatrix(m::MixedModel) = m.feterm.x
```
palday, line 38: With the new Xymat, the things that are non copies are explicitly views, right? Then it's pretty obvious based on the return type whether we're returning a copy or not. I'm wondering if we should document somewhere which of these return views, references and copies.

test/pls.jl
dmbates: Grumble, grumble - yes, okay.

src/mixedmodel.jl
dmbates: Probably.  Do you have any suggestions of where "somewhere" should be?

palday: Perhaps in the docstring? Then it's right there, both in code and where users look first.

docs/src/api.md
```diff
@@ -0,0 +1,28 @@
...
+## Non-Exported Functions
+
+Note that unless discussed elsewhere in the online documentation, non-exported functions should be considered implementation details.
+
+```@autodocs
+Modules = [MixedModels]
+Public  = false
+Order   = [:function]
+Filter = f -> !startswith(string(f), "_")
+```
```
dmbates, line 28: How is Public/Private determined?  Several of the functions documented in the `Non-Exported` section are in fact exported.

src/mixedmodel.jl
dmbates: Turns out that the `modelmatrix` method returns a stored matrix (`m.feterm.x`), not a view or a copy.  The `FeTerm` field contains the original model matrix from the `StatsModels` function and the `Xymat` field contains the augmented matrix with `y` in the last column.  If the model has been weighted then `Xymat` contains the original and the weighted `Xy` matrices.  For unweighted cases there are two references to the same matrix.

docs/src/api.md
palday: My understanding was that it's based on exporting. I'll see what I can dig up...

dmbates: I was also a bit surprised to see that in one case the key is `Private` and in the other it is `Public`.  I suppose I could just read the documentation for myself to see what is going on.

palday: I wonder if the keys do different things and it makes sense to specify both...

docs/src/constructors.md
```diff
@@ -260,9 +210,6 @@ DisplayAs.Text(ans) # hide
 ## Fitting generalized linear mixed models
 
 To create a GLMM representation
-```@docs
-GeneralizedLinearMixedModel
-```
 the distribution family for the response, and possibly the link function, must be specified.
```
palday, line 213: I can't make a suggestion on a deleted line, but can you remove the linebreak and add a comma?

src/generalizedlinearmixedmodel.jl
```diff
@@ -749,6 +737,11 @@ For Gaussian models, this parameter is often called σ².
 """
 varest(m::GeneralizedLinearMixedModel{T}) where {T} = dispersion_parameter(m) ? dispersion(m, true) : missing
 
+function StatsBase.weights(m::GeneralizedLinearMixedModel)
+    wts = m.wt
+    isempty(wts) ? ones(eltype(wts), nobs(m)) : wts
+end
```
palday, line 743:
```suggestion
function StatsBase.weights(m::GeneralizedLinearMixedMode{T}) where {T}
    wts = m.wt
    isempty(wts) ? ones(T, nobs(m)) : wts
end
```

src/mixedmodel.jl
```diff
@@ -6,6 +10,37 @@ Return a vector of condition numbers of the λ matrices for the random-effects t
...
+"""
+    dof_residual(m::MixedModel)
+
+  Return the residual degrees of freedom of the model.
+    
+!!! note
+  The residual degrees of freedom for mixed-effects models is not clearly defined due to partial pooling. 
+  The classical `nobs(m) - dof(m)` fails to capture the extra freedom granted by the random effects, but 
+  `nobs(m) - nranef(m)` would overestimate the freedom granted by the random effects. `nobs(m) - sum(leverage(m))` 
+  provides a nice balance based on the relative influence of each observation, but is computational
```
palday, line 26:
```suggestion
  provides a nice balance based on the relative influence of each observation, but is computationally 
```

docs/src/constructors.md
dmbates: Incorporated in a separate commit.

dmbates:
I don't understand why the CI checks are being cancelled, other than vague conspiracy theories about "cancel culture". :smiley_cat:

palday:
@dmbates  I've noticed some Github issues in the last few days. For some of it, I wouldn't be surprised if the JuliaStats organization is hitting usage limits, but I suspect there are also other issues since I've seen it occur for repos outside of JuliaStats.

dmbates:
@palday Is there anything further you would like me to address in this PR?  We can wait until you are settled back on this continent if you wish - there is nothing urgent here.

palday (approved):
lgtm for this PR. @dmbates Can you file an issue reminding us to document copy, reference, view for the various accessor methods (e.g. `response`)?

PR #522, opened 2021-05-14
Add full API listing

palday:
> @palday I think this is ready to go if the CI checks out.  Can you take a look at it and pull the trigger if you agree?  I can make the modifications to #514 after this lands.  It would probably be a good idea to make a separate section of the docs to describe methods for generics declared in `StatsBase`.  I can start on that if you agree.

_Originally posted by @dmbates in https://github.com/JuliaStats/MixedModels.jl/issues/520#issuecomment-841401385_

Here's an example of using `@autodoc` to generate the full API listing. It highlights that we have a few truly private functions (like `cpad`) that we should probably mark as such. 

I also want to figure out why `deviance` is leading to a Documenter warning for the autodocs...

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -14,6 +14,7 @@ makedocs(
         "bootstrap.md",
         "rankdeficiency.md",
         "mime.md",
+        "api.md",
         # "SimpleLMM.md",
         # "MultipleTerms.md",
         # "SingularCovariance.md",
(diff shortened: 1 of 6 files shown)


dmbates:
While working on the docs we may want to look at removing the `jmd` and `md` directories.

dmbates:
Shall we cherry-pick these changes into #520 and handle the method declarations and documentation in the same PR?

palday:
Yes to all of these suggestions.

I'll move the `jmd` and `md` to https://github.com/RePsychLing/MixedModelsTutorials.jl. 

I have some ideas for restructuring that repo -- I think it would be a good spot to park some nice worked examples that we don't want in our own docs because then they would take forever to build.

kliegl:
Exactly my thinking, too. Also befitting the name ...

PR #523, opened 2021-05-14
add support for BlockedSparse in `sparseL`

palday:
I'll immediately backport this to 3.7.1 if it looks fine.

cc @kliegl

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,10 @@ Run-time formula syntax
 * Methods for `Base./(::AbstractTerm, ::AbstractTerm)` are added, allowing
   nesting syntax to be used with `Term`s at run-time as well [#470]
 
+MixedModels v3.7.1 Release Notes
+========================
+* Add support for `condVar` for models with a BlockedSparse structure [#523]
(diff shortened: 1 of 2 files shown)


dmbates:
IIRC `BlockedSparse` blocks only occur in `A`, not in `L`, so the method may never get called in the normal course of events.  I'll check but I want to look at some other issues first.

palday:
@kliegl has an example where this occurs in one of the huge models.

dmbates:
I had it backwards - `BlockedSparse` can occur in `L` but not in `A`.  When creating `A` and `L` the block for `L` is created first then the block for `A` is created from it.  It is at that point that the `BlockedSparse` is converted to a `SparseMatrixCSC`.
```julia
function pushALblock!(A, L, blk)
    push!(L, blk)
    push!(A, deepcopy(isa(blk, BlockedSparse) ? blk.cscmat : blk))
end
```

We should watch for an example that can be used to test this (admittedly trivial) method.

dmbates (approved):

PR #526, opened 2021-05-25
LinearAlgebra.mul! instead of unscaledre!

dmbates:
- `unscaledre!` had methods that didn't really correspond to simulating and incorporating unscaled random effects.  
- These methods are rewritten here as `LinearAlgebra.mul!` methods.


palday (approved):
lgtm, thanks

PR #530, opened 2021-06-07
Add explanation for a broken test.  Correct a typo

dmbates:
I kept forgetting the reason that one of our tests is still broken and eventually decided to add a comment to remind myself.


palday (approved):
I'm a big fan of comments explaining past me to future me :smile:

PR #539, opened 2021-07-07
Add a progress meter for LMM fits

dmbates:
- This is just a proof of concept but it seems to work well (once I read the documentation for ProgressMeter and read our code a little more carefully).
- Try, for example
```julia
julia> using MixedModels

julia> using MKL

julia> mrk17 = MixedModels.dataset(:mrk17_exp1);

julia> form = @formula 1000 / rt ~ 1 + F*P*Q*lQ*lT + (1+F+P+Q+lQ+lT|subj) + (1+P+Q+lQ+lT|item);

julia> contr = merge(Dict(nm => EffectsCoding() for nm in (:F, :P, :Q, :lQ, :lT)), Dict(nm => Grouping() for nm in (:subj, :item)));

julia> m = fit(MixedModel, form, mrk17; contrasts=contr);
Progress:  1316  Time: 0 Time: 0:00:07 ( 5.53 ms/it)
```
During the iterations the current objective is displayed below the Progress line,

- Not entirely sure how the `Time: 0` ends up in that line.


Project.toml
```diff
@@ -34,7 +34,7 @@ JSON3 = "1"
 LazyArtifacts = "1"
 NLopt = "0.5, 0.6"
 PooledArrays = "0.5, 1"
-ProgressMeter = "1.5"
+ProgressMeter = "1.5, 1.6, 1.7"
```
palday, line 37: for versions > 1.0, specifying a minor version just gives a lower bound

(the problem is that we have a lot of pre-1.0 dependencies where minor versions are considered breaking)

src/linearmixedmodel.jl
```diff
@@ -397,15 +397,18 @@ function fit!(m::LinearMixedModel{T}; verbose::Bool = false, REML::Bool = false)
     end
     opt = Opt(optsum)
     optsum.REML = REML
+    prog = ProgressUnknown(; showspeed=true)
     function obj(x, g)
         isempty(g) || throw(ArgumentError("g should be empty for this objective"))
         val = objective(updateL!(setθ!(m, x)))
+        ProgressMeter.next!(prog; showvalues = [(:objective, val),])
         verbose && println(round(val, digits = 5), " ", x)
```
palday, line 405:
```suggestion
        verbose && ProgressMeter.next!(prog; showvalues = [(:objective, round(val, digits = 5)), (:theta, x)])
```

palday:
The CI failures seem to be a problem with the Windows VMs.

palday:
I think I had looked at making the iterations clear and print in the same spot, but then got sidetracked: https://github.com/timholy/ProgressMeter.jl/pull/204

Project.toml
dmbates: So this change will have no effect in practice?  If so, I will back it out.

palday: correct

src/linearmixedmodel.jl
dmbates: I would prefer not to try to show `(:theta, x)` at each iteration because I think it would be too distracting.  For the example I suggested using the `:mrk17_exp1` data set, the length of `x` is 36, which makes for a lot of writing to the console and, I fear, a lot of distraction during the iterations.

I like  printing out the objective in that, even if it is changing too quickly to be able to read the numbers, you can see what parts of the objective have stabilized.  I prefer to show it in full (i.e. without the call to `round`) so that the stabilization doesn't continues to be visible even if the changes are past the 5th digit.

I agree that it would be best to preface the call to `next!` with `verbose &&`.  Should we do the same in the call to `ProgressUnknown` and `ProgressMeter.finish!`?

Shall we change the "Progress" annotation to something like "Minimizing" or "Optimizing parameters:"?

I can make similar changes in the `fit` method for GLMMs once we settle on things here.

dmbates:
CI is failing for Julia 1.4 on Windows because it can't download packages correctly.  Should we switch the CI to julia 1.6 only?

src/linearmixedmodel.jl
palday: I completely agree with the problems of a big output and am happy with that proposal -- I was mostly just mimicking the old output. :smile: 

I have a vague preference for "Minimizing"  -- it's both shorter and more specific.

palday:
Since this is for the breaking 4.0 release, let's just go ahead and make support start at 1.6. Can you also update Project.toml? (And even though it's not a feature, a note to NEWS?)

Project.toml
dmbates: I will take a look to see when ProgressUnknown and other features that we now use were introduced and set it accordingly.

src/linearmixedmodel.jl
dmbates: "Minimizing" it is then.

dmbates:
Will the changes re CI on version 1.6 (and, sigh!, add a news item - I feel like a surly teenager rolling his eyes about the stupid rules his parents have) when I get back from a walk.

palday:
@dmbates  I'll take care of it while you're out -- I have a minute while a computation runs. :smile:

dmbates:
Should the default value of `verbose` be changed to `true`?  I think this display is informative and may help educate users as to what is making their model fits slow.  On the other hand, I don't want to clutter up users' sessions with unwanted displays.

dmbates:
There is a problem with using the optional argument `verbose` for generalized linear mixed models in that it also causes the deviance evaluation to become verbose.  As a temporary fix I will add an optional `progress` argument separate from `verbose` but that is just to get things working.  We can decide on a more general approach later.

palday:
What does verbose output look like now for a quick fit like `sleepstudy`?

dmbates:
> What does verbose output look like now for a quick fit like `sleepstudy`?

For different machines it could be that nothing shows up or just the final summary.  Perhaps we should change the optional argument name to `progress` throughout and explain in the documentation that this uses the facilities of the `ProgressMeter` package which doesn't do anything if the minimization time is much less than 1 second.

By the way, I looked in the ProgressMeter package for a NEWS.md file to check on what version is needed for which capabilities and there isn't a NEWS.md file.  So there!

palday:
I like the idea of changing the argument name to `progress`. 

> By the way, I looked in the ProgressMeter package for a NEWS.md file to check on what version is needed for which capabilities and there isn't a NEWS.md file. So there!

:grin:

Project.toml
dmbates: i backed out the change

dmbates:
@palday Can you take another look, please?

src/generalizedlinearmixedmodel.jl
```diff
@@ -269,16 +274,19 @@ function fit!(
         optsum.final = copy(optsum.initial)
     end
     setpar! = fast ? setθ! : setβθ!
+    prog = ProgressUnknown("Minimizing"; showspeed=true)
     function obj(x, g)
         isempty(g) || throw(ArgumentError("g should be empty for this objective"))
         val = deviance(pirls!(setpar!(m, x), fast, verbose), nAGQ)
         verbose && println(round(val, digits = 5), " ", x)
```
palday, line 281:
```suggestion
```

src/generalizedlinearmixedmodel.jl
```diff
@@ -241,19 +241,24 @@ fit(
...
-    fit!(m::GeneralizedLinearMixedModel[, verbose = false, fast = false, nAGQ = 1])
+    fit!(m::GeneralizedLinearMixedModel[, verbose=false, fast=false, nAGQ=1, progress=true])
 
 Optimize the objective function for `m`.
 
 When `fast` is `true` a potentially much faster but slightly less accurate algorithm, in
 which `pirls!` optimizes both the random effects and the fixed-effects parameters,
 is used.
+
+if `progress` is `true`, the default, a `ProgressMeter.ProgressUnknown` counter is displayed
```
palday, line 252:
```suggestion
If `progress` is `true`, the default, a `ProgressMeter.ProgressUnknown` counter is displayed.
```

src/generalizedlinearmixedmodel.jl
```diff
@@ -241,19 +241,24 @@ fit(
...
+during the iterations to minimize the deviance.  There is a delay before this display is initialized
+and it may not be shown at all for models that are optimized quickly.
```
palday, line 254:
```suggestion
and it may not be shown at all for models that are optimized quickly.

If `verbose` is `true`, then both the intermediate results of both the nonlinear optimization and PIRLS are also displayed on standard output.
```

src/generalizedlinearmixedmodel.jl
palday: I initially thought we should remove this, but I think it actually makes sense to keep it in and have `verbose` and `progress` be completely independent for GLMM. `progress` corresponds to `progress in LMM, while `verbose` is well, exceptionally verbose and mostly interesting for diagnosing problems.

palday:
I think it's nearly there -- what do you think of my comments for `verbose` vs. `progress` in GLMM?

dmbates:
> I think it's nearly there -- what do you think of my comments for `verbose` vs. `progress` in GLMM?

Those look good, thanks.  Both committed now.

dmbates:
Shall we modify our test fits to have `progress=false`?  Right now the progress results are clogging up the output because they are not overwriting a line but instead just add to a transcript.

palday:
@dmbates Yes, then this is done. :rocket:

src/linearmixedmodel.jl
```diff
@@ -775,21 +775,22 @@ function reevaluateAend!(m::LinearMixedModel)
 end
 
 """
-    refit!(m::LinearMixedModel[, y::Vector]; REML=m.optsum.REML)
+    refit!(m::LinearMixedModel[, y::Vector]; REML=m.optsum.REML, progress::Bool=false)
```
palday, line 778:
```suggestion
    refit!(m::LinearMixedModel[, y::Vector]; REML=m.optsum.REML, progress::Bool=true)
```

for consistency with `fit`

src/linearmixedmodel.jl
```diff
@@ -775,21 +775,22 @@ function reevaluateAend!(m::LinearMixedModel)
...
 
 Refit the model `m` after installing response `y`.
 
-If `y` is omitted the current response vector is used.
+If `y` is omitted the current response vector is used.  Note that the default for
+`progress` is `false` on refits (whereas it is `true` in the `fit!` methods.)
 """
-function refit!(m::LinearMixedModel; REML=m.optsum.REML)
-    fit!(unfit!(m); REML=REML)
+function refit!(m::LinearMixedModel; REML=m.optsum.REML, progress::Bool=false)
```
palday, line 785:
```suggestion
function refit!(m::LinearMixedModel; REML=m.optsum.REML, progress::Bool=true)
```

dmbates:
@palday Took a bit longer than I imagined as I needed to add a default to refit! methods as well.  I chose `progress::Bool=false` for `refit!`.  I also replaced a couple of occurrences of the now deprecated `model_response` with `response` in the tests.

src/linearmixedmodel.jl
```diff
@@ -775,21 +775,22 @@ function reevaluateAend!(m::LinearMixedModel)
...
+    fit!(unfit!(m); REML=REML, progress=progress)
 end
 
-function refit!(m::LinearMixedModel, y; REML=m.optsum.REML)
+function refit!(m::LinearMixedModel, y; REML=m.optsum.REML, progress::Bool=false)
```
palday, line 789:
```suggestion
function refit!(m::LinearMixedModel, y; REML=m.optsum.REML, progress::Bool=true)
```

palday:
@dmbates while you were writing that comment, I was already commenting on the choice of default for `refit!`. I'm fine with `false` as the default, but `true` would be consistent with `fit!`

dmbates:
> @dmbates while you were writing that comment, I was already commenting on the choice of default for `refit!`. I'm fine with `false` as the default, but `true` would be consistent with `fit!`

My reasoning was that for a refit! the user has already had the chance to see the progress meter for the optimization so the novelty may have worn off.  I would be okay with changing the default to `true` for consistency with `fit!` but then we would need to go through all the calls to refit! in the tests to add `progress=false` to them.  Let me know if you want me to do that.

palday:
I had a similar thought -- typically the user knows better what to expect with `refit` and model is already constructed and everything is precompiled, so that bit of overhead is gone. 

We could take a middle ground that will make things easier in the future if we expose more options to the user: have a `kwargs...` to `refit!` that is splatted into the call to `fit!` and then the defaults are just carried over.

dmbates:
@palday the `kwargs...` approach sounds good.  Would you be willing to make that change?

palday:
yep, will do

test/pirls.jl
```diff
@@ -91,18 +91,25 @@ end
...
         @test gm2r.θ ≈ gm2.θ atol=1e-3
 
-        refit!(gm2r, 1 .- gm2.y; fast=false)
+        # swapping successes and failures to give us the same model
+        # but with opposite signs. healthy ≈ 1 - response(gm2r)
+        # but defining it in terms of the original values avoids some
+        # nasty floating point issues
+        healthy = @. (cbpp.hsz - cbpp.incid) / cbpp.hsz
+        refit!(gm2r, healthy; fast=false, progress=false)
+        @test length(gm2r.optsum.final) == 5
```
palday, line 105: @dmbates going the `kwargs...` route made me realize that we weren't passing all the arguments here ... which then lead to the discovery that the "flipped" response was numerically pathological for fast fits. Hence the changes here.

palday (approved):
As soon as the CIs pass, please merge. :smile:

dmbates:
I seem to be playing "whack-a-mole" again trying to track down all the instances of `fit(` or `refit!(` in the tests and adding `progress=false`.  The tests on Mac OSX indicate that there may be a problem with the bootstrap tests now.  I feel that the `parametricbootstrap` function should turn off the progress meter unconditionally.

Does it seem that we are going about this the wrong way?  I don't mind a bit of whack-a-mole but this seems to be going on longer than is wise.

By the way, the Mac tests seem to be slower and hence better at finding cases where the progress meter kicks in.

palday:
Interesting on the Mac tests ....

I agree that `parametricbootstrap` should unconditionally set `progress=false` -- it has its own progress meter anyway.

I think there's not much to be done about the whack-a-mole -- that's just what happens when you change defaults. :shrug: I have a somewhat slower computer I can also run the tests on and see if that reveals any more cases.

PR #540, opened 2021-07-15
Release 4.0

palday:
- [x] update README (should be easy to adapt highlights from NEWS or maybe just drop the new major release bit and add a reference to NEWS)
- [x] run JuliaFormatter 
- [ ] create a new branch from current `main` called `4.x` (after merging)

~@dmbates do you want to take care of the formatting?~ done!


dmbates (approved):
- This looks good to go for me whenever you are satisfied.

- I think that running the code through the formatter was a good idea.  Thanks for doing that.  In some ways consistency of style is more important than the particular style being used.

PR #543, opened 2021-07-20
Fix support table column

palday:


dmbates (approved):
Thanks for catching that.

PR #545, opened 2021-07-22
Added opt. fname argument to sparseL and condVar

dmbates:
- `sparseL` gains an optional `fname` argument.  When a grouping factor name is given for this argument the return value consists of the lower-right triangle starting at that the rows and columns for that factor.
- `condVar` gains a method with a second argument `fname`.  When given the conditional variance array for that factor only is returned.
- the default `condVar` now just iterates over the grouping factor names.  ~(The old code is retained for comparison but will never be executed - it should be removed after review of this PR.)~
- the main use of `condVar` is in `MixedModelsMakie.caterpillar` which can now be simplified somewhat.
- this should provide considerable time savings for models like those fit by @kliegl to the FFGK21 dataset when only the caterpillar plot for the Schools is desired.  This still needs to be checked out.
- Needs specific tests for the newly introduced methods.

### DIFF
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -319,7 +319,7 @@ function condVar(m::LinearMixedModel{T}, fname) where {T}
     Lblk = LowerTriangular(densify(sparseL(m; fname=fname)))
     blk = findfirst(isequal(fname), fnames(m))
     λt = Array(m.λ[blk]') .* sdest(m)
-    vsz =  size(λt, 2)
+    vsz = size(λt, 2)
     ℓ = length(m.reterms[blk].levels)
     val = Array{T}(undef, (vsz, vsz, ℓ))
     scratch = Matrix{T}(undef, (size(Lblk, 1), vsz))
@@ -329,9 +329,9 @@ function condVar(m::LinearMixedModel{T}, fname) where {T}
         ldiv!(Lblk, scratch)
         mul!(view(val, :, :, b), scratch', scratch)
     end
-    val
+    return val
 end
-    
+
 function _cvtbl(arr::Array{T,3}, trm) where {T}
     return merge(
         NamedTuple{(fname(trm),)}((trm.levels,)),
@@ -1005,9 +1005,7 @@ are to be included.
  to `fname` are dropped. The default is the first, i.e., leftmost block and hence all blocks.
 """
 function sparseL(
-    m::LinearMixedModel{T};
-    fname::Symbol=first(fnames(m)),
-    full::Bool=false,
+    m::LinearMixedModel{T}; fname::Symbol=first(fnames(m)), full::Bool=false
 ) where {T}
     L, reterms = m.L, m.reterms
     sblk = findfirst(isequal(fname), fnames(m))
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -1001,8 +1001,8 @@ Return the lower Cholesky factor `L` as a `SparseMatrix{T,Int32}`.
 `full` indicates whether the parts of `L` associated with the fixed-effects and response
 are to be included.
 
-When `fname` is the name of a grouping factor the blocks to the left of that block are
-dropped.
+`fname` specifies the first grouping factor to include. Blocks to the left of the block corresponding
+ to `fname` are dropped. The default is the first, i.e., leftmost block and hence all blocks.
 """
 function sparseL(
     m::LinearMixedModel{T};
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -313,31 +313,6 @@ diagonal blocks from the conditional variance-covariance matrix,
 """
 function condVar(m::LinearMixedModel{T}) where {T}
     return [condVar(m, fnm) for fnm in fnames(m)]
-    s = sdest(m)
-    @static if VERSION < v"1.6.1"
-        spL = LowerTriangular(SparseMatrixCSC{T,Int}(sparseL(m)))
-    else
-        spL = LowerTriangular(sparseL(m))
-    end
-    nre = size(spL, 1)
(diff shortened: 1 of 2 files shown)


src/linearmixedmodel.jl
```diff
@@ -1002,19 +1019,30 @@ function _coord(A::Matrix)
...
-    sparseL(m::LinearMixedModel{T}; full::Bool=false) where {T}
+    sparseL(m::LinearMixedModel; fname::Symbol=first(fnames(m)), full::Bool=false)
 
 Return the lower Cholesky factor `L` as a `SparseMatrix{T,Int32}`.
 
 `full` indicates whether the parts of `L` associated with the fixed-effects and response
 are to be included.
+
+When `fname` is the name of a grouping factor the blocks to the left of that block are
+dropped.
```
palday, line 1030:
```suggestion
`fname` specifies the first grouping factor to include. Blocks to the left of the block corresponding
 to `fname` are dropped. The default is the first, i.e., leftmost block and hence all blocks.
```

src/linearmixedmodel.jl
```diff
@@ -312,7 +312,7 @@ diagonal blocks from the conditional variance-covariance matrix,
     s² Λ(Λ'Z'ZΛ + I)⁻¹Λ'
 """
 function condVar(m::LinearMixedModel{T}) where {T}
-    L = m.L
+    return [condVar(m, fnm) for fnm in fnames(m)]
```
palday, line 315: This potentially recomputes the blocks to the right of `fname`  in `sparseL`  several times.

src/linearmixedmodel.jl
```diff
@@ -312,7 +312,7 @@ diagonal blocks from the conditional variance-covariance matrix,
...
     s = sdest(m)
     @static if VERSION < v"1.6.1"
```
palday, line 317: I'm tempted to up the julia compat to 1.6.1 for the whole package so that we can drop this branch.

src/linearmixedmodel.jl
dmbates: Recomputing blocks to the right, which are smaller blocks, may not be the worst outcome.  The current version, with the recompute, is in the same ballpark as computing just the first block on the `mrk17_exp1` example
```julia
julia> @benchmark condVar($m6, :item)
BenchmarkTools.Trial: 13 samples with 1 evaluation.
 Range (min … max):  371.041 ms … 403.260 ms  ┊ GC (min … max): 0.52% … 0.48%
 Time  (median):     399.209 ms               ┊ GC (median):    0.48%
 Time  (mean ± σ):   396.114 ms ±   8.839 ms  ┊ GC (mean ± σ):  0.46% ± 0.14%

  ▁                               █        ▁ ▁         ██  ▁ ▁█  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁█▁█▁▁▁▁▁▁▁▁▁██▁▁█▁██ ▁
  371 ms           Histogram: frequency by time          403 ms <

 Memory estimate: 75.74 MiB, allocs estimate: 4015.

julia> @benchmark condVar($m6, :subj)
BenchmarkTools.Trial: 336 samples with 1 evaluation.
 Range (min … max):  12.751 ms … 18.624 ms  ┊ GC (min … max): 0.00% … 11.51%
 Time  (median):     14.469 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   14.869 ms ±  1.039 ms  ┊ GC (mean ± σ):  2.87% ±  5.10%

                                              █          ▆     
  ▄▁▁▁▁▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▇▁▁▁▁▁█▁▁▁ ▆
  12.8 ms      Histogram: log(frequency) by time      14.3 ms <

 Memory estimate: 14.73 MiB, allocs estimate: 1172.

julia> @benchmark condVar($m6)
BenchmarkTools.Trial: 13 samples with 1 evaluation.
 Range (min … max):  386.563 ms … 423.627 ms  ┊ GC (min … max): 0.50% … 0.53%
 Time  (median):     417.341 ms               ┊ GC (median):    0.56%
 Time  (mean ± σ):   415.208 ms ±   9.577 ms  ┊ GC (mean ± σ):  0.55% ± 0.03%

  ▁                                      █ ▁  ▁      ▁▁  █  █    
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁█▁▁█▁▁▁▁▁▁██▁▁█▁▁█▁▁ ▁
  387 ms           Histogram: frequency by time          423 ms <

 Memory estimate: 90.46 MiB, allocs estimate: 5192.
```
I suggest that we just leave this as it is and remove the rest of the function, which is vestigial and will not be executed.

src/linearmixedmodel.jl
```diff
@@ -358,7 +350,7 @@ function condVartables(m::MixedModel{T}) where {T}
     return NamedTuple{fnames(m)}((map(_cvtbl, condVar(m), m.reterms)...,))
 end
 
-function pushALblock!(A, L, blk)
+function _pushALblock!(A, L, blk)
```
dmbates, line 353: I changed the name of this function to indicate that it was an internal utility.

src/linearmixedmodel.jl
palday: Works for me.

palday (approved):
awesome! ready to tag 4.1? :smile:

dmbates:
> awesome! ready to tag 4.1? smile

Sounds good to me.

PR #547, opened 2021-07-28
compat fixes for Julia 1.7

### LINKED ISSUE #532: Adjustments for julia-v1.7.0
When the workaround for https://github.com/JuliaLang/julia/issues/40048 is merged we can start testing on julia-v1.7.0-DEV

In `test/runtests.jl` we should replace the call to determine the BLAS in use with `BLAS.get_config()` for v1.7.0-DEV

A couple of instances of `qr(x, Val(true))` need to be replaced by `qr(x, ColumnNorm())` for recent versions of v1.7.0-DEV.

palday:
Closes #532 

- deleted a dead path now that we've dropped pre 1.6 support
- added an internal wrapper method to dodge some deprecation warnings in 1.7+


dmbates (approved):
Thank you for doing this.  Creating the `pivoted_qr` function was a good solution to avoid special-casing the code at every call to the decomposition.

I took the liberty of dropping the `_zerocorr` code while doing the code cleanup.  You are welcome to change that if you don't think it belongs here.

PR #548, opened 2021-08-05
Fix ReMat's docstring + contravariance issue

palday:
I stumbled across these issues when mucking about with the internal constructors to see if there would be an easy to do something like a multi-membership model (cf. https://bbolker.github.io/mixedmodels-misc/notes/multimember.html)


dmbates (approved):
Looks good.  Thanks.

Issue #549, opened 2021-08-12
Add `fitlog` kwarg for logging the parameter vector during optimization


palday:
As discussed in [some visualization work](https://github.com/RePsychLing/SMLP2021/pull/14#issuecomment-897900060), we should add a `fitlog=nothing` kwarg to the `fit` and `fit!` methods. In the default case, nothing changes. In the case that `fitlog` is a collection supporting `push!`, then we `push!(fitlog, SVector{length(val)}(val)` in the `obj` function being optimized.

dmbates:
One approach, which, unfortunately, would be an incompatible change in an underlying struct, would be to change the `optinfo` fields to have `initial` and `final` replaced by, say, a Vector{StaticVector{T,npar}} perhaps called `parameters`.  The convention would be that what is now `initial` would become `first(parameters)` and what is now `final` would become `last(parameters)`.  An optional integer-valued argument to `fit`, with a name reflecting the `thinning` of the parameter log, would indicate how frequently parameter vectors and objective values were to be pushed onto the vector of parameters and an accompanying vector of objective values.  The default, 0, would indicate that only the initial and final values are to be pushed onto the vector.  The condition in the objective function evaluation on whether to push would be like
```julia
    !iszero(thin) && iszero(rem(iter, thin))
``` 

I regret that it seems that the way to stimulate my thinking about incompatible changes is to release a new version of the package.

palday:
We can have it both ways -- we can introduce `fitlog` for now and when we've accumulated a few other breaking changes, do your suggestion for 5.0. I've been thinking about reworking `OptSummary` anyway so that we access NLopt through the JuMP interface. That would make it easier in the future to look at other optimizers.

dmbates:
Sounds like a plan.  What do you think of using a `thin`-type of argument instead of a yes/no for the logging to save, e.g., every 10th parameter vector?

palday:
I like the idea of adding a `thin` argument. We're rapidly approaching an MCMC-like approach to this :wink:

PR #551, opened 2021-08-17
Allow specifying sigma in LinearMixedModel

palday:
This was done to support meta-analytic type models. There might be other cases where we can use a fixed residual variation to do tricks with observation-level random effects, but I'm not sure.

I've also done some formatting tweaks that take advantage of kwarg syntax in Julia 1.5+ (since we now require Julia 1.6).

I'll tackle #549 after this is merged (or rejected) as some of the changes would create a merge conflict.

If we one day allow specifying constraints on the betas, I would suggest doing this as a new type `ConstrainedLinearMixedModel` and then define a separate `objective` method for that so that we don't keep adding conditionals to `objective`. (And reallistically, we should should also define separate `setβ!`, `setθ!` and `setβθ!` methods.)


NEWS.md
```diff
@@ -1,3 +1,12 @@
+MixedModels v4.1.0 Release Notes
+========================
+* Add support for specifying a fixed value of `σ`, the residual standard deviation,
+  in `LinearMixedModel`. `fit` takes a keyword-argument `σ`. For models constructed
+  separately, `σ` can be specified by setting the corresponding field in the `optsum`
```
dmbates, line 5:
```suggestion
  separately, `σ` can be specified by setting `optsum.sigma`. [#551]
```
The field is called `optsum.sigma` below and the phrasing used here might be interpreted as `optsum.σ`.

src/linearmixedmodel.jl
```diff
@@ -169,38 +161,16 @@ function LinearMixedModel(
...
-        contrasts=contrasts,
-        progress=progress,
-        REML=REML,
-    )
+function fit(::Type{LinearMixedModel}, f::FormulaTerm, tbl;
+             wts=[], contrasts=Dict{Symbol,Any}(), progress::Bool=true,
+             REML::Bool=false, σ=nothing)
+    return fit(LinearMixedModel, f, Tables.columntable(tbl);
+               wts, contrasts, progress, REML, σ)
 end
```
dmbates, line 169: These formatting changes might be undone by `format` with BlueStyle

palday: This might be a good time to go ahead and add a JuliaFormatter CI step -- I'll look into it.

src/linearmixedmodel.jl
```diff
@@ -662,7 +618,12 @@ Return negative twice the log-likelihood of model `m`
 function objective(m::LinearMixedModel{T}) where {T}
     wts = m.sqrtwts
     denomdf = T(ssqdenom(m))
-    val = logdet(m) + denomdf * (one(T) + log2π + log(pwrss(m) / denomdf))
+    σ = m.optsum.sigma
+    val = if σ === nothing
```
dmbates, line 622:
```suggestion
    val = if isnothing(σ)
```

src/linearmixedmodel.jl
```diff
@@ -1171,7 +1134,7 @@ end
 
 Returns the estimate of σ², the variance of the conditional distribution of Y given B.
 """
-varest(m::LinearMixedModel) = pwrss(m) / ssqdenom(m)
+varest(m::LinearMixedModel) = isnothing(m.optsum.sigma) ?  pwrss(m) / ssqdenom(m) : m.optsum.sigma
```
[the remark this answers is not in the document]
palday: For this one, I went with the conditional branch to avoid evaluating `pwrss(m)` if it's not being used.

src/linearmixedmodel.jl
```diff
@@ -860,6 +821,8 @@ function restoreoptsum!(m::LinearMixedModel, io::IO)
     end
     ops.optimizer = Symbol(dict.optimizer)
     ops.returnvalue = Symbol(dict.returnvalue)
+    # provides compatibility with fits saved before the introduction of fixed sigma
```
dmbates, line 824: good catch

dmbates (approved):

src/linearmixedmodel.jl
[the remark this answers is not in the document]
dmbates: Okay, that's a good reason for doing it your way.  I withdraw my suggestion.

PR #552, opened 2021-08-17
add `fitlog`

### LINKED ISSUE #549: Add `fitlog` kwarg for logging the parameter vector during optimization

palday:
~Addresses (but does not close) #549 (because we want to change `OptSummary` later to be smarter about this).~

Closes #549. There is a minor inefficiency in that `initial`, `final`, `finitial`, `fmin` store information duplicated in the `fitlog`. Not duplicating this information (by cheating and using properties that redirect to the relevant field entries) turns out to make the bookkeeping a lot messier, so I've let this inefficiency remain (which also keeps better compatibility). I use `Vector` inside of the log instead of `SVector` because the changing size of the parameter vector in GLMM creates problems. 

It also seems that I might have found another case where julialang/julia#40048 isn't as fixed as we had hoped -- I get a big stackoverflow type inference dump as a Heisenbug when stepping through the `contra` testset, BUT it continues afterwards with the correct result.

### DIFF
--- a/src/generalizedlinearmixedmodel.jl
+++ b/src/generalizedlinearmixedmodel.jl
@@ -331,13 +331,14 @@ function fit!(
             xmin_[i] = zero(T)
         end
     end
+    !isnothing(fitlog) && (loglength = length(fitlog))
     if xmin ≠ xmin_
         if (zeroobj = obj(xmin_, T[])) ≤ (fmin + 1.e-5)
             fmin = zeroobj
             copyto!(xmin, xmin_)
-        else
-            # remove unused zero'ing step
-            !isnothing(fitlog) && pop!(fitlog)
+        elseif !isnothing(fitlog) && (length(fitlog) > loglength)
+            # remove unused extra log entry
+            pop!(fitlog)
         end
     end
     ## ensure that the parameter values saved in m are xmin
--- a/src/generalizedlinearmixedmodel.jl
+++ b/src/generalizedlinearmixedmodel.jl
@@ -335,6 +335,9 @@ function fit!(
         if (zeroobj = obj(xmin_, T[])) ≤ (fmin + 1.e-5)
             fmin = zeroobj
             copyto!(xmin, xmin_)
+        else
+            # remove unused zero'ing step
+            pop!(fitlog)
         end
     end
     ## ensure that the parameter values saved in m are xmin
--- a/src/generalizedlinearmixedmodel.jl
+++ b/src/generalizedlinearmixedmodel.jl
@@ -174,8 +174,7 @@ function fit(
     fast::Bool=false,
     nAGQ::Integer=1,
     progress::Bool=true,
-    fitlog::Union{Nothing,AbstractVector}=nothing,
-    thin::Int=1,
+    thin::Int=typemax(Int),
 )
     return fit(
         GeneralizedLinearMixedModel,
@@ -190,7 +189,6 @@ function fit(
         fast,
         nAGQ,
         progress,
-        fitlog,
         thin,
     )
 end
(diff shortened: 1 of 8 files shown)


src/generalizedlinearmixedmodel.jl
```diff
@@ -284,16 +307,24 @@ function fit!(
     end
     setpar! = fast ? setθ! : setβθ!
     prog = ProgressUnknown("Minimizing"; showspeed=true)
+    # start from zero for the initial call to obj before optimization
+    iter = 0
     function obj(x, g)
         isempty(g) || throw(ArgumentError("g should be empty for this objective"))
         val = deviance(pirls!(setpar!(m, x), fast, verbose), nAGQ)
+        !isnothing(fitlog) && iszero(rem(iter, thin)) && push!(fitlog, (copy(x), val))
```
dmbates, line 315:
```suggestion
        !isnothing(fitlog) && isone(rem(iter, thin)) && push!(fitlog, (copy(x), val))
```
I know I suggested `iszero` but it may be better to ensure that the record starts with the first evaluation.

src/linearmixedmodel.jl
```diff
@@ -413,14 +450,22 @@ function fit!(m::LinearMixedModel{T}; progress::Bool=true, REML::Bool=false) whe
     opt = Opt(optsum)
     optsum.REML = REML
     prog = ProgressUnknown("Minimizing"; showspeed=true)
+    # start from zero for the initial call to obj before optimization
+    iter = 0
     function obj(x, g)
         isempty(g) || throw(ArgumentError("g should be empty for this objective"))
         val = objective(updateL!(setθ!(m, x)))
+        !isnothing(fitlog) && iszero(rem(iter, thin)) && push!(fitlog, (copy(x), val))
```
dmbates, line 458:
```suggestion
        !isnothing(fitlog) && isone(rem(iter, thin)) && push!(fitlog, (copy(x), val))
```
Same here.

src/generalizedlinearmixedmodel.jl
palday: The initial evaluation at theta_0 is handled explicitly below -- or did you want to capture the first step after that?

dmbates: No I didn't see that the initial evaluation is explicitly handled.  It should be okay to leave this as `iszero` then.

palday:
@dmbates I took the opportunity to add a plot showing the different convergence pattern of Nelder-Mead and BOBYQA: https://juliastats.org/MixedModels.jl/previews/PR552/optimization/#Modifying-the-optimization-process

test/pirls.jl
```diff
@@ -13,7 +13,11 @@ include("modelcache.jl")
 
 @testset "contra" begin
     contra = dataset(:contra)
-    gm0 = fit(MixedModel, only(gfms[:contra]), contra, Bernoulli(); fast=true, progress=false)
+    thin = 5
+    gm0 = fit(MixedModel, only(gfms[:contra]), contra, Bernoulli(); fast=true, progress=false, thin)
```
dmbates, line 17: That syntax leads me astray.  I keep forgetting that it expands to `, thin=thin)`

dmbates (approved):
Thanks for making these changes.  I think it is much improved.

palday:
@dmbates Are you happy with tagging a new release with this and constrained sigma? That locks us into this API until 5.0 lands.

dmbates:
Can you hold off tagging a new release until I work on the leverage changes
a bit more?  They won't be breaking changes - just a different way of
evaluating the same results.

palday:
Yep, easy.

PR #556, opened 2021-08-24
NEWS update for 545

palday:
@dmbates actually, while I've paused releasing 4.1, should we add a `ranef(model, fname)` method to complement the `condVar` method?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@ MixedModels v4.1.0 Release Notes
 * Faster version of `leverage` for `LinearMixedModel` allowing for experimentation
   using the sum of the leverage values as an empirical degrees of freedom for the
   model. [#553], see also [#535]
+* Optimized version of `condVar` with an additional method for extracting only the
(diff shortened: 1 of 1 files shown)


dmbates (commented):
Seems like a good idea to create a `ranef(m, :gf)` method.  Let me take a look and get back to you in the discussion of this PR.  (And thanks for adding the update to `NEWS.md` which I always forget to do.)

dmbates (approved):
I think that writing a `ranef(m, :gf)` method is a bit more than I want to take on right now.  It is just finicky enough that it would start to be a time sink.  I should probably concentrate on notebooks at this time.

palday:
>  I think that writing a `ranef(m, :gf)` method is a bit more than I want to take on right now. It is just finicky enough that it would start to be a time sink.

That's exactly why I didn't just do it myself. :smile:

kliegl:
> @dmbates actually, while I've paused releasing 4.1, should we add a `ranef(model, fname)` method to complement the `condVar` method?

I will check it out. Thanks so much!

PR #568, opened 2021-09-16
Add columntable method for OptSummary

dmbates:


src/optsummary.jl
```diff
@@ -90,6 +90,34 @@ function OptSummary(
     )
 end
 
+"""
+    columntable(s::OptSummary, [stack::Bool=false])
+
+Return `s.fitlog` as a `Tables.columntable`.
+
+When `stack` is false (the default), there will be 3 columns in the result:
+- `iter`: the iteration number
```
palday, line 99: It's technically the sample number, i.e. `iteration / thin`

test/pls.jl
```diff
@@ -132,7 +132,21 @@ end
     # since we're caching the fits, we should get it back to being correctly fitted
     # we also take this opportunity to test fitlog
     @testset "fitlog" begin
+        thin = 1
+        refit!(fm1; REML=false, progress=false, thin)
         fitlog = fm1.optsum.fitlog
+        fitlogtbl = columntable(fm1.optsum)
+        @test length(fitlogtbl) == 3
+        @test keys(fitlogtbl) == (:iter, :objective, :θ)
+        @test length(first(fitlogtbl)) > 15   # can't be sure of exact length
```
palday, line 141: since `thin=1`, this should be just `feval`

src/optsummary.jl
dmbates: Changed

palday (approved):

test/pls.jl
dmbates: We'll defer on this one a bit because presently it is `feval + 1`

PR #574, opened 2021-09-24
Correct a typo

dmbates:
The name should be the abstract data type, not the package name, shouldn't it?


palday (approved):
Good catch, thanks!

dmbates:
Well you know how obsessive I am about NEWS entries being present and correct, right?

PR #578, opened 2021-11-10
Better handling of constant response

palday:
For GLMM, this means allow constructing a model with a constant response.
This is useful for simulation.
For both GLMM and LMM, catch an initial PosDefException.
If the PosDefException is from a constant response, throw ArgumentError.

Technically this is a "feature" so I've done a minor version bump instead of patch.


test/pls.jl
```diff
@@ -181,6 +181,7 @@ end
     refit!(fm, float(MixedModels.dataset(:dyestuff2)[:yield]); progress=false) # restore the model in the cache
 end
 
+
```
palday, line 184:
```suggestion
```

palday:
The drop in coverage is from the `rethrow()` lines when the `PosDefException` occurs without a constant response. I don't have a good pathological example for eliciting that handy.

dmbates (approved):

Well, except for that test error thing.

palday:
@dmbates Yeah, it looks like subtle linear algebra differences lead to GLMM failing at an unexpected point for GLMM, so I've just moved to always testing for constant response in `fit!(::GeneralizedLinearMixedModel)`. The test with `any` should exit pretty quickly for most datasets where the response is not constant; the big exception would be a dataset that had been sorted by the response ....

palday:
(The test should still be faster than trying to add error trapping into the objective evaluation.)

PR #603, opened 2022-04-07
rework `predict` docstring

palday:
Thanks for contributing!

- [x] add entry in NEWS.md

No behavioral changes, but we updated a docstring enough that a patch release might be worthwhile.

@kliegl is the new explanation clearer?

### DIFF
--- a/src/predict.jl
+++ b/src/predict.jl
@@ -47,6 +47,11 @@ while including previously observed levels of the other grouping variables.
 In the future, it may be possible to specify a subset of the grouping variables
 or overall random-effects structure to use, but not at this time.
 
+!!! note
+    `new_re_levels` impacts only the behavior for previously unobserved random
+    effects levels, i.e. new RE levels. For previously observed random effects
+    levels, predictions take both the fixed and random effects into account.
+
 For `GeneralizedLinearMixedModel`, the `type` parameter specifies
 whether the predictions should be returned on the scale of linear predictor
 (`:linpred`) or on the response scale (`:response`). If you don't know the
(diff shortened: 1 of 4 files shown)


src/predict.jl
```diff
@@ -26,18 +26,26 @@ Predict response for new data.
...
-`:population` (return population values, i.e. fixed-effects only),
-`:missing` (return `missing`), `:error` (error on this condition;
-the error type is an implementation detail). If you want simulated values
-for unobserved levels of the grouping variable, consider the
-[`simulate!`](@ref) and `simulate` methods.
+values of the grouping variable are handled. Possible values are:
+
+- `:population`: return population values for the relevant grouping variable.
+   In other words, treat the associated random effect as 0.
+   If all grouping variables have new levels, then this is equivalent to the
```
ararslan, line 33:
```suggestion
   If all grouping variables have new levels, then this is equivalent to
```
Otherwise it's "the just the"

src/predict.jl
```diff
@@ -26,18 +26,26 @@ Predict response for new data.
...
+   just the fixed effects.
+- `:missing`: return `missing`.
+- `:error`: error on this condition. The error type is an implementation detail:
+   you should not rely on a particular type of error being returned.
```
ararslan, line 37:
```suggestion
   you should not rely on a particular type of error being thrown.
```
?

ararslan (commented):
Two minor things, otherwise LGTM.

palday:
@ararslan as someone who uses MixedModels but didn't write the code in question, do the descriptions of the options for `new_re_levels` make sense ?

kliegl:
@palday  How about including a sentence:   Random effects (not only fixed effects) are included for predictions involving grouping variables with observed levels.

ararslan (approved):

dmbates (approved):

PR #612, opened 2022-05-03
Stop exporting removed `describeblocks`

nalimilan:
This makes StatsKit tests fail when checking that no exported function conflicts with other packages it reexports.


palday (approved):
woops, looks like we forgot to remove the export when we fully removed even the `@deprecate` bit.

PR #615, opened 2022-05-18
attempt rescaling of poorly scaled problems

### LINKED ISSUE #611: LoadError: PosDefException: matrix is not positive definite; Cholesky factorization failed
We have been using an older version of MixedModels (1.1.6) for a while and decided to upgrade to the latest 4.6.1.  We ran into the subject errors with some of the models that used to run, though.  One example with data is attached here.  I did some experiment and it seems to be related to the fact that our model weights are highly correlated with the dependent variable.  But this model ran without issue in the older version.  I wonder if some tolerance changed since the older version.  BTW, I was using Julia 1.1.0 before and attempted to upgrade it to 1.6.5 going with the latest MixedModels.

model setup is as follows (VERSION refers to julia version)- 

```julia
using DataFrames, MixedModels, CSV
if Int(VERSION.minor) >= 6
  using CategoricalArrays
end
model_formula = @formula(
    volume ~ 1 + base_price + ppi + 
    (1 | brand_category_channel_type) + ((0 + ppi) | brand_category_channel_type) +
    (1 | channel_type_sub_brand) + ((0 + base_price) | channel_type_sub_brand)
)
if Int(VERSION.minor) < 6
    model_ready_data = CSV.read("model_ready_data_small.csv")
    weights = CSV.read("weights.csv")
else
    model_ready_data = CSV.read("model_ready_data_small.csv", DataFrame)
    weights = CSV.read("weights.csv", DataFrame)
end
weight_vector = Float64.(weights[!, :weight_vector])
if Int(VERSION.minor) < 6
    model = MixedModels.LinearMixedModel(model_formula, model_ready_data, weights=weight_vector);
else
    model_ready_data[!, "brand_category_channel_type"] = categorical(model_ready_data[!, "brand_category_channel_type"])
    model_ready_data[!, "channel_type_sub_brand"] = categorical(model_ready_data[!, "channel_type_sub_brand"])
    model = MixedModels.LinearMixedModel(model_formula, model_ready_data, wts = weight_vector)
end
model_fit = MixedModels.fit!(model);
```

[model_ready_data_small.csv](https://github.com/JuliaStats/MixedModels.jl/files/8606358/model_ready_data_small.csv)
[weights.csv](https://github.com/JuliaStats/MixedModels.jl/files/8606359/weights.csv)

palday:
Closes #611

This should help address some problematic models. The basic idea is to rescale the initial identity scaling by dividing by a large number, where "large" is the product of the largest weight by the largest response value. In other words, for poorly constrained problems, we change the initial guess to already have a good amount of shrinkage and see if that helps.

I tried more naive ways of rescaling the weights or even standardizing all variables in the model, but that didn't help. @dmbates if you're happy with this solution, I can add tests and updatee NEWs, etc., but I have rather mixed feelings about it myself.

- [x] add tests
- [x] add entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+MixedModels v4.6.5 Release Notes
+========================
+* Attempt recovery when the initial parameter values lead to an invalid covariance matrix by rescaling [#615]
+* Return `finitial` when the optimizer drifts into a portion of the parameter space that yields a (numerically) invalid covariance matrix [#615]
+
+
 MixedModels v4.6.4 Release Notes
 ========================
 * Support transformed responses in `predict` [#614]
@@ -347,3 +353,4 @@ Package dependencies
 [#604]: https://github.com/JuliaStats/MixedModels.jl/issues/604
 [#608]: https://github.com/JuliaStats/MixedModels.jl/issues/608
 [#614]: https://github.com/JuliaStats/MixedModels.jl/issues/614
+[#615]: https://github.com/JuliaStats/MixedModels.jl/issues/615
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MixedModels"
 uuid = "ff71e718-51f3-5ec2-a782-8ffcbfa3c316"
 author = ["Phillip Alday <me@phillipalday.com>", "Douglas Bates <dmbates@gmail.com>", "Jose Bayoan Santiago Calderon <jbs3hp@virginia.edu>"]
-version = "4.6.4"
+version = "4.6.5"
 
 [deps]
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
--- a/src/linearmixedmodel.jl
+++ b/src/linearmixedmodel.jl
@@ -428,6 +428,11 @@ function fit!(
     if optsum.feval > 0
         throw(ArgumentError("This model has already been fitted. Use refit!() instead."))
     end
+    if all(==(first(m.y)), m.y)
+        throw(
+            ArgumentError("The response is constant and thus model fitting has failed")
+        )
+    end
     opt = Opt(optsum)
     optsum.REML = REML
     prog = ProgressUnknown("Minimizing"; showspeed=true)
@@ -440,7 +445,7 @@ function fit!(
             objective(updateL!(setθ!(m, x)))
         catch ex
             ex isa PosDefException || rethrow()
-            iter == 1 && rethrow()
+            iter == 0 && rethrow()
             m.optsum.finitial
         end
         iszero(rem(iter, thin)) && push!(fitlog, (copy(x), val))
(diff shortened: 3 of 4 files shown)


dmbates:
Is it enough to change the `initial` field or does the `initial_step` field of the `OptSummary` object?

palday:
@dmbates The block it fails at is:

```julia
4×4 Matrix{Float64}:
     -2.09724e7  34383.4    29916.1    -78929.7
 -34369.8        -1334.4   -10224.6        -1.22374e5
 -29893.4        10223.1    -3655.35  -163610.0
  78803.2        -5724.27   -3884.24        2.10496e6
```

But I hadn't pursued it beyond that.

palday:
> Is it enough to change the `initial` field or does the `initial_step` field of the `OptSummary` object also need to be adjusted?

The initial step is just an empty vector -- we let NLopt do the hard work figuring that out. I have given some thought to adjusting the initial step for GLMM when we're initializing from a prior LMM/GLM but I hadn't really had a chance to experiment with what would be useful.

dmbates (approved):

dmbates:
Is it possible to construct a test that will reliably trigger this condition?

palday:
@dmbates I came up with a test based on an error we occasionally get in shrinkage plots -- there's a limit to how much you can re-inflate data that requires some amount of shrinkage to fit. 😄 

I also added in a secondary check that returns `finitial` when the optimizer drifts into an area where there isn't enough shrinkage. Why `finitial`? Generally, it will be the (near) worst case scenario value, so the optimizer won't view it as an optimum and using `Inf` messes up the quadratic approximation in BOBYQA. I couldn't add a reliable test in for this, because it's particularly sensitive to the exact parameter values the optimizer examines in a given step, but it did actually recover from an error I got when doing `refit!(simulate!(MersenneTwister(42), model))` for the model in #611. Are you still okay with all this recovery code?

palday:
@dmbates well having tested this on OpenBLAS+MKL on x86-64 and OpenBLAS on Apple silicon, I've found one example that works for now (using `fulldummy` so that shrinkage is absolutely required to fit). We may have to ditch this test in the future.

dmbates:
@palday This looks ready to go from my point of view.

dmbates:
@palday We probably should document in the code the reason for setting the value to `m.optsum.finitial` when evaluation of the objective throws an error.

palday:
@dmbates At first I thought that the value shouldn't be recorded twice because the line before that is `empty!(fitlog)`. But now I wonder if the doubled recording is the result of NLopt evaluating the first step after we've already evaluated it explicitly. I'll drop the line.

The longterm idea was to unconditionally store the first and last steps in `fitlog` and have the final and initial fields in optsum point to the first and last entries of the fit log.

palday:
@dmbates just double checked and the first entry isn't repeated for me with the current code.

dmbates:
@palday I think Nlopt has to evaluate the objective at the initial parameter values because we only pass it the parameters, we don't pass it the objective value.

If we want to have fitlog contain the initial and final parameter and objective values and have the `.initial`, `.final` fields just point to those values we should expand the schema for fitlog to include the evaluation number (and increment it at the beginning of the `obj` function, not in the middle as is done now).  Either that or abandon the idea of using `thin` and just incorporate a Boolean `logevaluations` or some name like that.  I doubt very much that many users will choose values of `thin` other than the default, `typemax(Int)` or `1`.  It is a symptom of my tendency to over-complicate matters that we use `thin` instead of a Boolean.

dmbates:
> @dmbates just double checked and the first entry isn't repeated for me with the current code.

if you set `thin = 1` it is
```julia
julia> ds = MixedModels.dataset(:dyestuff)
Arrow.Table with 30 rows, 2 columns, and schema:
 :batch  String
 :yield  Int16

julia> contrasts = Dict{Symbol, AbstractContrasts}(:batch => Grouping())
Dict{Symbol, AbstractContrasts} with 1 entry:
  :batch => Grouping()

julia> thin = 1
1

julia> dsm01 = let
           form = @formula yield ~ 1 + (1|batch)
           fit(MixedModel, form, ds; contrasts, thin)
       end
 yield ~ 1 + (1 | batch)
   logLik   -2 logLik     AIC       AICc        BIC    
  -163.6635   327.3271   333.3271   334.2501   337.5307

Variance components:
            Column    Variance Std.Dev.
batch    (Intercept)  1388.3332 37.2603
Residual              2451.2501 49.5101

  Fixed-effects parameters:
(Intercept)  1527.5     17.6946  86.33    <1e-99

julia> first(dsm01.optsum.fitlog, 3)
3-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.0], 327.76702162461663)
 ([1.0], 327.76702162461663)
 ([1.75], 331.03619322245146)
```

palday:
@dmbates Huh, that's weird. I think our tests don't catch that because they're doing `refit!` instead of `fit`, but ideally the results should be the same.

PR #639, opened 2022-09-21
Profile the objective

dmbates:
- The `profile` method for `LinearMixedModel` currently profiles  w.r.t. `σ`, the fixed-effects coefficients, the `σ` parameters, and the `θ` parameters (but not the correlation parameters).
- The `MixedModelProfile` struct returned contains a table, `tbl`, of values of ζ (the signed square root of the change in objective from the optimum) and  β  and θ as a `Vector{NamedTuple}`.  The `TypedTables.Table` generic is re-exported from this package to allow for presenting this row-table in a table format.
- Forward and reverse interpolation splines are added for parameters to and from ζ

- [x] add entry in NEWS.md


dmbates:
I need to rewrite the destructuring syntax in the old, tedious syntax if we are to continue to test on the LTS version.

palday:
> I need to rewrite the destructuring syntax in the old, tedious syntax if we are to continue to test on the LTS version.

@dmbates add a Compat.jl dependency and use `@compat` -- I like the new syntax, but I would still like to support LTS as long a possible.

Project.toml
```diff
@@ -24,9 +26,11 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 StatsModels = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
 StructTypes = "856f2bd8-1eba-4b0a-8007-ebc267875bd4"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
+TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
```
palday, line 30: BSplineKit and TypedTables need compat entries :)

dmbates:
`BSplineKit v0.12.0` requires julia v1.8    To be able to test on Julia LTS I allowed v0.11 of `BSplineKit`.  This means that after contour interpolation is available, the tests will need to be restricted to julia v1.8 or later.  Contour interpolation uses periodic splines, which became available in `BSplineKit` v0.12

palday:
I would prefer to keep LTS compat ... I can see how hard it would be to get BSplineKit to support 1.6 (and whether the author is amenable to that....)

dmbates:
If you have time to look at that, sure.  I expect that the periodic splines, which were introduced in BSplineKit v0.12.0, will be needed in MixedModelsMakie but probably not in MixedModels.

The way I imagine the division of responsibility is that MixedModels will add the forward and backward splines to the `MixedModelProfile`, because they are used to obtain profile-based confidence intervals.  These splines are natural cubic splines.  The reverse splines, mapping the zeta values to the fixed-effects coefficients, are used to create these profile-based confidence intervals.  The periodic splines are used for contour interpolation, which should be in MixedModelsMakie, I think.

That is, we can probably live with a compat entry in MixedModels.jl allowing 0.11,0.12 for BSplineKit.

palday:
@dmbates Okay, I won't be making a compat fix. The problem is a missing method `strides(::SparseMatrixCSC)`, which comes up in an ldiv operation. (https://github.com/JuliaLang/julia/pull/44027 added an appropriate method). Then we'll do your proposed compat fix and when it comes time to plotting the contours in MixedModelsMakie, we'll do a `@static if VERSION` and give an informative error for Julia < 1.8.

dmbates:
Sounds good.  Thanks for looking into that.

src/profile.jl
```diff
@@ -23,7 +23,7 @@ function Base.copy(ret::ReMat{T,S}) where {T,S}
         copy(ret.λ),
         ret.inds,
         ret.adjA,
-        copy(ret.scratch))
+        copy(ret.scratch)) # likely don't need to copy scratch, as it is temporary storage.
```
palday, line 26: I copied it to guarantee that we would never have a race condition: we're not currently doing any relevant threading on these things, but I could see that happening one day

dmbates: Seems reasonable.  Thanks for responding.

dmbates:
@palday My plan is to make the show method for a `MixedModelProfile` just show the `tbl` field.

I like the way the Table is set up now.  I still need to work on profiling the theta parameters or the vc's and cp's as I guess we will take to calling them.  One way to profile the theta parameters is to change the definition of the objective function in the optimization to fold the value of the parameter being profiled back in to the theta vector.  For the time being I think I will write a new function to do the conditional optimization and later we can see if we want to modify the one in linearmixedmodels.jl to allow for theta with one of its components fixed.

dmbates:
@palday Do you have any insight into the test failures? As far as I can see I haven't modified anything that should cause that test to fail?  It may be a Heisenbug if you look at the index range reported in the error.  I'm not altogether sure that the formula makes sense for the GLM used to initialize the estimates of the fixed-effects.  The GLM is fit during the construction of the GeneralizedLinearMixedModel.

palday:
I've restarted CI to see if it was a one-off.

For posteriority, the test failure was:
```julia
previous formula misspecification: Error During Test at /home/runner/work/MixedModels.jl/MixedModels.jl/test/misc.jl:14
  Test threw exception
  Expression: MixedModel(#= /home/runner/work/MixedModels.jl/MixedModels.jl/test/misc.jl:14 =# @formula(yield ~ 0 + (1 | batch)), dyestuff, Poisson()) isa GeneralizedLinearMixedModel
  BoundsError: attempt to access 0×0 Matrix{Float64} at index [1:140628461236480, 1:140628461236480]
```

If it turns out to be legitimate failure, I have a few more ideas.

palday:
@dmbates I bet this is from one of the recent changes in GLM.jl .... I have an idea how to fix and will ping you soon 🙂

dmbates:
> @dmbates I bet this is from one of the recent changes in GLM.jl .... I have an idea how to fix and will ping you soon slightly_smiling_face

@palday That was my guess too.  Has GLM now changed to using the QR decomposition rather than the Cholesky?

palday:
@dmbates recent changes in GLM that might impact this:
- better tolerance of some floating point weirdness in binomial GLMs (1.8.1, most recent)
- change in definition of how some quantities are computed for null models, including models without an intercept

Wait, I just found a better suspect, which snuck into the [1.8.1 patch release](https://github.com/JuliaStats/GLM.jl/releases/tag/v1.8.1) even though it should have been it's own 1.9 release

```
Implementation of dropcollinear feature in GeneralizedLinearModel (#488) (@mousum-github)
``` 


For my own reference, it looks like there are some BLAS differences between the CI runners and BLAS differences + the above things are often enough to create Heisenbugs around some edge cases.

Failed nightly:
```julia
Julia Version 1.9.0-DEV.1657
Commit 35d12890aba (2022-10-25 07:47 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 2 × Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, haswell)
  Threads: 1 on 2 virtual cores
```

Failed LTS:
```julia
Julia Version 1.6.7
Commit 3b76b25b64 (2022-07-19 15:11 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) Platinum 8171M CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake-avx512)
```

Passed CI
```julia
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 2 × Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, icelake-server)
```

Failed CI
```
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 2 × Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, skylake-avx512)
  Threads: 1 on 2 virtual cores
```

dmbates:
@palday I am leaning towards profiling the standard deviations of the random effects separately from the correlation parameters.  I think your introduction of the capability to set a value of sigma in the optsum will allow us to evaluate the profiled objective with respect to the standard deviation of a random effect without a lot of new code. 

We just redefine the objective function as
```julia
    function obj(x, g)
        isempty(g) || throw(ArgumentError("g must be empty"))
        updateL!(setθ!(m, x))
        m.optsum.sigma = vc[] / norm(view(m.reterms[i].λ.data, j, :))
        return objective(m)
    end
```
for the value of the standard deviation in `vc[]` (`vc` has to be a Ref in this formulation) of the `j`th component of the random effects associated with the `i`th grouping factor in the model `m`.

This may not be the most effective way of doing the conditional optimization but it is very low impact in terms of writing new code.

I'll work on fleshing this idea out a bit more when I return from my walk.

src/profile.jl
```diff
@@ -0,0 +1,170 @@
...
+        betaprofile!(prj, val, bb, j, objective) < threshold || break
+        bb += st
+    end
+    return sort!(Table(merge((; par = fill(Symbol("β$j"), length(val.σ))), val)), by = r -> r.ζ)
+end
+
+struct MixedModelProfile
+    tbl::Table                       # Table containing ζ, σ, β, and θ from each conditional fit
+    fecnames::Vector{String}         # Fixed-effects coefficient names
+    facnames::Vector{Symbol}         # Names of grouping factors
```
palday, line 78: the other two names are vectors (of vectors) of strings

palday: Since this is for us, let's keep it easy.

But we should give this a second thought for consistency with `coefnames` etc. when it comes time for the next breaking release.

dmbates:
@palday I have been thinking about the fields of the MixedModelProfile type.  I feel it would be best to keep a shallow copy of the LinearMixedModel object itself in the profile object.  That is, store the read-only fields that define the model and allocate copies of the fields in optsum and in reterms that are updated by setθ!.   Then we won't need to store the various coefficient names and grouping factor names and the parmap in the MixedModelProfile.

One advantage of this is that after fitting a model and profiling the fit we only need to store the profile because we can regenerate the model.

Does this seem like a reasonable approach?

palday:
> Although L does get modified by updateL! I don't think we should copy that whole field, which can be very large. We just need to enforce the convention that evaluating any properties of the model should only be done after updateL! unless you are sure that the contents of L are up-to-date.

I think this is okay if we are very clear in documentation about the shared reference and all the implications:
- not thread safe
- can leave a different object (the original model) in an inconsistent state if interrupted

dmbates:
@palday There are still some glitches in the examples, mostly due to scaling issues when choosing an initial step for the components of θ.  That will need more thought and better code.

I am working through the examples and trying out the `zetaplot` in the profile branch of `MixedModelsMakie`.  That plot needs decluttering.

dmbates:
There are many changes in commit 3cc7b53 that I would appreciate experiences with and comments upon @palday @kliegl.  I have switched to storing Tuples in most of the columns of the table and storing the values of the standard deviations of the random effects as part of the "sigma" column.  It turns out that for a table like this there isn't much of an advantage in using StaticArrays.SVectors rather than tuples.  I have some code to start profiling the random-effects standard deviations.  The problem is deciding how large a first step to take and I have a few ideas about that.

The confint method produces confidence intervals on all of the parameters that have been profiled.  Corner cases still need to be considered.  Nevertheless I am encouraged by this progress.

kliegl:
@dmbates Exciting!  Will take a look tomorrow.

kliegl:
@dmbates Using the MixedModels#profile branch, where do I find the final two plot functions?

```
using CairoMakie, MixedModels, MixedModelsMakie
CairoMakie.activate!(; type="svg")
# works
m1 = fit(MixedModel, @formula(yield ~ 1 + (1|batch)), MixedModels.dataset(:dyestuff))
pr01 = profile(m1);
confint(pr01)    # default is 95% coverage
# where do I find the following functions?
zetaplot(pr01)  # default intervals are at 50%, 80%, 90%, 95% and 99% coverage
profiledensity(pr01)  # the density functions that would produce these profiles
```

palday:
@kliegl you also need `MixedModelsMakie#profile` for plotting.

kliegl:
Works!  Thanks. I am set to try it out on some of my current work. Great.

src/MixedModels.jl
```diff
@@ -50,6 +53,8 @@ export @formula,
     LogLink,
     MixedModel,
     MixedModelBootstrap,
+    MixedModelProfile,
+    Not,
```
palday, line 57: why export `Not`?

src/profile.jl
```diff
@@ -0,0 +1,321 @@
...
+    for i in eachindex(sink, source, bd)
+        @inbounds sink[i] = max(source[i], bd[i] + incr)
+    end
+    return sink
+end
+
+function σvals(m::LinearMixedModel{T}) where {T}
+    @compat (; σ, reterms) = m
+    isone(length(reterms)) && return σvals(only(reterms), σ)
+    return (collect(Iterators.flatten(σvals.(reterms, σ)))...,)
```
palday, line 48: I'm not sure this is type stable -- the multiple grouping factor branch returns a tuple, but the other branch doesn't

src/profile.jl
```diff
@@ -0,0 +1,321 @@
...
+end
+
+function FeProfile(m::LinearMixedModel, j::Integer)
+    Xy = m.Xymat.xy
+    xcols = collect(axes(Xy, 2))
+    ycol = pop!(xcols)
+    notj = deleteat!(xcols, j) # indirectly check that j ∈ xcols
+    y₀ = Xy[:, ycol]
+    xⱼ = Xy[:, j]
+    feterm = FeTerm(Xy[:, notj], m.feterm.cnames[notj])
```
palday, line 58: would it be possible to use a view here?

src/profile.jl
```diff
@@ -0,0 +1,321 @@
...
+    rev::Dict{Symbol}     # Interpolation splines for a parameter as a function of ζ
+end
+
+"""
+    profile(m::LinearMixedModel)
+
+Return a `MixedModelProfile` for the objective of `m` with respect to the fixed-effects coefficients.
+"""
+function profile(m::LinearMixedModel{T}; threshold = 4) where {T}
+    final = copy(refit!(m).optsum.final)
```
palday, line 112: Is the `refit!` necessary?

dmbates: Just being extra-cautious while developing the code.  Past-me has a habit of leaving the model object in a corrupt state.

src/profile.jl
dmbates: I think I was copying those vectors and matrices out of an abundance of caution, because I didn't want to accidentally modify values in the original model structure.

Would it be okay to "stick a pin" in this and remind ourselves to follow up later on whether views can be used instead?

That reminds me that we may want to consider setting some of the fields in the larger structures as `const` (I think the change allowing that was in Julia v1.8.0 so we may need to put that inside a check on version numbers.)

src/MixedModels.jl
dmbates: I'm "Not" sure why I did that.  I'll revert it.

dmbates:
@palday @kliegl I have a Quarto document where I am developing the interpolated contour plots.  I find it easier to write scripts first then figure out how I should structure the functions.  Anyway, it is working.  If you want a preview edition I can send it to you however you want to receive it.  I probably won't make a repository because it is not really permanent but I can send you the .qmd file then all you need to do is create a quarto project directory, start Julia and add the necessary packages then render the document.

kliegl:
@dmbates I would like the .qmd file -- email is fine.

dmbates:
I forgot, it's Quarto so there is a built-in publishing mechanism.  See https://dmbates.quarto.pub/plotting-profiles/

dmbates:
Damn, trying to profile the standard deviations of the random effects is taking its toll on me.  I have been working for weeks on it and every time I think I have a good way of doing it, one or more of our examples blows up in my face.

I am going to commit some code that works for several examples but fails on rather simple examples like the dyestuff data.  This is just to avoid my cursing the universe if my computer fails and I lose all my work for the last couple of months.

dmbates:
On further examination it seems that the problems start when a value of zero for the theta parameter is passed to the objective function within `MixedModels.profilevc`, which returns `Inf`, and things go downhill from there.  Will try to find a workaround.

dmbates:
This PR is *finally* ready for others to take it out for a spin.

It adds a `profile` function that can be applied to a `LinearMixedModel`.  At present it profiles all the parameters except for the correlation parameters, returning a `MixedModelProfile` object.
This object contains the original model (hopefully with any changes made to values during the profiling process restored - but this would be a good thing to check), a row-table called `tbl` which contains the parameter values and zeta values from the profiling and two `Dict`s of natural, cubic interpolation splines.  The `fwd` Dict contains an interpolation spline of zeta versus the parameter being profiled, and the `rev` dict is the opposite direction.  These splines are created using the BSplineKit interpolate function.  Some of these splines wiggle a bit too much for my taste so I plan to look at the code in that package a bit more carefully.

The parameters should be subdivided into beta's, the fixed-effects parameters, theta's, elements of the relative covariance factor, and sigma's, the variance components - but on the standard deviation scale.  For now, `confint` applied to the profile produces confidence intervals on all of these parameters.  We should consider allowing for confidence intervals on only the beta's or the beta's and sigma's.  The theta's should probably be kept distinct from the other parameters.  They are used for optimization and we may want to investigate their contours, etc., but they are indirect and not of that much interest for inference by themselves.

At this point the rho parameters are evaluated for the other profiles but are not profiled themselves.  They are tricky - even trickier than the variance component parameters.  We will need to decide if the effort to write code to profile them is warranted.

There is a `profile` branch of the MixedModelsMakie package that provides `zetaplot` to plot the forward splines.  Again, this should probably have an argument to choose just beta's or just beta's and sigma's for the plot as those will be the parameters of greatest interest.  I'll be working on that code next.

kliegl:
Could one make an argument that profiling is "advantageous" for thetas, but bootstrapping is "good enough" for CPs?

dmbates:
@palday I merged the current main branch in the last commit.  If you have time to take a look and make sure I didn't botch things up I would appreciate it.  No rush.

palday:
@dmbates merge looks good 🎉

dmbates:
@palday Thanks for checking that.

dmbates:
@palday I am considering modifying the `confint` method for a `MixedModelProfile` to return the intervals for the fixed-effects and standard deviation parameters only, as is done in the `shortestcovint` method for a bootstrap.  I don't see the purpose of including profile-based intervals on the theta parameters but not a shortest coverage interval.  Does this make sense to you?

dmbates:
@palday Seems reasonable to add a `confint` method for a parametric bootstrap object.  I will do that.

dmbates:
@palday What would you think of bumping the Julia dependency for this branch to 1.8.  This branch depends on `BSplineKit` v0.14 for the periodic splines and also, I think, for extrapolation of natural splines beyond the limits of the nodes.  That version of BSplineKit depends on julia-1.8 or later.  As it stands this branch can't pass any CI tests for versions of Julia earlier than 1.8

palday:
Go ahead and bump it. If I get a few minutes, I might see what it would take to get BSplineKit to support 1.6, but there is enough good stuff in the last few releases that I suspect users should be on 1.8+ anyway.

dmbates:
@palday As I see it, this PR needs:

- tests  - I haven't looked at the coverage in detail but I assume it will need several additional tests
- docs - At least perfunctory documentation for all the profiling approaches should be added
- style enforcement - I should have been running JuliaFormatter before commits but I didn't always do that.  I presume this is something we can delay until we are close to a squash and merge.

palday:
@dmbates I've pulled in current main so that this branch is up to date with the other recent changes (including CI).

dmbates:
@palday Thanks for merging the changes in main.  It will help me when looking at code coverage.

src/MixedModels.jl
```diff
@@ -59,6 +63,7 @@ export @formula,
     ReMat,
     SeqDiffCoding,
     SqrtLink,
+    Table,
```
palday, line 66: do we need to export this?

src/bootstrap.jl
```diff
@@ -193,6 +193,22 @@ function allpars(bsamp::MixedModelFitCollection{T}) where {T}
     )
 end
 
+function StatsBase.confint(bsamp::MixedModelBootstrap{T}; level::Real=0.95) where {T}
+    cutoff = sqrt(quantile(Chisq(1), level))
+    # Creating the table is somewhat wasteful because columns are created then immediately skipped.
+    tbl = Table(bsamp.tbl)
```
palday, line 199: we do have a `shortestcovint(bsamp::MixedModelFitCollection{T}, level=0.95)` method that we could use and then filter

src/bootstrap.jl
```diff
@@ -193,6 +193,22 @@ function allpars(bsamp::MixedModelFitCollection{T}) where {T}
...
+    par = sort!(collect(filter(k -> !(startswith(string(k), 'θ') || string(k) == "obj"), propertynames(tbl))))
+    for p in par
+        l, u = shortestcovint(sort!(copyto!(v, getproperty(tbl, p))), level)
+        push!(lower, l)
+        push!(upper, u)
+    end
+    return DictTable(; par, lower, upper)
+end
+
 function Base.getproperty(bsamp::MixedModelFitCollection, s::Symbol)
```
palday, line 212: we should better document which properties are available somewhere

dmbates:
@palday Thanks for the comments.  I was just about to push another series of changes, which I will do, then go back to address your comments.

src/MixedModels.jl
dmbates: I would like to.  See issue #662.  I have made a change in this PR so that the `tbl` field of a `MixedModelProfile` is `Table` and not a `Vector{NamedTuple}`.  This means that many of the calls to `Table` that I had in code, just to format that table nicely in the display. are no longer needed.  But I would still like to have the `Table` constructor available.

src/bootstrap.jl
dmbates: It's quite possible that the logic in that section is more convoluted than necessary - a not uncommon problem for me.  Let me look at it afresh to see if I can figure out what I was thinking.

dmbates: Ah, I see what the issue is.  It has to do with the names of the parameters in the Table representation of the bootstrap results relative to the `allpars` representation.  In the allpars representation there are three columns whose combined values determine the parameter, which can be something like $\sigma$, `batch`, `(Intercept)` whereas in the Table (property name `tbl`) that would be just $\sigma1$.  Eventually I think the parameter names should just be symbols like $\beta1$, $\sigma1$, ... .  A separate extractor can return the mapping from symbol to parameter type, grouping factor, column designator.  Trying to embed all that information in each row of the .allpars table just created a mess.

Regarding using `shortestcovint` applied to the bootstrap sample to get the intervals, that goes through the .allpars construction, which is probably more long-winded than the .tbl construction.  It might be best to just add an optional argument to the .tbl construction so that it skips the $\theta$ and `obj` columns.

src/bootstrap.jl
dmbates: Absolutely.

dmbates:
@palday Any suggestions of what we should do with the many conversations initiated by the formatter bot?  We could adopt each one separately by clicking the "Commit suggestion" box but that seems like a lot of busy work.  (I suppose it is a sign of how spoiled we are that clicking a mouse button a couple of dozen times seems like "busy work".)

I think I would prefer to run `JuliaFormatter.format()` on a local copy then make one commit.  I'm not sure in that case what to do with all these proposed changes.  Would they just end up being flagged as resolved?

palday:
> I think I would prefer to run `JuliaFormatter.format()` on a local copy then make one commit. I'm not sure in that case what to do with all these proposed changes. Would they just end up being flagged as resolved?

Works for me and I think so!

dmbates:
@palday Seems to have worked.  I still have a couple of changes to commit.  When I created the `:tbl` property for the parametric bootstrap result I didn't include the correlation parameter estimates in the table so I am working on that now.

dmbates:
I am going to change commit `dc5de78` from returning a `NamedTuple` of `DictTable`s to return a `NamedTuple` of `Table`s.  I had thought that it would be good to use a `DictTable` for each grouping factor so that one could easily look up the entry from the level of the grouping factor.  But I found out that I usually want to sort the values and look at the large and small conditional modes.  It turn out that it is somewhat convoluted to get the values of a `DictTable` without the keys being attached.

src/MixedModels.jl
```diff
@@ -100,9 +105,12 @@ export @formula,
     model_response,
     nobs,
     objective,
+    objective!,
     parametricbootstrap,
     pirls!,
     predict,
+    profile,
+    profileσ,
```
palday, line 113:
```suggestion
    profileσ,
    profilevc,
```

src/linearmixedmodel.jl
```diff
@@ -796,6 +805,34 @@ function objective(m::LinearMixedModel{T}) where {T}
...
+"""
+    objective!(m::LinearMixedModel, θ)
+
+Equivalent to `objective(updateL!(setθ!(m, θ)))`.
+
+When `m` has a single, scalar random-effects term, `θ` can be a scalar
+
+Note that these methods modify `m`.
+The calling function is responsible for restoring the optimal `θ`.
+"""
```
palday, line 817:
```suggestion
"""
    objective!(m::LinearMixedModel, θ)
    objective!(m::LinearMixedModel)

Equivalent to `objective(updateL!(setθ!(m, θ)))`.

When `m` has a single, scalar random-effects term, `θ` can be a scalar

The one-argument method curries and returns a single-argument function of `θ`.

Note that these methods modify `m`.
The calling function is responsible for restoring the optimal `θ`.
"""
```

src/mixedmodel.jl
```diff
@@ -91,16 +91,16 @@ StatsAPI.nobs(m::MixedModel) = length(m.y)
...
+    return Table(
+    [NamedTuple{nms}((l, view(mat, :, i)...),) for (i, l) in enumerate(trm.levels)]
+)
 end
 
 """
     raneftables(m::MixedModel; uscale = false)
 
-Return the conditional means of the random effects as a `NamedTuple` of columntables
+Return the conditional means of the random effects as a `NamedTuple` of `DictTable`s
```
palday, line 103:
```suggestion
Return the conditional means of the random effects as a `NamedTuple` of Table.jl-compliant tables.

!!! note
    The API guarantee is only that there the NamedTuple contains Tables.jl tables and not on the particular concrete type of each table.
```

`DictTable` is technically a column table, so this is nonbreaking, but it's good to highlight the actual guarantee.

src/profile/fixefpr.jl
```diff
@@ -0,0 +1,107 @@
...
+    xcols = collect(axes(Xy, 2))
+    ycol = pop!(xcols)
+    notj = deleteat!(xcols, j) # indirectly check that j ∈ xcols
+    y₀ = Xy[:, ycol]
+    xⱼ = Xy[:, j]
+    feterm = FeTerm(Xy[:, notj], m.feterm.cnames[notj])
+    reterms = [copy(ret) for ret in m.reterms]
+    mnew = fit!(
+        LinearMixedModel(y₀ - xⱼ * m.β[j], feterm, reterms, m.formula); progress=false
+    )
```
palday, line 45: should the profiling functions take a progress argument and then when `progress=true`, should some `@info` updates and pass `progress` to the `fit!` calls?

src/profile/fixefpr.jl
```diff
@@ -0,0 +1,107 @@
...
+    # not sure this next call makes sense - should the second argument be m.optsum.final?
```
palday, line 46: I think starting from the original model makes sense

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
+"""
+    TableColumns
+
+A structure containing the column names for the numeric part of the profile table.
+
+The struct also contains a Dict giving the column ranges for Symbols like `:σ` and `:β`.
+Finally it contains a scratch vector used to accumulate to values in a row of the profile table.
+"""
+struct TableColumns{T<:AbstractFloat}
+    cnames::Tuple
```
palday, line 10: It might make sense to have

```suggestion
    cnames::Vector{Symbol}
```

We're already allocating a vector for constructing this struct and then we have only concrete, non-union types as fields. That means the struct itself can be allocated more efficiently.

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
...
+    if length(corrpos) > 0
+        ρvals!(view(v, positions[:ρs]), corrpos, m)
+        setθ!(m, view(v, positions[:θ]))
+    end
+    return NamedTuple{cnames,NTuple{length(cnames),T}}((v...,))
+end
+
+function parsej(sym::Symbol) # return the index from symbol names like :θ1, :θ01, etc.
+    symstr = string(sym)                                     # convert Symbol to a String
+    return parse(Int, SubString(symstr, nextind(symstr, 1))) # drop first Unicode character and parse as Int
```
palday, line 83: There's also [`graphemes`](https://docs.julialang.org/en/v1/stdlib/Unicode/#Unicode.graphemes), which if I recall correctly, appeared in 1.8. (The `compat` entry seems to be missing.)

src/profile/fixefpr.jl
```diff
@@ -0,0 +1,107 @@
...
+        mnew.optsum.initial, mnew.optsum.final, mnew.lowerbd; incr=0.05
+    )
+    return FeProfile(mnew, tc, y₀, xⱼ, j)
+end
+
+function betaprofile!(
+    pr::FeProfile{T}, tc::TableColumns{T}, βⱼ::T, j::Integer, obj::T, neg::Bool
+) where {T}
+    prm = pr.m
+    refit!(prm, mul!(copyto!(pr.m.y, pr.y₀), pr.xⱼ, βⱼ, -1, 1); progress=false)
```
palday, line 57:
```suggestion
    refit!(prm, mul!(copyto!(prm.y, pr.y₀), pr.xⱼ, βⱼ, -1, 1); progress=false)
```

src/profile/fixefpr.jl
```diff
@@ -0,0 +1,107 @@
...
+    (; cnames, v) = tc
+    pnm = (; p=sym)
+    j = parsej(sym)
+    prj = FeProfile(m, tc, j)
+    st = stderror[j] * 0.5
+    bb = β[j] - st
+    tbl = [merge(pnm, mkrow!(tc, m, zero(T)))]
+    while true
+        ζ = betaprofile!(prj, tc, bb, j, objective, true)
+        push!(tbl, merge(pnm, NamedTuple{cnames,NTuple{length(cnames),T}}((v...,))))
```
palday, line 85:
```suggestion
        push!(tbl, merge(pnm, NamedTuple{cnames,NTuple{length(cnames),T}}(tuple(v))))
```

I think also works and avoids the splatting penalty in models with lots of coefficients

src/profile/fixefpr.jl
```diff
@@ -0,0 +1,107 @@
...
+        if abs(ζ) > threshold
+            break
+        end
+        bb -= st
+    end
+    reverse!(tbl)
+    bb = β[j] + st
+    while true
```
palday, line 93: if we're not dependent on the order within `tbl`, you could probably make this into a single loop that goes from 0 to threshold and then has consecutive calls to `betaprofile!` with beta +/- step.

src/mixedmodel.jl
dmbates: Can you edit the suggestion to replace "Table.jl" with "Tables.jl" and, in the note, replace "only that there the" with "only that the"?

src/profile/profile.jl
```diff
@@ -0,0 +1,54 @@
...
+"""
+    profile(m::LinearMixedModel; threshold = 4)
+
+Return a `MixedModelProfile` for the objective of `m` with respect to the fixed-effects coefficients.
+
+Profiling starts at the parameter estimate and continues until reaching a parameter bound or the absolute
+value of ζ exceeds `threshold`.
+"""
+function profile(m::LinearMixedModel{T}; threshold=4) where {T}
+    final = copy(refit!(m).optsum.final)
```
palday, line 23: should we call `refit!` or should we document that it's expected that only fit models are passed?

src/profile/profile.jl
```diff
@@ -0,0 +1,54 @@
...
+    m.optsum.fmin = objective!(m, final)
+    for s in filter(s -> startswith(string(s), 'θ'), keys(first(val.tbl)))
+        profileθj!(val, s, tc; threshold)
+    end
+    profileσs!(val, tc)
+    objective!(m, final)   # restore the parameter estimates
+    updateL!(setθ!(m, final))
+    copyto!(m.optsum.final, final)
+    m.optsum.fmin = objective(m)
+    m.optsum.sigma = nothing
```
palday, line 40: are some of these calls redundant?

dmbates (commented):
The change `src/profile/utilities.jl` to store the `cnames` field  of the `TableColumns` struct as a `Vector{Symbol}` instead of as an `NTuple{N,Symbol}` seems to be causing problems downstream, because the `mkrow` method uses the `cnames` property as `NamedTuple{cnames,NTuple{length(cnames),T}}` so it needs to be converted to an `NTuple{N,Symbol}` anyway.  Better to do that once and store the result than in every call to `mkrow`, I think.

Okay if I back out this change, @palday ?

src/profile/sigmapr.jl
```diff
@@ -0,0 +1,67 @@
...
+end
+
+"""
+    _facsz(m, σ, objective)
+
+Return a factor such that refitting `m` with `σ` at its current value times this factor gives `ζ ≈ 0.5`
+"""
+function _facsz(m::LinearMixedModel{T}, σ::T, obj::T) where {T}
+    i64 = inv(64)
+    m.optsum.sigma = σ * exp(i64)
```
palday, line 23: I'm guessing that LLVM will generally catch that `exp(i64)` is const and generate ASM appropriately, but would it make sense to move that out of the multiplication?

src/profile/sigmapr.jl
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+    profileσ(m::LinearMixedModel, tc::TableColumns; threshold=4)
+
+Return a Table of the profile of `σ` for model `m`.  The profile extends to where the magnitude of ζ exceeds `threshold`.
+"""
+function profileσ(m::LinearMixedModel{T}, tc::TableColumns{T}; threshold=4) where {T}
+    (; σ, optsum) = m
+    isnothing(optsum.sigma) ||
+        throw(ArgumentError("Can't profile σ, which is fixed at $(m.optsum.sigma)"))
```
palday, line 35:
```suggestion
        throw(ArgumentError("Can't profile σ, which is fixed at $(optsum.sigma)"))
```

src/profile/sigmapr.jl
```diff
@@ -0,0 +1,67 @@
...
+    θ = copy(optsum.final)
+    θinitial = copy(optsum.initial)
+    _copy_away_from_lowerbd!(optsum.initial, optsum.final, optsum.lowerbd)
```
palday, line 38: I wonder if we should reduce the size of the initial step? The default initial step probably takes us quite far from the optimum, assuming that the optimum in the profiling steps is close-ish to the MLE.

src/profile/sigmapr.jl
```diff
@@ -0,0 +1,67 @@
...
+    return mkrow!(tc, m, (neg ? -one(T) : one(T)) * sqrt(m.objective - obj))
+end
+
+"""
+    _facsz(m, σ, objective)
+
+Return a factor such that refitting `m` with `σ` at its current value times this factor gives `ζ ≈ 0.5`
+"""
+function _facsz(m::LinearMixedModel{T}, σ::T, obj::T) where {T}
+    i64 = inv(64)
```
palday, line 22: also strictly speaking, we should cast this to `T`

src/profile/thetapr.jl
```diff
@@ -0,0 +1,103 @@
...
+    ζold = zero(T)
+    tbl = [merge(pnm, mkrow!(tc, m, ζold))]    # start with the row for ζ = 0
+    δj = inv(T(64))
+    θj = final[j]
+    θ[j] = θj - δj
+    while (abs(ζold) < threshold) && θ[j] ≥ lbj && length(tbl) < 100  # decreasing values of θ[j]
+        ζ = sign(θ[j] - θj) * sqrt(profileobj!(m, θ, opt, osj) - fmin)
+        push!(tbl, merge(pnm, mkrow!(tc, m, ζ)))
+        θ[j] == lbj && break
+        δj /= (4 * abs(ζ - ζold))   # take smaller steps when evaluating negative zeta
```
palday, line 61: why smaller steps for negative zeta?

dmbates: Because problems with lack of monotonicity in the reverse spline always occur with decreasing values of θ.  The value of  ζ flattens out as constrained components of θ approach zero.  Attempts to create the reverse spline often end up with wild swings up and down when trying to invert a pattern that has flattened.  This is a crude attempt to fix the situation by adding more points on the left hand side of the estimate.

palday:
> The change `src/profile/utilities.jl` to store the `cnames` field  of the `TableColumns` struct as a `Vector{Symbol}` instead of as an `NTuple{N,Symbol}` seems to be causing problems downstream, because the `mkrow` method uses the `cnames` property as `NamedTuple{cnames,NTuple{length(cnames),T}}` so it needs to be converted to an `NTuple{N,Symbol}` anyway.  Better to do that once and store the result than in every call to `mkrow`, I think.
> 
> Okay if I back out this change, @palday ?

@dmbates Yep, no worries

src/profile/thetapr.jl
palday: Oh right, you had mentioned that at some point.

src/mixedmodel.jl
palday: @dmbates done

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
...
+    sz = length(m.θ)
+    append!(nmvec, _generatesyms('θ', sz))
+    positions[:θ] = (lastpos + 1):(lastpos + sz)
+    return TableColumns((nmvec...,), positions, zeros(T, length(nmvec)), corrpos)
+end
+
+function mkrow!(tc::TableColumns{T,N}, m::LinearMixedModel{T}, ζ::T) where {T,N}
+    (; cnames, positions, v, corrpos) = tc
+    v[1] = ζ
+    MixedModels.fixef!(view(v, positions[:β]), m)
```
palday, line 70:
```suggestion
   fixef!(view(v, positions[:β]), m)
```

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
...
+    v[first(positions[:σ])] = m.σ
+    σvals!(view(v, positions[:σs]), m)
+    MixedModels.getθ!(view(v, positions[:θ]), m)  # must do this first to preserve a copy
```
palday, line 73:
```suggestion
    getθ!(view(v, positions[:θ]), m)  # must do this first to preserve a copy
```

src/profile/vcpr.jl
```diff
@@ -0,0 +1,74 @@
+
+function profilevc(m::LinearMixedModel{T}, val::T, rowj::AbstractVector{T}) where {T}
+    optsum = m.optsum
+    function obj(x, g)
+        isempty(g) || throw(ArgumentError("g must be empty"))
+        updateL!(setθ!(m, x))
+        optsum.sigma = val / norm(rowj)
```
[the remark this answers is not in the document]
dmbates: I don't think it can be removed from the closure because `norm(rowj)` depends on the value of `x` through the `setθ!(m, x)` call.

palday:
@dmbates I noticed the inconsistency when writing the CI comparison test:

https://github.com/JuliaStats/MixedModels.jl/blob/c16e6741fbf8bac99c05409ac1928c007337634c/test/bootstrap.jl#L158-L168

src/linearmixedmodel.jl
```diff
@@ -996,6 +1033,24 @@ function setθ!(m::LinearMixedModel{T}, θ::AbstractVector) where {T}
     return m
 end
 
+# This method is nearly identical to the previous one but determining a common signature
+# to collapse these to a single definition would be tricky, so we repeat ourselves.
+function setθ!(m::LinearMixedModel{T}, θ::NTuple{N,T}) where {T,N}
```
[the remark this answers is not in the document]
dmbates: Neither of those seem to work with julia v1.8.5

src/profile/profile.jl
dmbates: I think you are right that the call `updateL!(setθ!(m, final))` is redundant immediately after the call to `objective!(m, final)`

src/profile/sigmapr.jl
dmbates: I will add a coersion to `T`.

src/profile/sigmapr.jl
dmbates: I will move the evaluation of `exp` to a separate assignment, which should help.

dmbates: I checked with @code_warntype and the compiler does infer that `expi64` in the revised code is constant.

src/profile/profile.jl
```diff
@@ -0,0 +1,56 @@
+struct MixedModelProfile{T<:AbstractFloat}
```
palday, line 1:
```suggestion
"""
     MixedModelProfile{T<:AbstractFloat}
     
Type representing a likelihood profile of a [`LinearMixedModel`](@ref), including associated interpolation splines.

The function [`profile`](@ref) is used for computing profiles, while [`confint`](@ref) provides a useful method for constructing confidence intervals from a `MixedModelProfile`.

!!! note
   The exact fields and their representation are considered implementation details and are 
   **not** part of the public API.
"""
```

src/profile/profile.jl
```diff
@@ -0,0 +1,56 @@
...
+    profile(m::LinearMixedModel; threshold = 4)
+
+Return a `MixedModelProfile` for the objective of `m` with respect to the fixed-effects coefficients.
+
+`m` is `refit!` if `!isfitted(m)`.
+
+Profiling starts at the parameter estimate and continues until reaching a parameter bound or the absolute
+value of ζ exceeds `threshold`.
+"""
+function profile(m::LinearMixedModel{T}; threshold=4) where {T}
```
palday, line 24:
```suggestion
function profile(m::LinearMixedModel; threshold=4)
```

I don't think we use the type parameter anywhere in the method body

src/profile/sigmapr.jl
```diff
@@ -0,0 +1,68 @@
+"""
+    refitσ!(m::LinearMixedModel{T}, σ::T, tc::TableColumns{T}, obj::T, neg::Bool)
+
+Refit the model `m` with the given value of `σ` and return a NamedTuple of information about the fit.
+
+`obj` and `neg` allow for conversion of the objective to the `ζ` scale and `tc` is used to return a NamedTuple
+"""
```
palday, line 7:
```suggestion
!!! note
   This method is internal and may change or disappear in a future release
   without being considered breaking.

"""
```

src/profile/sigmapr.jl
```diff
@@ -0,0 +1,68 @@
...
+    expi64 = exp(i64)     # help the compiler infer it is a constant
+    m.optsum.sigma = σ * expi64
+    return exp(i64 / (2 * sqrt(refit!(m; progress=false).objective - obj)))
+end
+
+"""
+    profileσ(m::LinearMixedModel, tc::TableColumns; threshold=4)
+
+Return a Table of the profile of `σ` for model `m`.  The profile extends to where the magnitude of ζ exceeds `threshold`.
+"""
```
palday, line 32:
```suggestion
!!! note
   This method is called by `profile` and currently considered internal.
   As such, it may change or disappear in a future release without being considered breaking.
   
"""
```

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
+"""
+    TableColumns
+
+A structure containing the column names for the numeric part of the profile table.
+
+The struct also contains a Dict giving the column ranges for Symbols like `:σ` and `:β`.
+Finally it contains a scratch vector used to accumulate to values in a row of the profile table.
+"""
```
palday, line 8:
```suggestion

!!! note
   This is an internal structure used in [`MixedModelProfile`](@ref).
   As such, it may change or disappear in a future release without being considered breaking.

"""
```

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
...
+                normalize!(r)
+            end
+            lasti = i
+        end
+        v[ii] = dot(view(λ, j, :), view(λ, k, :))
+    end
+    return v
+end
+
+"""
```
palday, line 134: I feel like this could be useful in a few other places where we need to nudge things away from the boundary for refitting, but that's a follow-up for later

src/profile/vcpr.jl
```diff
@@ -0,0 +1,74 @@
...
+        objctv = objective(m)
+        return objctv
+    end
+    opt = Opt(optsum)
+    NLopt.min_objective!(opt, obj)
+    fmin, xmin, ret = NLopt.optimize!(opt, copyto!(optsum.final, optsum.initial))
+    _check_nlopt_return(ret)
+    return fmin, xmin
+end
+
```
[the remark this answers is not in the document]
palday: xref #689

src/profile/profile.jl
```diff
@@ -0,0 +1,56 @@
...
+    end
+    profileσs!(val, tc)
+    objective!(m, final)   # restore the parameter estimates
+    copyto!(m.optsum.final, final)
+    m.optsum.fmin = objective(m)
+    m.optsum.sigma = nothing
+    return MixedModelProfile(m, Table(val.tbl), val.fwd, val.rev)
+end
+
+function StatsBase.confint(pr::MixedModelProfile; level::Real=0.95)
```
palday, line 46:
```suggestion
"""
    confint(pr::MixedModelProfile; level::Real=0.95)
    
Compute profile confidence intervals for coefficients and variance components, with confidence level level (by default 95%).

!!! note
    The API guarantee is for a Tables.jl compatible table. The exact return type is an 
    implementation detail and may change in a future minor release without being considered
    breaking.
    
!!! note 
   The "row names" indicating the associated parameter name are guaranteed to be unambiguous,
   but their precise naming scheme is not yet stable and may change in a future release
   without being considered breaking.
"""
function StatsAPI.confint(pr::MixedModelProfile; level::Real=0.95)
```

src/linearmixedmodel.jl
```diff
@@ -359,12 +359,22 @@ function condVartables(m::MixedModel{T}) where {T}
     return NamedTuple{fnames(m)}((map(_cvtbl, condVar(m), m.reterms)...,))
 end
 
+function StatsBase.confint(m::MixedModel{T}; level=0.95) where {T}
```
palday, line 362:
```suggestion
"""
    confint(pr::MixedModelProfile; level::Real=0.95)
    
Compute profile confidence intervals for (fixed effects) coefficients, with confidence level `level` (by default 95%).

!!! note
    The API guarantee is for a Tables.jl compatible table. The exact return type is an 
    implementation detail and may change in a future minor release without being considered
    breaking.
  
"""
function StatsBase.confint(m::MixedModel{T}; level=0.95) where {T}
```

src/linearmixedmodel.jl
```diff
@@ -359,12 +359,22 @@ function condVartables(m::MixedModel{T}) where {T}
...
+    cutoff = sqrt.(quantile(Chisq(1), level))
+    β, std = m.β, m.stderror
+    return DictTable(;
+        coef=coefnames(m),
+        lower=β .- cutoff .* std,
+        upper=β .+ cutoff .* std
```
palday, line 368: what happens in rank deficiency?

src/profile/utilities.jl
```diff
@@ -0,0 +1,165 @@
...
+    σvals!(view(v, positions[:σs]), m)
+    getθ!(view(v, positions[:θ]), m)  # must do this first to preserve a copy
+    if length(corrpos) > 0
+        ρvals!(view(v, positions[:ρs]), corrpos, m)
+        setθ!(m, view(v, positions[:θ]))
+    end
+    return NamedTuple{cnames,NTuple{N,T}}((v...,))
+end
+
+function parsej(sym::Symbol) # return the index from symbol names like :θ1, :θ01, etc.
```
palday, line 81:
```suggestion
"""
    parsej(sym::Symbol)

Return the index from symbol names like `:θ1`, `:θ01`, etc.

!!! note
    This method is internal.
"""
function parsej(sym::Symbol)
```

src/bootstrap.jl
```diff
@@ -165,6 +165,28 @@ function allpars(bsamp::MixedModelFitCollection{T}) where {T}
     )
 end
 
+function StatsBase.confint(bsamp::MixedModelBootstrap{T}; level::Real=0.95) where {T}
```
palday, line 168:
```suggestion
"""
    confint(pr::MixedModelBootstrap; level::Real=0.95)
    
Compute bootstrap confidence intervals for coefficients and variance components, with confidence level level (by default 95%).

!!! note
    The API guarantee is for a Tables.jl compatible table. The exact return type is an 
    implementation detail and may change in a future minor release without being considered
    breaking.
    
!!! note 
   The "row names" indicating the associated parameter name are guaranteed to be unambiguous,
   but their precise naming scheme is not yet stable and may change in a future release
   without being considered breaking.

See also [`shortestcovint`](@ref).
"""
function StatsBase.confint(bsamp::MixedModelBootstrap{T}; level::Real=0.95) where {T}
```

src/profile/vcpr.jl
```diff
@@ -0,0 +1,74 @@
...
+function profilevc(m::LinearMixedModel{T}, val::T, rowj::AbstractVector{T}) where {T}
```
palday, line 2:
```suggestion
"""
     profilevc(m::LinearMixedModel{T}, val::T, rowj::AbstractVector{T}) where {T}

Profile an element of the variance components.

!!! note
   This method is called by `profile` and currently considered internal.
   As such, it may change or disappear in a future release without being considered breaking.
"""
function profilevc(m::LinearMixedModel{T}, val::T, rowj::AbstractVector{T}) where {T}
```

src/profile/vcpr.jl
```diff
@@ -0,0 +1,74 @@
...
+        return objctv
+    end
+    opt = Opt(optsum)
+    NLopt.min_objective!(opt, obj)
+    fmin, xmin, ret = NLopt.optimize!(opt, copyto!(optsum.final, optsum.initial))
+    _check_nlopt_return(ret)
+    return fmin, xmin
+end
+
+function profileσs!(val::NamedTuple, tc::TableColumns{T}; nzlb=1.0e-8) where {T}
```
palday, line 18:
```suggestion
"""
     profileσs!(val::NamedTuple, tc::TableColumns{T}; nzlb=1.0e-8) where {T}
     
Profile the variance components.

!!! note
   This method is called by `profile` and currently considered internal.
   As such, it may change or disappear in a future release without being considered breaking.
"""
function profileσs!(val::NamedTuple, tc::TableColumns{T}; nzlb=1.0e-8) where {T}
```

test/pls.jl
```diff
@@ -501,6 +515,25 @@ end
         @test bic(fm) ≈ bic(m)
         @test coef(fm) ≈ coef(m)
     end
+    @testset "profile" begin
+        pr = profile(last(models(:sleepstudy)))
```
palday, line 519: I got a warning here:

```
┌ Warning: Reverse spline for θ3 is not monotone.
```

test/pls.jl
```diff
@@ -501,6 +515,25 @@ end
...
+        tbl = pr.tbl
+        @test length(tbl) >= 122
+        ci = confint(pr)
+        @test isa(ci, TypedTables.DictTable)
+        @test propertynames(ci) == (:par, :estimate, :lower, :upper)
+        @test collect(ci.par) == [:β1, :β2, :σ, :σ1, :σ2]
+#        @test isapprox(   # these values are not reproducible
+#            ci.lower.values,
+#            [237.68069406557657, 7.3586529384660375, 0.5413630480965158, -0.13854853868801975, 0.11978197717507284, 22.898262163628825];
+#            atol=1.e-3)
```
palday, line 529:
```suggestion
        @test isapprox(
            ci.lower.values,
            [237.681, 7.359, 22.898, 14.381, 0.0];
            atol=1.e-3)
        @test isapprox(
            ci.upper.values,
            [265.130, 13.576, 28.858, 37.718. 8.753];
            atol=1.e-3)
```

test/pls.jl
```diff
@@ -501,6 +515,25 @@ end
         @test bic(fm) ≈ bic(m)
         @test coef(fm) ≈ coef(m)
     end
+    @testset "profile" begin
+        pr = profile(last(models(:sleepstudy)))
+        tbl = pr.tbl
+        @test length(tbl) >= 122
+        ci = confint(pr)
+        @test isa(ci, TypedTables.DictTable)
```
palday, line 523:
```suggestion
        @test Tables.istable(ci)
```

test/pls.jl
```diff
@@ -501,6 +515,25 @@ end
...
+        @test collect(ci.par) == [:β1, :β2, :σ, :σ1, :σ2]
+#        @test isapprox(   # these values are not reproducible
+#            ci.lower.values,
+#            [237.68069406557657, 7.3586529384660375, 0.5413630480965158, -0.13854853868801975, 0.11978197717507284, 22.898262163628825];
+#            atol=1.e-3)
+        @test first(only(filter(r -> r.p == :σ && iszero(r.ζ), pr.tbl)).σ) == last(models(:sleepstudy)).σ
+    end
+    @testset "confint" begin
+        ci = confint(last(models(:sleepstudy)))
+        @test isa(ci, TypedTables.DictTable)
```
palday, line 534:
```suggestion
        @test Tables.istable(ci)
```

test/pls.jl
```diff
@@ -525,6 +558,15 @@ end
     tokens = Set(split(String(take!(io)), r"\s+"))
     @test "Corr." in tokens
     @test "-0.89" in tokens
+    @testset "profile" begin
+        contrasts = Dict(:item => Grouping(), :subj => Grouping(), :prec => EffectsCoding(; base="maintain"),
+        :spkr => EffectsCoding(), :load => EffectsCoding())
+        kbf03 = @formula rt_trunc ~ 1+prec+spkr+load+(1+prec|item)+(1|subj)
+        kbpr03 = profile(fit(MixedModel, kbf03, MixedModels.dataset(:kb07); contrasts, thin=1))
```
palday, line 565:
```suggestion
        kbpr03 = profile(fit(MixedModel, kbf03, MixedModels.dataset(:kb07); contrasts, thin=1, progress=false))
```

test/pls.jl
```diff
@@ -525,6 +558,15 @@ end
...
+        prtbl = Table(kbpr03.tbl)
+        @test length(propertynames(prtbl)) == 15
+        @test length(kbpr03.tbl) > 200
```
palday, line 568:
```suggestion
        @test ength(Tables.columnnames(kbpr03.tbl))
        @test length(Tables.rows(kbpr03.tbl)) > 200
```

palday (approved):
I think this is mostly good to go for an initial merge and release. 

- I've marked all comment-conversations as resolved that I think are nonblocking.
- I've added a number of docstring suggestions that also highlight where the API guarantees are and which things might still change in the 4.x series. There are three broad things that I think we might revisit:
    - The exact internal structure of `MixedModelProfile`, but I think that is not really interesting for most users, who just want the profile to compute the CI.
    - The concrete Tables.jl-compatible return types. I don't really have any objection to the current DictTable, but I suspect that we'll only really know what works as we get more experience with using this feature in practice.
    - The parameter names in the various tables. For the fixed effects, I would really like to see these match `coefnames` or `fixefnames`, but that is the type of polish that we can make in a smaller, more focused PR.
- Related to nonblocking polish, I've opened up a discussion for that: #688.
- I've added a NEWS entry that tries to highlight the "awesome new feature but some details might change!" nature of the release. Feel free to modify.
- I've bumped the version appropriately.

Feel free to keep or reject all my suggestions.

src/linearmixedmodel.jl
dmbates: I'm not sure.  Part of me just wants to tell users to ensure that their model matrices have full rank but that might be an unpopular suggestion.  I'll have to look up what happens with the `stderror` property in that case.

test/pls.jl
dmbates: That warning is expected.  It is tricky to ensure that the reverse spline is monotone.  I suggest leaving that in there so that we know we have an issue to address.

palday:
@dmbates I think I've messed up the indents on some of the `!!! note` blocks. When you're all done, I make those changes locally and push before you merge.

palday:
> @dmbates I think I've messed up the indents on some of the `!!! note` blocks. When you're all done, I make those changes locally and push before you merge.

fixed

PR #645, opened 2022-10-04
improve "API" docs

### LINKED ISSUE #644: document across public-facing dependencies
`MixedModels.jl` would be much more user-friendly and accessible to noobs such as myself if the docs included or cross-linked to dependencies that export relevant functions, ideally with a search functionality that spans all such packages.

I keep searching through the docs looking for things that one would expect, but not finding them because they are defined in dependencies of `MixedModels.jl` that I didn't even know exist.

palday:
hopefully fixes #644

- [x] use StatsAPI directly instead of via StatsBase
- [x] move to `using` and explicit overload method definitions instead of `import` 
- [x] have `RegressionModel` and `StatisticalModel` API be obvious in our docs

- [ ] add entry in NEWS.md


palday:
@dmbates I've changed from `import` to `using` with explicit annotation of 'owner' modules for functions we're extending with new methods. It makes things a little bit more verbose, but I think it makes it a little bit easier to keep track of when we're doing something new versus extending other definitions. Feel free to object. 😄

dmbates:
Looks good to me.  I am fine with having the qualified names to make it clearer where the generic is defined and documented.  Do you want a review or are you still working on this?

palday:
@dmbates I think I'm ready for review.

src/MixedModels.jl
```diff
@@ -15,23 +15,21 @@ using ProgressMeter
...
 using StatsModels: TableRegressionModel
-
 using StatsFuns: log2π, normccdf
 
-import Base: *
-import DataAPI: levels, refpool, refarray, refvalue
-import GLM: dispersion, dispersion_parameter
-import NLopt: Opt
-import StatsBase: fit, fit!
+import Base: ==
```
dmbates, line 32: Why is the import of `==` required?  I couldn't find a case of an `==` method being defined.

dmbates: I think that line is redundant.  I was able to comment it out and still pass the tests.

src/linalg/rankUpdate.jl
```diff
@@ -91,7 +91,7 @@ end
 
 function rankUpdate!(
     C::HermOrSym{T,Diagonal{T,Vector{T}}}, A::StridedMatrix{T}, α, β
-) where {T,S}
+) where {T}
```
dmbates, line 94: Good catch on this and the following change.

dmbates (approved):
Just that one change in `MixedModels.jl` and it looks good to go.

src/MixedModels.jl
palday: oops, I missed removing it.

palday:
```suggestion
```

src/linalg/rankUpdate.jl
palday: I received a warning on Julia 1.8.2 when I loaded MixedModels!

PR #648, opened 2022-10-05
Combine updateL! operations on diagonal blocks

### LINKED ISSUE #643: Amalgamate copyto! and scaleinflate! in updateL!
Right now the sequence in `updateL!` is to `copyto!` each block of `A` to `L` then to iterate over the `reterms` with the first operation being `scaleinflate!` of the diagonal block.  I think the logic will be cleaner if the `copyto!`  operation is combined with the `scaleinflate!` operation.  It certainly seems so when storing the diagonal block of `L` in RectangularFullPacked format.

dmbates:
closes #643 

- Previously the sequence of operations in `updateL!` was to copy blocks of `A` to the corresponding block of `L` then iterate over the random effects terms scaling the diagonal block and inflating the diagonal.
- This PR combines those operations on the diagonal blocks, replacing separate calls to `copyto!` and `scaleinflate!` by `copyscaleinflate!`.
- This reduces the number of times iterating over the diagonal elements, which may help with further enhancements such as using [RectangularFullPacked](https://github.com/JuliaLinearAlgebra/RectangularFullPacked.jl) storage for triangular blocks of `L`.
- Closes #643

- [x] add entry in NEWS.md

### DIFF
--- a/src/remat.jl
+++ b/src/remat.jl
@@ -572,9 +572,9 @@ function copyscaleinflate!(Ljj::Matrix{T}, Ajj::Diagonal{T}, Λj::ReMat{T,1}) wh
 end
 
 function copyscaleinflate!(
-    Ljj::UniformBlockDiagonal{T}, 
+    Ljj::UniformBlockDiagonal{T},
     Ajj::UniformBlockDiagonal{T},
-    Λj::ReMat{T,S},
+    Λj::ReMat{T,S}
 ) where {T,S}
     λ = Λj.λ
     dind = diagind(S, S)
@@ -592,7 +592,7 @@ end
 function copyscaleinflate!(
     Ljj::Matrix{T},
     Ajj::UniformBlockDiagonal{T},
-    Λj::ReMat{T,S},
+    Λj::ReMat{T,S}
 ) where {T,S}
     copyto!(Ljj, Ajj)
     n = LinearAlgebra.checksquare(Ljj)
--- a/src/remat.jl
+++ b/src/remat.jl
@@ -550,31 +550,35 @@ function rowlengths(A::ReMat)
 end
 
 """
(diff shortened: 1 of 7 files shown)


src/remat.jl
```diff
@@ -550,31 +550,35 @@ function rowlengths(A::ReMat)
...
 end
 
-function scaleinflate!(Ljj::Matrix{T}, Λj::ReMat{T,1}) where {T}
+function copyscaleinflate!(Ljj::Matrix{T}, Ajj::Diagonal{T}, Λj::ReMat{T,1}) where {T}
+    fill!(Ljj, zero(T))
     lambsq = abs2(only(Λj.λ.data))
-    @inbounds for i in diagind(Ljj)
-        Ljj[i] *= lambsq
-        Ljj[i] += one(T)
+    @inbounds for (i, a) in enumerate(Ajj.diag)
```
palday, line 568: since we only ever create `Diagonal{T, Vector{T}}` this shouldn't cause trouble, but strictly speaking, there could be `Diagonal{T, StarWarsVector{T}}` which would cause undefined behavior with `@inbounds`

palday (approved):

palday:
coverage suggests you could delete this method:

```julia
function Base.copyto!(dest::UniformBlockDiagonal{T}, src::UniformBlockDiagonal{T}) where {T}
    sdat = src.data
    ddat = dest.data
    size(ddat) == size(sdat) || throw(DimensionMismatch(""))
    copyto!(ddat, sdat)
    return dest
end
```

src/remat.jl
dmbates: Anyone who chooses to use a `StarWarsVector` in real life deserves undefined behavior. :smile:

PR #653, opened 2022-10-12
allow predicting from a single observation

### LINKED ISSUE #625: Predict: only one level found
I wanted to `predict` a few values, but it said

```jl
julia> MixedModels.predict(fitval, d)
ERROR: ArgumentError: only one level found: 2016 (need at least two to compute contrasts).
```

My goal is to compute the prediction for one level, so I don't want to pass two levels.

palday:
closes #625

- [x] add entry in NEWS.md

### DIFF
--- a/src/grouping.jl
+++ b/src/grouping.jl
@@ -29,8 +29,9 @@ struct Grouping <: StatsModels.AbstractContrasts end
 Base.getproperty(g::Grouping, prop::Symbol) = prop == :levels ? nothing : getfield(g, prop)
 
 # special-case categorical terms with Grouping contrasts.
-StatsModels.modelcols(::CategoricalTerm{Grouping}, d::NamedTuple) =
-    error("can't create model columns directly from a Grouping term")
+function StatsModels.modelcols(::CategoricalTerm{Grouping}, d::NamedTuple)
+    return error("can't create model columns directly from a Grouping term")
+end
 
 function StatsModels.ContrastsMatrix(
     contrasts::Grouping, levels::AbstractVector{T}
--- a/src/grouping.jl
+++ b/src/grouping.jl
@@ -34,7 +34,9 @@ function StatsModels.modelcols(::CategoricalTerm{Grouping}, d::NamedTuple)
 end
 
 # copied from StatsModels@463eb0a
-function StatsModels.ContrastsMatrix(contrasts::Grouping, levels::AbstractVector{T}) where {T}
+function StatsModels.ContrastsMatrix(
+    contrasts::Grouping, levels::AbstractVector{T}
+) where {T}
 
     # if levels are defined on contrasts, use those, validating that they line up.
     # what does that mean? either:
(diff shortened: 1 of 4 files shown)


src/grouping.jl
```diff
@@ -32,3 +32,43 @@ Base.getproperty(g::Grouping, prop::Symbol) = prop == :levels ? nothing : getfie
...
+        )
+    end
+
+    # do conversion AFTER checking for levels so users get a nice error message
+    # when they've made a mistake with the level types
+    c_levels = convert(Vector{T}, c_levels)
+
+    n = length(c_levels)
+    # not validating this allows for prediction of only a single level of the grouping factor
+
```
kleinschmidt, line 69:
```suggestion
```

src/grouping.jl
```diff
@@ -32,3 +32,43 @@ Base.getproperty(g::Grouping, prop::Symbol) = prop == :levels ? nothing : getfie
...
+    baseind = 1
+    tnames = StatsModels.termnames(contrasts, c_levels, baseind)
+    mat = StatsModels.contrasts_matrix(contrasts, baseind, n)
```
kleinschmidt, line 72: given that our methods for Grouping ignore all these we can just use nonsense values...

```suggestion
    mat = StatsModels.contrasts_matrix(contrasts, baseind, 0)
```

kleinschmidt (commented):
given that the Grouping struct itself doesn't keep track of the levels, I think all the validation stuff can just be skipped altogether.  also, would it break anything to just remove the termnames/contrasts_matrix methods altogether?  I think the only place they get used is in these constructors...

src/grouping.jl
```diff
@@ -32,3 +32,13 @@ Base.getproperty(g::Grouping, prop::Symbol) = prop == :levels ? nothing : getfie
...
 end
+
+# copied and then adapted from StatsModels@463eb0a
+function StatsModels.ContrastsMatrix(
+    contrasts::Grouping, levels::AbstractVector{T}
+) where {T}
+    baseind = 1
+    tnames = StatsModels.termnames(contrasts, c_levels, baseind)
+    mat = StatsModels.contrasts_matrix(contrasts, baseind, 0)
+    return StatsModels.ContrastsMatrix(mat, tnames, c_levels, contrasts)
```
kleinschmidt, line 43:
```suggestion
    tnames = StatsModels.termnames(contrasts, levels, baseind)
    mat = StatsModels.contrasts_matrix(contrasts, baseind, 0)
    return StatsModels.ContrastsMatrix(mat, tnames, levels, contrasts)
```

src/grouping.jl
```diff
@@ -22,8 +22,8 @@ julia> schema((; grp = string.(1:100_000)), Dict(:grp => Grouping()))
 struct Grouping <: StatsModels.AbstractContrasts end
 
 # return an empty matrix
-StatsModels.contrasts_matrix(::Grouping, baseind, n) = zeros(0, 0)
-StatsModels.termnames(::Grouping, levels::AbstractVector, baseind::Integer) = levels
+# StatsModels.contrasts_matrix(::Grouping, baseind, n) = error("Grouping terms don't have associated contrasts")
+# StatsModels.termnames(::Grouping, levels::AbstractVector, baseind::Integer) = levels
```
kleinschmidt, line 26: probably better to just clean up this cruft, OR leave the methods as they were before with a note that they're _probably_ not needed since we've added a direct constructor for ContrastsMatrix

src/grouping.jl
```diff
@@ -32,3 +32,9 @@ Base.getproperty(g::Grouping, prop::Symbol) = prop == :levels ? nothing : getfie
 function StatsModels.modelcols(::CategoricalTerm{Grouping}, d::NamedTuple)
     return error("can't create model columns directly from a Grouping term")
 end
+
+function StatsModels.ContrastsMatrix(
+    contrasts::Grouping, levels::AbstractVector{T}
+) where {T}
```
kleinschmidt, line 38: type param no longer actually used...

```suggestion
function StatsModels.ContrastsMatrix(
    contrasts::Grouping, levels::AbstractVector
)
```

kleinschmidt (approved):
love 2 delete code.

src/grouping.jl
palday: I left it in there to make sure we're more specific than the methods in StatsModels, but I'll test leaving it out.

kleinschmidt: I _think_ it's teh same (union type either way) but better to make sure.

PR #654, opened 2022-10-18
more tinkering on 638

palday:
I made it return the same result as lme4.


dmbates (approved):
Thanks for cleaning up those redundant `.data` extractors in `src/remat.jl`.

PR #660, opened 2022-12-20
patch bump

palday:
This is releasing a bug/compat fix for GLM.jl, which had breaking-ish behavior in its 1.8 release. This should also address why we're failing PkgEval.

- [x] add entry in NEWS.md


ararslan (approved):
🆗

PR #664, opened 2023-01-25
integrate changes from StatsModels 0.7 (upcoming release)

kleinschmidt:
WIP.  The major thing that needs to be addressed is how `FunctionTerm`s are represented (https://github.com/JuliaStats/StatsModels.jl/pull/183).

Don't merge until
- StatsModels 0.7 released
- GLM release with StatsModels 0.7 compat released https://github.com/JuliaStats/GLM.jl/pull/512


palday:
@kleinschmidt This is what's failing (it's a non sense model but it was chosen to show off the formula syntax on something quick to fit)

```julia
using MixedModels
sleepstudy = MixedModels.dataset(:sleepstudy)
fit(MixedModel, @formula(reaction ~ 1 + days + (1|subj) + zerocorr(days|subj)), sleepstudy,
    contrasts = Dict(:days => DummyCoding()))
```

kleinschmidt:
> @kleinschmidt This is what's failing (it's a non sense model but it was chosen to show off the formula syntax on something quick to fit)
> 
> ```julia
> using MixedModels
> sleepstudy = MixedModels.dataset(:sleepstudy)
> fit(MixedModel, @formula(reaction ~ 1 + days + (1|subj) + zerocorr(days|subj)), sleepstudy,
>     contrasts = Dict(:days => DummyCoding()))
> ```

hmmmmmmmm that makes me think there may be an actual error there, it looks like `(Intercept)` is showing up twice when it should only be once...

kleinschmidt:
AH I think I know what's happened here 😞 

the hacky `FunctionTerm{typeof(|)}`-has-infinite-degree thing is causing the "bare" ranef to get sorted last:

```julia
julia> ff = @formula(reaction ~ 1 + days + (1|subj) + zerocorr(days|subj))
FormulaTerm
Response:
  reaction(unknown)
Predictors:
  1
  days(unknown)
  (days,subj)->zerocorr(days | subj)
  (subj)->1 | subj
```

I think we can work around this by adding a method for `degree(::FunctionTerm{typeof(zerocorr)})` buuuut I don't like this precedent...

PR #665, opened 2023-02-07
set Julia minimum compat to 1.8

palday:
Thanks for contributing!

- [x] add entry in NEWS.md


dmbates (approved):
Thanks for doing this and for updating the version of ubuntu on which the checks are made.

PR #671, opened 2023-03-07
hide fit progress during precompilation

palday:
Thanks for contributing!

- [ ] add entry in NEWS.md


dmbates (approved):
Thanks for thinking of this.  I had noticed the output during pre-compilation but had not thought to suppress it.

PR #674, opened 2023-04-10
Deprecate mulithreading in bootstrap

### LINKED ISSUE #640: Failures in threaded_replicate test
I am pretty consistently getting about a 50% failure rate on that test.  @palday Do you also get failures on that?  If so, can we suppress it until we find out why it is failing?

palday:
closes #640 

- [x] add entry in NEWS.md

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -124,21 +124,18 @@ end
 _is_logging(io) = isa(io, Base.TTY) == false || (get(ENV, "CI", nothing) == "true")
 
 """
-    replicate(f::Function, n::Integer; use_threads=false)
+    replicate(f::Function, n::Integer; hide_progress=false)
 
 Return a vector of the values of `n` calls to `f()` - used in simulations where the value of `f` is stochastic.
 
 `hide_progress` can be used to disable the progress bar. Note that the progress
 bar is automatically disabled for non-interactive (i.e. logging) contexts.
-
-!!! warning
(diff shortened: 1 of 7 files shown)


src/utilities.jl
```diff
@@ -124,39 +124,30 @@ end
...
-    If `f()` is not thread-safe or depends on a non thread-safe RNG,
-    then you must set `use_threads=false`. Also note that ordering of replications
-    is not guaranteed when `use_threads=true`, although the replications are not
-    otherwise affected for thread-safe `f()`.
 """
 function replicate(f::Function, n::Integer; use_threads=false, hide_progress=false)
-    # no macro version yet: https://github.com/timholy/ProgressMeter.jl/issues/143
+    use_threads && Base.depwarn(
+        "use_threads is deprecated and will be removed in a future release",
+        :parametricbootstrap,
```
dmbates, line 137: Should this be `:replicate,` ?

dmbates (approved):
I think that the name of the function needs to be changed in a `@depwarn` call in src/utilities.jl.  Otherwise looks good to go.  Thanks for doing this.

src/utilities.jl
palday: woops, yes, that was a copy-pasto.

palday:
```suggestion
        :replicate,
```

PR #676, opened 2023-04-12
support rank deficiency in prediction

### LINKED ISSUE #642: `predict` will give incorrect answers when `newdata` is rank deficient
This is because the new model matrix will be pivoted, but the coefficients from the initial fit shouldn't be.

palday:
closes #642

- [x] add entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 MixedModels v4.10.0 Release Notes
 ==============================
+* Rank deficiency in prediction is now supported, both when the original model was fit to rank-deficient data and when the new data are rank deficient. The behavior is consistent but potentially be surprising when both old and new data are rank deficient. See the `predict` docstring for an example. [#676]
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,5 +1,6 @@
 MixedModels v4.10.0 Release Notes
 ==============================
+* Rank deficiency in prediction is now supported, both when the original model was fit to rank-deficient data and when the new data are rank deficient. The behavior is consistent but potentially be surprising when both old and new data are rank deficient. See the `predict` docstring for an example. [#676]
```
dmbates, line 3: "but may be surprising ..." instead of "but potentially be surprising ..."?

dmbates (approved):
One suggested change in wording in the release notes.

NEWS.md
palday:
```suggestion
* Rank deficiency in prediction is now supported, both when the original model was fit to rank-deficient data and when the new data are rank deficient. The behavior is consistent but may be surprising when both old and new data are rank deficient. See the `predict` docstring for an example. [#676]
```

PR #634, opened 2022-08-11
raneftables returning NamedTuple of DictTables

dmbates:
- This is the first stage exploration of returning tabular representations for some of the `MixedModel` properties through `TypedTables.jl`
- Change the result of `raneftables` to be a `NamedTuple` of `DictTables`
    + An individual table gives an informative and attractive printed representation
    + Printing the `NamedTuple` of `DictTables` is messy and confusing
    + Extracting properties from the `DictTable` can be awkward b/c the column name "(Intercept)", purposely chosen to be difficult to represent as a `Symbol`, must be written as `var"(Intercept)"` or an equivalent 

For example,
```julia
julia> slm01 = let
           form = @formula reaction ~ 1 + days + (1+days|subj)
           fit(MixedModel, form, sleep; contrasts)
       end;

julia> raneftables(slm01)  # confusing display
(subj = {"S308" = (subj = "S308", var"(Intercept)" = 2.8158188821572576, days = 9.075511758123813), "S309" = (subj = "S309", var"(Intercept)" = -40.04844169634357, days = -8.644079444480187), "S310" = (subj = "S310", var"(Intercept)" = -38.43306372090656, days = -5.513397995849837), "S330" = (subj = "S330", var"(Intercept)" = 22.832112155574148, days = -4.658717421575887), "S331" = (subj = "S331", var"(Intercept)" = 21.549840457094, days = -2.944492928895566), "S332" = (subj = "S332", var"(Intercept)" = 8.815541388546052, days = -0.23520071393588618), "S333" = (subj = "S333", var"(Intercept)" = 16.441907858079695, days = -0.15880872279991523), "S334" = (subj = "S334", var"(Intercept)" = -6.996670702793803, days = 1.0327272102714635), "S335" = (subj = "S335", var"(Intercept)" = -1.0375872924983427, days = -10.599415818014403), "S337" = (subj = "S337", var"(Intercept)" = 34.66629369085548, days = 8.6323845720063), …},)

julia> retbl = only(raneftables(slm01))  # cleaner display
DictTable with 2 columns and 18 rows:
 subj   (Intercept)  days
 ─────┬───────────────────────
 S308 │ 2.81582      9.07551
 S309 │ -40.0484     -8.64408
 S310 │ -38.4331     -5.5134
 S330 │ 22.8321      -4.65872
 S331 │ 21.5498      -2.94449
 S332 │ 8.81554      -0.235201
 S333 │ 16.4419      -0.158809
 S334 │ -6.99667     1.03273
[...]

PR #682, opened 2023-04-25
undo #634, return Table not DictTable

dmbates:
- Return a `TypedTables.Table` not a `DictTable` in `raneftables`.  Manipulation of a `DictTable` becomes too awkward for anything other than extracting a row by key.

- [x] add entry in NEWS.md


src/mixedmodel.jl
```diff
@@ -92,7 +92,7 @@ StatsAPI.predict(m::MixedModel) = fitted(m)
 
 function retbl(mat, trm)
     nms = (fname(trm), Symbol.(trm.cnames)...)
-    return DictTable(
+    return Table(
     [NamedTuple{nms}((l, view(mat, :, i)...),) for (i, l) in enumerate(trm.levels)]
 )
```
palday, line 97:
```suggestion
    nt = [NamedTuple{nms}((l, view(mat, :, i)...),) for (i, l) in enumerate(trm.levels)]
    return Table(nt)
```

dmbates:
@JuliaRegistrator register

PR #684, opened 2023-04-25
use Suppressor.jl to hide log messages in tests

palday:
- [ ] ~add entry in NEWS.md~
- [ ] ~after opening this PR, add a reference and run `docs/NEWS-update.jl` to update the cross-references.~

- [ ] ~I've bumped the version appropriately~


dmbates:
Could this be implemented with `Logging.set_min_enabled_level` and avoid having a dependency on a package outside of the stdlib?

dmbates:
> Could this be implemented with `Logging.set_min_enabled_level` and avoid having a dependency on a package outside of the stdlib?

Not sure where I got that name but it doesn't appear to exist.  I guess that setting the logging level is trickier than I imagined, which, I suppose, is why the Suppressor package was created.

Is it feasible to wrap a section in `Logging.disable_logging(Logging.Warn)` `Logging.disable_logging(Logging.Debug)`?

palday:
This is "just" for the tests so won't impact most users. Suppressor is a pretty lightweight package.

dmbates (approved):

Issue #691, opened 2023-05-15
Redefinition of `models` method in test scripts


dmbates:
Julia v1.10.0-DEV versions are complaining about redefinition of `models` methods in the test scripts
```
Test Summary: | Pass  Total  Time
Missing Omit  |    2      2  0.5s
WARNING: Method definition models(Symbol) in module Main at /Users/dmbates/.julia/dev/MixedModels/test/modelcache.jl:47 overwritten on the same line (check for duplicate calls to `include`).
Test Summary: | Pass  Total  Time
isnested      |    5      5  1.3s
```

Should we wrap the `include` calls for `modelcache.jl` in a test of whether the method exists?

palday:
Yes, that sounds like a good idea to reduce the visual clutter.

dmbates:
Closed by #692

PR #692, opened 2023-05-16
avoid repeated def of models method

dmbates:
Avoid redefinition of a method in an included file (#691)
Julia v1.10.0-DEV warns about this when running tests.

### DIFF
--- a/test/modelcache.jl
+++ b/test/modelcache.jl
@@ -44,8 +44,11 @@ using MixedModels: dataset
 
 @isdefined(allfms) || const global allfms = merge(fms, gfms)
 
-function models(nm::Symbol)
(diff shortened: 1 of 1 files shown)


palday (approved):

dmbates:
Thanks for looking at this, @palday .  I don't think there needs to be even a patch release for this change - it is minor and has no effect other than to avoid some warnings on a development version of Julia.

Should I modify `NEWS.md` before merging the PR? If so, what would I use as the `Release Notes` number?  Shall I just assume we will have a `v4.14.1`?

palday:
No need for NEWS or patch bump. It only affects tests and not user functionality.

PR #694, opened 2023-06-04
distributed and reduced precision bootstrap

palday:
The idea for the reduced precision bootstrap came out of my work looking at the optimizer/convergence behavior.

- [x] add entry in NEWS.md


dmbates:
The example of changing `ftol_rel` on the parametric bootstrap for the "maximal" kb07 model is most impressive!  On my computer it took 33 minutes to generate 1000 samples with the default tolerance, 1 minute with 1e-3 tolerance and 6 minutes with 1e-6.  And the conclusions from each of these samples are substantively the same.

dmbates:
@palday Is there anything in particular that you would like me to concentrate on when reviewing this PR?

palday:
@dmbates whether the general reduced precision approach is valid statistically.

dmbates:
Regarding the validity of allowing more slack in `ftol_abs`, the whole rationale of the parametric bootstrap is that of drawing a sample from a population of possible parameter values.  In everything that we have seen the progress of the iterations is to make big gains initially then settle down to a long, drawn-out validation of being at a local optimum.  If we are just generating a sample, a high level of precision of the estimates is unnecessary because it is just one result of several simulated results.

dmbates:
The check on the nightly version failed to compile some of the dependencies of BSplineKit.jl  I saw that on a locally compiled nightly too and I don't think we will need to take action on it.

dmbates (approved):
Thanks for doing this.  I think the `optsum_overrides` will be very useful.

palday:
> Regarding the validity of allowing more slack in `ftol_abs`, the whole rationale of the parametric bootstrap is that of drawing a sample from a population of possible parameter values. In everything that we have seen the progress of the iterations is to make big gains initially then settle down to a long, drawn-out validation of being at a local optimum. If we are just generating a sample, a high level of precision of the estimates is unnecessary because it is just one result of several simulated results.

@dmbates Exactly my thoughts! I just wanted a sanity check.

Thanks for the other edits. 😄

PR #695, opened 2023-06-15
update CI actions

palday:
No release or NEWS update necessary


kleinschmidt (approved):
seems fine to me.  why get rid of MKL?

palday:
> why get rid of MKL?

We weren't using it and the setup there predates libblastrampoline (Julia 1.7) and wouldn't actually load MKL on current julia.

PR #702, opened 2023-08-04
Enhance the tbl property of a parametricbootstrap result

dmbates:
- allow _generatesyms with string for tag
- incorporate correlation estimates in table of parameter estimates from bootstrap


dmbates:
I believe the wonderful Benchmark comparisons are artificial.  See Discussion #712

dmbates:
@palday I think this is ready to go now.  I can add to the documentation later.  The general idea is to extract the `tbl` property of a bootstrap result.  Plots of, e.g., empirical density estimates can be made using the `wide` approach in `AlgebraOfGraphics`.  I will document that later.

dmbates:
@palday I have updated the documentation but not yet switched the documentation to use [AlgebraOfGraphics.jl](https://github.com/MakieOrg/AlgebraOfGraphics.jl].  As an example of how to use AlgebraOfGraphics on the table of parametricbootstrap results
```
import CairoMakie
using AlgebraOfGraphics, MixedModels
using MixedModels: dataset

CairoMakie.activate!(; type="svg")
contrasts = Dict{Symbol, Any}(:subj => Grouping());

m01 = let f = @formula reaction ~ 1 + days + (1 + days|subj)
    fit(MixedModel, f, dataset(:sleepstudy); contrasts)
end

m01samp = parametricbootstrap(10_000, m01; hide_progress=true);
tbl = m01samp.tbl

draw(
    let σs = collect(filter(startswith('σ'), string.(propertynames(tbl))))
        data(tbl) *
        mapping(
            σs .=> "Bootstrap replicates of standard deviation parameters",
            color=dims(1) => renamer(σs),
        ) *
        density()
    end
)
```

Note that if we `import CairoMakie` instead of `using` then we do not need to qualify visualization functions like `density`. 

Do you think we should create a function in `MixedModelsMakie` to encapsulate the operations in the `draw` call?

test/bootstrap.jl
```diff
@@ -184,7 +184,12 @@ end
         @test pb1 == restorereplicates(seekstart(io), MixedModels.unfit!(deepcopy(m1)))
 
         @test pb1 ≈ restorereplicates(seekstart(io), m1, Float16) rtol=1
-    end
+        # these tests must be last because the tbl extractor changes the λ field in the sample
```
palday, line 187: Why does the extractor change the lambda field? Are we using the lambda field anywhere else where that change might lead to unexpected behavior depending on the order a user looks at the `tbl` property?

palday:
> Do you think we should create a function in `MixedModelsMakie` to encapsulate the operations in the `draw` call?

Yes, though potentially AoG support should be in a package extension, in case somebody is using only the Makie and not AoG components. 

We could also add support for plotting the sigmas to `ridgeplot`.

test/bootstrap.jl
dmbates: The way that the standard deviation and correlation parameters are evaluate from sigma and theta is to populate lambda with the values of the theta parameters and determine sigma times the length of each row in each element of lambda.  While doing this for a lower triangular component of lambda I normalize the row after taking its length then evaluate the inner products of the normalized rows, which gives the correlations.  If we were confident that the lambda value always corresponds to the last value of theta, which is what happens when it is generated, we could simply install that value of theta.  Otherwise we should extract theta from lambda, create the table and reinstall the saved theta.  It means we need to have get\theta code as well as set\theta!   Alternatively, and more simply, we could save a copy of lambda, create the table and copy the saved lambda back into the bootstrap struct.

palday: > Alternatively, and more simply, we could save a copy of lambda, create the table and copy the saved lambda back into the bootstrap struct.

I think this is the way to go.

dmbates:
> > Alternatively, and more simply, we could save a copy of lambda, create the table and copy the saved lambda back into the bootstrap struct.

I think this is the way to go. 

Done.

docs/src/bootstrap.md
```diff
@@ -38,41 +34,29 @@ using Random
 ```@example Main
 dyestuff = MixedModels.dataset(:dyestuff)
 m1 = fit(MixedModel, @formula(yield ~ 1 + (1 | batch)), dyestuff)
-DisplayAs.Text(ans) # hide
 ```
 
-To bootstrap the model parameters, first initialize a random number generator then create a bootstrap sample
+To bootstrap the model parameters, first initialize a random number generator then create a bootstrap sample and extract the `tbl` property, which is a `Table` - a lightweight type of dataframe.
```
palday, line 39:
```suggestion
To bootstrap the model parameters, first initialize a random number generator then create a bootstrap sample and extract the `tbl` property, which is a `Table` - a lightweight dataframe-like object.
```

src/bootstrap.jl
```diff
@@ -544,40 +549,89 @@ function tidyσs(bsamp::MixedModelFitCollection{T}) where {T}
     return result
 end
 
+_nρ(d::Diagonal) = 0
+_nρ(t::LowerTriangular) = kchoose2(size(t.data, 1))
+
+function σρnms(λ)
+    σsyms = _generatesyms('σ', sum(first ∘ size, λ))
+    ρsyms = _generatesyms('ρ', sum(_nρ, λ))
+    val = Symbol[]
```
palday, line 558:
```suggestion
    val = Symbol[]
    sizehint!(val, length(σsyms) + length(ρsyms))
```

src/bootstrap.jl
```diff
@@ -544,40 +549,89 @@ function tidyσs(bsamp::MixedModelFitCollection{T}) where {T}
...
+end
+
+function σρ!(v::AbstractVector, d::Diagonal, σ)
+    for dd in d.diag
+        push!(v, σ * dd)
+    end
+    return v
+end
+
+function σρ!(v::AbstractVector{T}, t::LowerTriangular{T}, σ::T) where {T}
```
palday, line 587: [noblock]

it would be go to have nominal docstrings for these internal methods -- between the name and the function body, I can figure out what they're doing, but I always have to remind myself. 😄

src/bootstrap.jl
```diff
@@ -544,40 +549,89 @@ function tidyσs(bsamp::MixedModelFitCollection{T}) where {T}
...
+    dat = t.data
```
palday, line 588:
```suggestion
    dat = parent(t)
```

src/bootstrap.jl
```diff
@@ -544,40 +549,89 @@ function tidyσs(bsamp::MixedModelFitCollection{T}) where {T}
...
-                v[vpos] = r.σ * norm(λr)
-                vpos += 1
-            end
+            σρ!(v, l, σ)
         end
-        push!(val, tblrowtyp(v))
+        append!(v, θ)
     end
-    return val
+    m = collect(transpose(reshape(v, (m, n))))
```
palday, line 632: `permutedims!` might be useful here

palday (approved):
looks good -- I think this counts as a 'feature', so can you

- [x] bump the minor version number
- [ ] add a NEWS entry

?

src/bootstrap.jl
```diff
@@ -544,40 +549,89 @@ function tidyσs(bsamp::MixedModelFitCollection{T}) where {T}
...
+    append!(syms, _generatesyms('β', length(β)))
+    push!(syms, :σ)
+    append!(syms, σρnms(λ))
+    return append!(syms, _generatesyms('θ', length(θ)))
+end
+
+function σρ!(v::AbstractVector, d::Diagonal, σ)
+    for dd in d.diag
+        push!(v, σ * dd)
+    end
```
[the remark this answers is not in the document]
dmbates: Interestingly, your suggestion was the original way it was written.  Then I thought it would be better to avoid the allocation of the vector.  Either way is fine with me.

docs/src/bootstrap.md
dmbates: Agreed

src/bootstrap.jl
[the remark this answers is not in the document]
palday: Then slight stylistic preference for `append!` :smile:

src/bootstrap.jl
```diff
@@ -444,7 +448,7 @@ Return the shortest interval containing `level` proportion for each parameter fr
     a breaking change.
 """
 function shortestcovint(bsamp::MixedModelFitCollection{T}, level=0.95) where {T}
-    allpars = bsamp.allpars
+    allpars = bsamp.allpars  # probably simpler to use .tbl instead of .allpars
```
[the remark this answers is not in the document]
dmbates: Sure.  The result will need to be expressed slightly differently because the column names of the table are named like :β1,  :β2,  :β3, :β4 and the current output uses the three-column naming convention.

src/bootstrap.jl
dmbates: My feeling was that the call to `sizehint!` might take more time than any resizing that needs to be done but it shouldn't be a big consideration.

palday: Yeah, ideally we're talking about relatively small arrays, but then again somebody will try to fit a maximal model with 100+ covariance parameters.... :thinking:

dmbates:
@palday Would you take one more look and squash and merge if this seems okay to you?

PR #706, opened 2023-08-19
support for NLopt 1.0

palday:
This bumps NLopt support for the 1.0 release, which moves MathOptInterface support to an extension. This dramatically improves load times for MixedModels, which doesn't need MathOptInterface. I've dropped support for NLopt 0.5 because the callbacks don't work on Julia 1.5+ (due to the stack allocation of immutable structs -- the struct was changed to mutable in 0.6), since we're already compat bounded to Julia 1.8.


palday:
image

dmbates:
I may have done the approvals of PR's backwards.  I approved and committed the PR that bumps the version to 4.18.0 whereas this PR bumps the version to 4.17.1.  Should that be changed now to 4.18.1?  I haven't alerted the JuliaRegistrator to the version bump to 4.18.0

dmbates (approved):

palday:
@dmbates we can just roll them together into a single release as 4.18.0.

PR #710, opened 2023-08-22
add benchmark GHA

palday:


dmbates:
I think many of simple scalar benchmarks are unnecessary because they take so little time.  They can show large relative differences because the denominator is so small.  We may want to just run the crossed vector benchmarks.

benchmark/run.jl
```diff
@@ -0,0 +1,9 @@
+using Pkg
+Pkg.develop(PackageSpec(path=pwd()))
```
ararslan, line 2: `pwd()` can differ depending on where the user has invoked Julia, `cd`'d to, etc. It's a good option when the code is written inline in the YAML file but is a bit iffy to have in a file. To ensure you're always pointing at the right thing, I'd make this:
```suggestion
Pkg.develop(PackageSpec(path=dirname(@__DIR__)))
```

benchmark/benchmarks.jl
```diff
@@ -85,15 +85,15 @@ function fitbobyqa(dsnm::Symbol, i::Integer)
...
-SUITE["simplescalar"] = BenchmarkGroup(["single", "simple", "scalar"])
-for (ds, i) in [
-    (:dyestuff, 1),
-    (:dyestuff2, 1),
-    (:pastes, 1),
-    (:sleepstudy, 1),
-]
-    SUITE["simplescalar"][string(ds, ':', i)] = @benchmarkable fitbobyqa($ds, $i)
-end
+# SUITE["simplescalar"] = BenchmarkGroup(["single", "simple", "scalar"])
```
ararslan, line 88: Could you add a comment about why this is commented out so somebody in the future can uncomment as applicable?

palday:
```suggestion
# these tests are so fast that they can be very noisy because the denominator is so small, 
# so we disable them by default for auto-benchmarking
# SUITE["simplescalar"] = BenchmarkGroup(["single", "simple", "scalar"])
```

PR #711, opened 2023-08-23
remove Pkg.add from Benchmark

palday:


dmbates (approved):
Thanks for cleaning this up.

I feel that we should think about whether we want to continue to do package benchmarks.  It seems that the results are so variable as not to be worthwhile.

For example, if you check out the main branch and run the package benchmark you are essentially benchmarking the main branch against itself and you frequently end up with "significant" differences.

palday:
> Thanks for cleaning this up.
> 
> I feel that we should think about whether we want to continue to do package benchmarks. It seems that the results are so variable as not to be worthwhile.
> 
> For example, if you check out the main branch and run the package benchmark you are essentially benchmarking the main branch against itself and you frequently end up with "significant" differences.

I agree. I can disable this as CI and we can revisit this if we find a good set of benchmarks / thresholds so that we're not just seeing noise in GitHub's server farm.

PR #735, opened 2023-12-14
nest log and throws tests (1.11-nightly compat)

palday:
I don't know what's changed in the nightly but something about catching the exception messes up the `@test_logs` check even though the error is still being logged. This fixes that and simplifies things a bit.


dmbates (commented):
I incorporated the changes to avoid `diagind` so that the tests on nightly make it through.

dmbates (approved):

dmbates:
Would it be worthwhile doing a patch release with this PR?  It only affects operations on the development branch.

palday:
@dmbates I don't think this needs to be released -- it's mostly useful for us testing against the nightly builds. If someone really needs it, well it's easy enough to do a quick release.

PR #738, opened 2024-01-17
Replace broadcasted lambda with explicit loop

palday:
Thanks for contributing!

- [x] add entry in NEWS.md


dmbates:
@palday This looks reasonable.  Should we bring it up-to-date with the current main branch and merge it?  In other words, is there a reason that you have left this PR as a Draft?

PR #740, opened 2024-01-26
use FMA where possible in fitting

palday:
Thanks for contributing!

- [x] add entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,5 @@
+* Use `muladd` where possible to enable fused multiply-add (FMA) on architectures with hardware support. FMA will generally improve computational speed and gives more accurate rounding. [#740]
+
 MixedModels v4.22.4 Release Notes
 ==============================
 * Switch to explicit imports from all included packages (i.e. replace `using Foo` by `using Foo: Foo, bar, baz`) [#748]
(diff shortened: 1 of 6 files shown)


dmbates:
It seems that `muladd` is more general than `fma` but falls back on `fma` when available.  Is there a reason not to prefer `fma` over `muladd`?

palday:
There are two differences between `muladd` and `fma` that are potentially relevant:
1. `muladd` works with `missing`, but `fma` doesn't.
2. `fma` will use software FMA if it's not available in hardware, while `muladd` will simply fall back to the non-fused variant. This can introduce rounding differences in addition to the performance differences (software FMA is slower than muladd is slower than hardware FMA). 

If it were just (1), I would say use `muladd` everywhere. But I was unsure about (2) -- I was torn between consistent rounding across architectures and worse performance on legacy hardware. I don't have a strong preference either way, but I guess there is something consistent about using `muladd` everywhere and not thinking too hard about whether `missing` can arise in a given context or not.

palday:
After some more thought, I think it's better to use `muladd` everywhere. I will update accordingly.

dmbates (approved):

PR #744, opened 2024-02-18
A test using a deep copy of a model should reset the initial parameter values.

dmbates:
Because the initial values for the theta parameters are derived from the values stored in the reterms they will be the final parameter estimates if those terms are deep copied to create a new model.

The NLopt optimizers don't object to this but the PRIMA optimizers do.

### DIFF
--- a/test/matrixterm.jl
+++ b/test/matrixterm.jl
@@ -42,7 +42,11 @@ end
         @test typeof(X.x) <: SparseMatrixCSC
         @test X.rank == 28
         @test X.cnames == fe.cnames
-        m1 = fit!(LinearMixedModel(collect(m.y), X, deepcopy(m.reterms), m.formula); progress=false)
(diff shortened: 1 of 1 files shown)


dmbates:
The nightly tests are failing consistently on this PR.  It appears that the download of the datasets fails when installing `MixedModelsDatasets` package.

palday (approved):

palday:
> The nightly tests are failing consistently on this PR. It appears that the download of the datasets fails when installing `MixedModelsDatasets` package.

Yeah, I've run into this on other repositories as well and I don't know why it's happening -- I can't replicate with a local nightly build.

dmbates:
@palday Okay to merge this PR without bumping the patch number?  This only affects a test and doesn't change the result of the test - just the run time.

palday:
@dmbates Go for merge without patch bump. 😄

PR #748, opened 2024-03-01
Switch to explicit imports of names from packages

dmbates:
With the availability of [ExplicitImports.jl](https://github.com/ericphanson/ExplicitImports.jl) it is feasible to list all the symbols used from a package in the dependencies instead of importing all the exported symbols.

These changes are based on the output of `ExplicitImports.print_explicit_imports(MixedModels)`.

Did behavior change?  No, the behavior didn't (or shouldn't have) change
- [x] add entry in NEWS.md


dmbates:
It appears that the test failure on nightly is the old problem of not finding the datasets to download.  I'm not sure why the other Julia releases can do this but the nightly release fails to do so.

ericphanson:
btw do file an issue if you run into any issues! Also if you wish, you can add ExplicitExports as a test dep and use the `check_*` functions to ensure no implicit exports or stale explicit exports.

NEWS.md
```diff
@@ -1,3 +1,8 @@
+MixedModels v4.22.4 Release Notes
+==============================
+* Switch to explicit imports from all included packages (i.e. replace `using Foo` by `using Foo: Foo, bar, baz`) [#748]
+* Reset parameter values before a `deepcopy` in an test (doesn't change test result) [#744]
```
palday, line 4:
```suggestion
* Reset parameter values before a `deepcopy` in a test (doesn't change test result) [#744]
```

Project.toml
```diff
@@ -20,6 +20,7 @@ ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
```
palday, line 23: since we already have `StaticArrays`, can we grab `SVector` from there? I know this doesn't increase our total dependency load, but it's another thing to track.

src/MixedModels.jl
```diff
@@ -1,35 +1,51 @@
...
-using LinearAlgebra: BlasFloat, BlasReal, HermOrSym, PosDefException, copytri!
+using Arrow: Arrow
 using Base: Ryu, require_one_based_indexing
-using DataAPI: levels, refpool, refarray, refvalue
-using GLM: Link, canonicallink, linkfun, linkinv, dispersion, dispersion_parameter
+using BSplineKit: BSplineKit
+using BSplineKit.BSplines: BSplineOrder
+using BSplineKit.BoundaryConditions: Natural
+using BSplineKit.DifferentialOps: Derivative
+using BSplineKit.SplineInterpolations: SplineInterpolation, interpolate
```
palday, line 9: can these be grabbed from the toplevel module (i.e. are they exported from the submodule to the toplevel?)?

Project.toml
ericphanson: filed https://github.com/ericphanson/ExplicitImports.jl/issues/24 fyi. I think though if you do `using StaticArrays: SVector`, then it won't complain about the implicit import. But when it makes suggestions, it always does the "defining" module.

palday (approved):

ericphanson:
BTW I put a PR to improve the printing, so next time you wouldn't have to manually regroup as much: https://github.com/ericphanson/ExplicitImports.jl/pull/27

dmbates:
Thank you, @ericphanson , for implementing this change in the printing.  That's very helpful.

PR #752, opened 2024-03-05
an assortment of CI bumps

palday:
originally started on this to see if it fixed the nightly CI problem, but then just decided to bump everything while I was at it.


dmbates (approved):
Thanks for this.

PR #755, opened 2024-03-19
Support for rank deficiency in bootstrap 

### LINKED ISSUE #754: Error on `parametricbootstrap` for rank deficient models
Looks like https://github.com/JuliaStats/MixedModels.jl/blob/cfd302310ba66ad5008e97738bd6710cf5a7509a/src/bootstrap.jl#L250 is problematic for rank deficient models since `β_names` has one name per fixed effect (which is ≤ the number of coefficients) but `fixef!(βsc, m)` is being called with the full coefficient vector as the first argument.

palday:
closes #754 

- [x] add entry in NEWS.md

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -203,7 +203,11 @@ of coverage intervals.
 
 
 !!! note
-    All coefficients are bootstrapped, including inestimatable ones from the rank deficient case.
+    All coefficients are bootstrapped. In the rank deficient case, the inestimatable coefficients are
+    treated as -0.0 in the simulations underlying the bootstrap, which will generally result
+    in their estimate from the simulated data also being being inestimable and thus set to -0.0.
+    **However this behavior behavior may change in future releases to explicitly drop the
+    extraneous columns before simulation and thus not include their estimates in the bootstrap result.**
 """
 function parametricbootstrap(
     rng::AbstractRNG,
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -206,7 +206,7 @@ of coverage intervals.
     All coefficients are bootstrapped. In the rank deficient case, the inestimatable coefficients are
     treated as -0.0 in the simulations underlying the bootstrap, which will generally result
     in their estimate from the simulated data also being being inestimable and thus set to -0.0.
-    **However this behavior behavior may change in future releases to explicitly drop the
+    **However this behavior may change in future releases to explicitly drop the
     extraneous columns before simulation and thus not include their estimates in the bootstrap result.**
 """
 function parametricbootstrap(
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -80,7 +80,7 @@ been constructed with the same structure as the source of the saved replicates.
 
 The two-argument method constructs a [`MixedModelBootstrap`](@ref) with the same eltype as `m`.
 If an eltype is specified as the third argument, then a `MixedModelBootstrap` is returned.
-If a subtype of `MixedModelFitCollection` is specified as the third argument, then that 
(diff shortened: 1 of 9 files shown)


src/bootstrap.jl
```diff
@@ -103,7 +103,7 @@ function restorereplicates(
     rep = first(Tables.rows(tbl))
     allgood =
         length(rep.θ) == length(m.θ) &&
-        string.(propertynames(rep.β)) == Tuple(coefnames(m))
+        string.(propertynames(rep.β)) == Tuple(fixefnames(m))
```
palday, line 106:
```suggestion
        string.(propertynames(rep.β)) == Tuple(coefnames(m))
```

src/bootstrap.jl
```diff
@@ -200,6 +200,14 @@ fit during the bootstrapping process. For example, `optsum_overrides=(;ftol_rel=
 reduces the convergence criterion, which can greatly speed up the bootstrap fits.
 Taking advantage of this speed up to increase `n` can often lead to better estimates
 of coverage intervals.
+
+
+!!! note
+    All coefficients are bootstrapped. In the rank deficient case, the inestimatable coefficients are
+    treated as -0.0 in the simulations underlying the bootstrap, which will generally result
+    in their estimate from the simulated data also being being inestimable and thus set to -0.0.
+    **However this behavior behavior may change in future releases to explicitly drop the
```
ararslan, line 209:
```suggestion
    **However this behavior may change in future releases to explicitly drop the
```

src/simulate.jl
```diff
@@ -152,20 +149,17 @@ end
 function simulate!(
     rng::AbstractRNG, y::AbstractVector, m::LinearMixedModel{T}; β=m.β, σ=m.σ, θ=m.θ
 ) where {T}
-    length(β) == length(fixef(m)) ||
-        length(β) == length(coef(m)) ||
+    length(β) == length(m.feterm.piv) || length(β) == m.feterm.rank ||
```
ararslan, line 152: I don't think I understand this change. IIUC, the `piv` field of a model's `FeTerm` will always have length equal to `length(coef(model)) == size(modelmatrix(model), 2)`, and `length(fixef(model)) == rank(model)` in general but is less than `length(piv)` for rank deficiency. Assuming my understanding is correct, isn't this exactly the same as what was there?

src/simulate.jl
```diff
@@ -152,20 +149,17 @@ end
...
     isempty(θ) || setθ!(m, θ)
 
-    if length(β) ≠ length(coef(m))
-        padding = length(coef(m)) - length(β)
-        for ii in 1:padding
-            push!(β, -0.0)
-        end
+    if length(β) ≠ length(m.feterm.piv)
+        padding = length(model.feterm.piv) - m.feterm.rank
+        append!(β, fill(-0.0, padding))
```
ararslan, line 162: Also unsure how this differs functionally from what was there, which suggests to me that I'm missing something important about the lengths of these internal structures.

palday: doesn't differ! I assumed that a single `fill` and `append` would have better asymptotic performance than a looped `push!` (although that may not be the case depending on whether the backing memory is actually allocated -- it's generally allocated on resize to be larger than needed to avoid costs from multiple `push!`s)

ararslan (approved):
Confirmed that this works in the case that motivated #754 and that
```julia
dropped_idx = findall((coef(model) .=== -0.0) .& isnan.(stderror(model)))
dropped_coefs = coefnames(model)[dropped_idx]
all(nt -> !(nt.coefname in dropped_coefs) || nt.β !== -0.0, boot.β)
```
is `true`. (Extending the added test to the case of multiple dropped columns)

PR #760, opened 2024-03-25
add Apple silicon to CI

palday:
no release necessary

I like the idea of testing another CPU architecture to work as further sanity check on our numerical tolerances.


dmbates (approved):
Thanks for doing this.

PR #764, opened 2024-04-23
use `@everywhere ` in distributed bootstrap example

palday:
As pointed out in #763


dmbates (approved):
Thanks for making this change.

PR #767, opened 2024-05-15
delegate as many GLMM properties as possible to internal LMM

palday:
@dmbates is there a giant footgun I'm missing by swapping to "delegate by default" instead of "delegate by explicit list"? 

- [x] add entry in NEWS.md


dmbates (approved):
Seems reasonable to me.  The only footgun I can imagine is if there is a property of an LMM that wouldn't make sense for a GLMM, like σ, but all of those seem to be handled in earlier elseif blocks.

PR #772, opened 2024-06-21
Small performance tweaks

palday:
Thanks for contributing!

- [x] add entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-MixedModels v4.24.2 Release Notes
+MixedModels v4.25 Release Notes
 ==============================
 - Add type notations in `pwrss(::LinearMixedModel)` and `logdet(::LinearMixedModel)` to enhance type inference. [#773]
 - Take advantage of type parameter for `StatsAPI.weights(::LinearMixedModel{T})`. [#772]
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@ MixedModels v4.24.2 Release Notes
     - user-specified `σ` is actually used, defaulting to existing value
     - `REML` defaults to model's already specified REML value.
 - Clean up code of keyword convenience constructor for `OptSummary`. [#772]
+- Refactor thresholding parameters for forcing near-zero parameter values into `OptSummary`. [#772]
 
 MixedModels v4.24.1 Release Notes
 ==============================
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 MixedModels v4.24.2 Release Notes
 ==============================
 - Add type notations in `pwrss(::LinearMixedModel)` and `logdet(::LinearMixedModel)` to enhance type inference. [#773]
+- Take advantage of type parameter for `StatsAPI.weights(::LinearMixedModel{T})`. [#772]
+- Fix use of kwargs in `fit!((::LinearMixedModel)`: [#772]
+    - user-specified `σ` is actually used, defaulting to existing value
+    - `REML` defaults to model's already specified REML value.
 
 MixedModels v4.24.1 Release Notes
 ==============================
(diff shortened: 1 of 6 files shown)


src/linearmixedmodel.jl
```diff
@@ -408,7 +408,7 @@ function createAL(reterms::Vector{<:AbstractReMat{T}}, Xy::FeMat{T}) where {T}
             end
         end
     end
-    return A, L
+    return identity.(A), identity.(L)
```
palday, line 411: this will tighten up the eltype of the vector when all elements are of a single type

src/linearmixedmodel.jl
```diff
@@ -895,7 +893,7 @@ end
 
 The penalized, weighted residual sum-of-squares.
 """
-pwrss(m::LinearMixedModel{T}) where {T} = abs2(last(last(m.L)))::T
+pwrss(m::LinearMixedModel{T}) where {T} = abs2(last(last(m.L))::T)
```
palday, line 896: moving the type assertion inward means that `abs2` now dispatches on an inferred type

src/optsummary.jl
```diff
@@ -31,63 +31,40 @@ The latter four fields are MixedModels functionality and not related directly to
     the future to use a different subtype of `AbstractVector` (e.g., `StaticArrays.SVector`)
     for each snapshot without being considered a breaking change.
 """
-mutable struct OptSummary{T<:AbstractFloat}
+Base.@kwdef mutable struct OptSummary{T<:AbstractFloat}
```
palday, line 34: I hope that this will make some things easier to maintain

src/linearmixedmodel.jl
dmbates: It is a good idea but I don't think it will change things in the current setup. At least for `L`, `first(L)` is always `Diagonal` or `UniformBlockDiagonal` and `last{L)` is always `Matrix{T}`.

NEWS.md
```diff
@@ -1,6 +1,16 @@
...
+- Take advantage of type parameter for `StatsAPI.weights(::LinearMixedModel{T})`. [#772]
+- Fix use of kwargs in `fit!((::LinearMixedModel)`: [#772]
+    - user-specified `σ` is actually used, defaulting to existing value
+    - `REML` defaults to model's already specified REML value.
+- Clean up code of keyword convenience constructor for `OptSummary`. [#772]
+- Refactor thresholding parameters for forcing near-zero parameter values into `OptSummary`. [#772]
+
 MixedModels v4.24.1 Release Notes
 ==============================
-Add type notations in `pwrss(::LinearMixedModel)` and `logdet(::LinearMixedModel)` to enhance type inference. [#773]
```
dmbates, line 3: Thanks for picking up that I had set the version to the wrong value and had the notes in the wrong place.

src/linearmixedmodel.jl
```diff
@@ -36,8 +36,8 @@ struct LinearMixedModel{T<:AbstractFloat} <: MixedModel{T}
     sqrtwts::Vector{T}
     parmap::Vector{NTuple{3,Int}}
     dims::NamedTuple{(:n, :p, :nretrms),NTuple{3,Int}}
-    A::Vector{AbstractMatrix{T}}            # cross-product blocks
-    L::Vector{AbstractMatrix{T}}
+    A::Vector{<:AbstractMatrix{T}}            # cross-product blocks
+    L::Vector{<:AbstractMatrix{T}}
```
dmbates, line 40: Again, thanks for picking this up.  This is a sign that you write a lot of Julia code.

src/optsummary.jl
dmbates: Good idea.

dmbates (approved):

src/linearmixedmodel.jl
palday: Yeah, that matches my own sanity checks. I have some further ideas building upon this, but haven't had time to experiment more. 😄

PR #775, opened 2024-06-27
More sophisticated checks in restoreoptsum.

### LINKED ISSUE #774: restoreoptsum fails after change in JSON file structure in 4.25.0
One of the first checks in `restoreoptsum` is on the keys in the dictionary created from the JSON file.  If the optsum JSON file was created with a version prior to 4.25 attempts to restore this optsum under 4.25.0 will fail, with a somewhat misleading error message.  I will split the checks into two parts, one on the fields and one on the lower bounds and allow for patching pre-4.25 saved optsums to the 4.25 names.

dmbates:
Closes #774

- [x] add entry in NEWS.md

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ docs/jmd
 LocalPreferences.toml
 benchmark.md
 
+lcov.info
+coverage/
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+MixedModels v4.25.1 Release Notes
+==============================
+- Use more sophisticated checks on property names in `restoreoptsum` to allow for optsums saved by pre-v4.25 versions to be used with this version and later. 
+
 MixedModels v4.25 Release Notes
 ==============================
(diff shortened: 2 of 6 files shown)


palday:
@dmbates can you add a test with loading a 4.22 optsum? That would also serve as a potential test if we add even more fields (`restoreoptsum!` would break again and thus we would be reminded to update it).

dmbates:
I added tests and improved coverage slightly.

palday:
@dmbates I made a few changes directly:

- made sure that "optional" fields are actually loaded when they exist
- (fixed an old bug in how that was done for `fitlog`)
- issue a warning when loading an old fit

palday (approved):

dmbates:
Did I accidentally commit a `lcov.info` file?  I use the "Run tests with coverage" option in VS code which creates a `lcov.info` file and I see you have added that name to `.gitignore`

palday:
> Did I accidentally commit a `lcov.info` file? I use the "Run tests with coverage" option in VS code which creates a `lcov.info` file and I see you have added that name to `.gitignore`

No, but I almost did :joy:

dmbates:
@JuliaRegistrator register

PR #776, opened 2024-07-12
begin supporting `public` keyword

palday:
Thanks for contributing!

- [x] add entry in NEWS.md


dmbates (approved):
Good idea

PR #778, opened 2024-08-01
rework passing and handling of FE coefficients to `simulate`! for rank deficient models

palday:
- [x] add tests showing potential failure mode of old route 

- [x] add entry in NEWS.md

<details><summary> visual example that makes the old behavior clear </summary>

```julia
using DataFrames
using MixedModels
using StableRNGs
using Statistics
using Test

id = lpad.(string.(1:40), 2, "0")
B = ["b0", "b1", "b2"]
C = ["c0", "c1", "c2", "c3", "c4"]

df = DataFrame(reshape(collect(Iterators.product(B, C, id)), :), [:b, :c, :id])
df[!, :y] .= 0

filter!(df) do row
    b = last(row.b)
    c = last(row.c)
    return b != c
end

m = LinearMixedModel(@formula(y ~ 1 + b * c + (1|id)), df)
β = 1:rank(m)
σ = 1
simulate!(StableRNG(628), m; β, σ)
fit!(m)
boot = parametricbootstrap(StableRNG(271828), 1000,  m);

bootci = DataFrame(shortestcovint(boot))
filter!(:group => ismissing, bootci)
select!(bootci, :names => disallowmissing => :coef, :lower, :upper)
transform!(bootci, [:lower, :upper] => ByRow(middle) => :mean)
@test all(x -> isapprox(x[1], x[2]; atol=0.1),  zip(coef(m), bootci.mean))

using CairoMakie
using MixedModelsMakie
let f = Figure()
    ax = Axis(f[1,1])
    ridgeplot!(ax, boot)
    coefplot!(ax, m; color=:red)
    f
end

```

</details>

Red are the Wald CIs, Black is bootstrap.

before

after

### DIFF
--- a/src/simulate.jl
+++ b/src/simulate.jl
@@ -26,7 +26,7 @@ Overwrite the response (i.e. `m.trms[end]`) with a simulated response vector fro
 This simulation includes sampling new values for the random effects.
 
 `β` can be specified either as a pivoted, full rank coefficient vector (cf. [`fixef`](@ref))
-or as an unpivoted full dimension coefficient vector (cf [`coef`](@ref)), where the entries
+or as an unpivoted full dimension coefficient vector (cf. [`coef`](@ref)), where the entries
 corresponding to redundant columns will be ignored.
 
 !!! note
--- a/src/simulate.jl
+++ b/src/simulate.jl
@@ -158,7 +158,7 @@ function simulate!(
     isempty(θ) || setθ!(m, θ)
 
     if length(β) == length(pivot(m))
-        β = view(view(β,  pivot(m)), 1:rank(m))
+        β = view(view(β, pivot(m)), 1:rank(m))
     end
 
     # initialize y to standard normal
--- a/src/simulate.jl
+++ b/src/simulate.jl
@@ -26,7 +26,7 @@ Overwrite the response (i.e. `m.trms[end]`) with a simulated response vector fro
 This simulation includes sampling new values for the random effects.
 
 `β` can be specified either as a pivoted, full rank coefficient vector (cf. [`fixef`](@ref))
-or as an unpivoted full dimension coefficient vector (cf [`coef`](ref)), where the entries
+or as an unpivoted full dimension coefficient vector (cf [`coef`](@ref)), where the entries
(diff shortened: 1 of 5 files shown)


src/simulate.jl
```diff
@@ -147,20 +147,18 @@ function simulate!(
...
     σ = T(σ)
     θ = convert(Vector{T}, θ)
     isempty(θ) || setθ!(m, θ)
 
-    if length(β) ≠ length(pivot(m))
-        padding = length(pivot(m)) - rank(m)
-        append!(β, fill(-0.0, padding))
-        invpermute!(β, pivot(m))
+    if length(β) == length(pivot(m))
+        β = view(view(β, pivot(m)), 1:rank(m))
```
kleinschmidt, line 161: this is the branch that handles the `coef` case where we need to pull out the non-`-0.0` values in the correct order to multiply by `fullrank(m)` right?

src/simulate.jl
```diff
@@ -18,8 +18,8 @@ function simulate(m::MixedModel, args...; kwargs...)
 end
 
 """
-    simulate!(rng::AbstractRNG, m::MixedModel{T}; β=m.β, σ=m.σ, θ=T[])
-    simulate!(m::MixedModel; β=m.β, σ=m.σ, θ=m.θ)
+    simulate!(rng::AbstractRNG, m::MixedModel{T}; β=fixef(m), σ=m.σ, θ=T[])
+    simulate!(m::MixedModel; β=fixef(m), σ=m.σ, θ=m.θ)
```
kleinschmidt, line 22: I think it's worth documenting _somewhere_ the difference between the possible kinds of inputs to this function.  `coef`-style (all values in input model matrix order, pivoted out betas are `-0.0`) vs. `fixef`-style (only betas for full-rank columns, in pivot order)

kleinschmidt (approved):

src/simulate.jl
palday: correct.

src/simulate.jl
palday: 73d50ae0e96543728e41d48d3a6dc50d7362288e

PR #785, opened 2024-09-24
fix flipped lower and upper in profile CI

### LINKED ISSUE #784: `confint(::MixedModelProfile)` swaps `lower` and `upper` when profiled model is fitted with `REML`
All works fine for models fitted with ML:

```julia
m = fit(MixedModel, @formula(reaction ~ 1 + days + (1|subj)), MixedModels.dataset(:sleepstudy));
confint(profile(m))
```

```julia
DictTable with 3 columns and 4 rows:
 par   estimate  lower    upper
 ────┬───────────────────────────
 β1  │ 251.405   231.992  270.818
 β2  │ 10.4673   8.88655  12.048
 σ   │ 30.8954   27.8139  34.5911
 σ1  │ 36.0121   26.0079  52.9358
```

But with REML, `lower` and `upper` values are seemingly swapped for the βs:

```julia
refit!(m, REML=true);
confint(profile(m))
```

```julia
DictTable with 3 columns and 4 rows:
 par   estimate  lower    upper
 ────┬───────────────────────────
 β1  │ 251.405   265.901  236.909
 β2  │ 10.4673   11.6974  9.23715
 σ   │ 30.9912   27.8913  34.7109
 σ1  │ 37.1238   26.6233  55.1929
```

palday:
closes #784 

- [x] add entry in NEWS.md

### DIFF
--- a/src/profile/fixefpr.jl
+++ b/src/profile/fixefpr.jl
@@ -102,6 +102,10 @@ function profileβj!(
     ζv = getproperty.(tbl, :ζ)
     βv = getproperty.(tbl, sym)
     val.fwd[sym] = interpolate(βv, ζv, BSplineOrder(4), Natural())
+    if sym == :β1
+        @debug "" sym
+        @debug "" collect(zip(ζv, βv))
+    end
     val.rev[sym] = interpolate(ζv, βv, BSplineOrder(4), Natural())
     return val
 end
--- a/src/profile/fixefpr.jl
+++ b/src/profile/fixefpr.jl
@@ -102,10 +102,6 @@ function profileβj!(
     ζv = getproperty.(tbl, :ζ)
     βv = getproperty.(tbl, sym)
     val.fwd[sym] = interpolate(βv, ζv, BSplineOrder(4), Natural())
-    if sym == :β1
-        @debug "" sym
-        @debug "" collect(zip(ζv, βv))
-    end
     val.rev[sym] = interpolate(ζv, βv, BSplineOrder(4), Natural())
     return val
 end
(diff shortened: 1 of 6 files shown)


src/profile/fixefpr.jl
```diff
@@ -102,6 +102,10 @@ function profileβj!(
     ζv = getproperty.(tbl, :ζ)
     βv = getproperty.(tbl, sym)
     val.fwd[sym] = interpolate(βv, ζv, BSplineOrder(4), Natural())
+    if sym == :β1
+        @debug "" sym
+        @debug "" collect(zip(ζv, βv))
```
ararslan, line 107: Did you intend to leave this in or is it leftover from debugging?

palday: yes and yes. :wink: I was going to leave it for now in case I get the urge to do a deeper dive in the near future about what's causing the reversal. But I can also strip out it and add it if I ever get around to that.

src/profile/fixefpr.jl
```diff
@@ -102,6 +102,10 @@ function profileβj!(
...
+    end
```
palday, line 108:
```suggestion
```

src/profile/fixefpr.jl
ararslan: I remember seeing an issue quite a while back where having logging macro calls in your code at all came with a performance penalty, even if all uses were in branches that were never executed. I assume it's been fixed in the meantime but don't actually know.

palday: Eeek. That's a good reason to drop it though this isn't a particularly hot section of the code.

ararslan (approved):
I have basically 0 familiarity with this part of the code but what you have here looks reasonable at least. I imagine if there were downstream ramifications then other tests would be affected but that seems not to be the case. So... LGTM

PR #791, opened 2024-11-05
restore and save optsum for GLMM

### LINKED ISSUE #790: What is current best practice for saving a GLMM?
It seems that saveoptsum/restoreoptsum! only works for linear models. Is there a current solution for saving the fit of a GLMM? Thanks!

palday:
closes #790 

- [x] add entry in NEWS.md

### DIFF
--- a/src/serialization.jl
+++ b/src/serialization.jl
@@ -27,7 +27,21 @@ function restoreoptsum!(
     if length(setdiff(allowed_missing, keys(dict))) > 1 # 1 because :lowerbd
         @warn "optsum was saved with an older version of MixedModels.jl: consider resaving."
     end
+
+    # GLMM case with fast=slow
+    theta_beta_len = length(m.θ) + length(m.β)
+    if length(dict.initial) == theta_beta_len
+        resize!(ops.initial, theta_beta_len)
+        resize!(ops.final, theta_beta_len)
+        if length(ops.lowerbd) == length(m.θ)
+            prepend!(ops.lowerbd, fill(-Inf, length(m.β)))
+        end
+    end
+
     if any(ops.lowerbd .> dict.initial) || any(ops.lowerbd .> dict.final)
+        @debug "" ops.lowerbd
+        @debug "" dict.initial
+        @debug "" dict.final
         throw(ArgumentError("initial or final parameters in io do not satisfy lowerbd"))
     end
     for fld in (:feval, :finitial, :fmin, :ftol_rel, :ftol_abs, :maxfeval, :nAGQ, :REML)
@@ -68,6 +82,12 @@ function restoreoptsum!(m::LinearMixedModel{T}, filename; kwargs...) where {T}
     end
 end
 
+function restoreoptsum!(m::GeneralizedLinearMixedModel, fname; kwargs...)
+    restoreoptsum!(m.LMM, fname; kwargs...)
+    deviance!(m)
+    return m
+end
+
 """
     saveoptsum(io::IO, m::LinearMixedModel)
     saveoptsum(filename, m::LinearMixedModel)
@@ -84,5 +104,7 @@ function saveoptsum(filename, m::LinearMixedModel)
     end
 end
 
-# TODO: write methods for GLMM
+function saveoptsum(io, m::GeneralizedLinearMixedModel)
+    return saveoptsum(io, m.LMM)
+end
 # TODO, maybe: something nice for the MixedModelBootstrap
--- a/src/serialization.jl
+++ b/src/serialization.jl
@@ -10,9 +10,10 @@ function restoreoptsum!(m::MixedModel, filename; kwargs...)
     end
 end
 
-function restoreoptsum!(m::LinearMixedModel{T}, io::IO;
-    atol::Real=zero(T),
-    rtol::Real=atol > 0 ? zero(T) : √eps(T)) where {T}
+function restoreoptsum!(
+    m::LinearMixedModel{T}, io::IO; atol::Real=zero(T),
+    rtol::Real=atol > 0 ? zero(T) : √eps(T)
+) where {T}
     dict = JSON3.read(io)
     ops = restoreoptsum!(m.optsum, dict)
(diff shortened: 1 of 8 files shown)


src/serialization.jl
```diff
@@ -1,15 +1,76 @@
...
-) where {T}
+function restoreoptsum!(m::MixedModel, filename; kwargs...)
+    return open(filename, "r") do io
+        return restoreoptsum!(m, io; kwargs...)
+    end
+end
+
+function restoreoptsum!(m::LinearMixedModel{T}, io::IO;
+    atol::Real=zero(T),
+    rtol::Real=atol > 0 ? zero(T) : √eps(T)) where {T}
```
ararslan, line 15: Looks like the code style got out of whack. This suggestion makes it consistent with what it was before the addition of the method that takes a filename.
```suggestion
function restoreoptsum!(
    m::LinearMixedModel{T}, io::IO; atol::Real=zero(T),
    rtol::Real=atol > 0 ? zero(T) : √eps(T)
) where {T}
```

src/serialization.jl
```diff
@@ -1,15 +1,76 @@
...
+    dict = JSON3.read(io)
+    ops = restoreoptsum!(m.optsum, dict)
+    for (par, obj_at_par) in (:initial => :finitial, :final => :fmin)
+        if !isapprox(
+            objective(updateL!(setθ!(m, getfield(ops, par)))), getfield(ops, obj_at_par);
+            rtol, atol,
+        )
+            throw(
+                ArgumentError(
+                    "model m at $par does not give stored $obj_at_par within given tolerances"
```
ararslan, line 25:
```suggestion
                    "model at $par does not match stored $obj_at_par within atol=$atol, rtol=$rtol"
```
The addition of "m" here isn't informative as the method takes only a single model, plus it's a positional argument so its name isn't meaningfully user-facing. But it could be nice to parrot back the tolerances that were used for the failed comparison.

src/serialization.jl
```diff
@@ -1,15 +1,76 @@
...
+                ),
+            )
+        end
+    end
+    return m
+end
+
+function restoreoptsum!(m::GeneralizedLinearMixedModel{T}, io::IO;
+    atol::Real=zero(T),
+    rtol::Real=atol > 0 ? zero(T) : √eps(T)) where {T}
```
ararslan, line 35: Another formatting mishap it looks like
```suggestion
function restoreoptsum!(
    m::GeneralizedLinearMixedModel{T}, io::IO; atol::Real=zero(T),
    rtol::Real=atol > 0 ? zero(T) : √eps(T)
) where {T}
```

src/serialization.jl
```diff
@@ -1,15 +1,76 @@
...
+        if length(ops.lowerbd) == length(m.θ)
+            prepend!(ops.lowerbd, fill(-Inf, length(m.β)))
+        end
+        setpar! = setβθ!
+        varyβ = false
+    else # fast=true
+        setpar! = setθ!
+        varyβ = true
+        if length(ops.lowerbd) != length(m.θ)
+            deleteat!(ops.lowerbd, 1:length(m.β))
```
ararslan, line 54: I suppose it can never be the case that `length(ops.lowerbd) < length(m.β)` or that `ops.lowerbd` has indices that don't start at 1?

src/serialization.jl
```diff
@@ -1,15 +1,76 @@
...
+    end
+    restoreoptsum!(ops, dict)
+    for (par, obj_at_par) in (:initial => :finitial, :final => :fmin)
+        if !isapprox(
+            deviance(pirls!(setpar!(m, getfield(ops, par)), varyβ), dict.nAGQ),
+            getfield(ops, obj_at_par); rtol, atol,
+        )
+            throw(
+                ArgumentError(
+                    "model m at $par does not give stored $obj_at_par within given tolerances"
```
ararslan, line 65: Same suggestion as above
```suggestion
                    "model at $par does not match stored $obj_at_par within atol=$atol, rtol=$rtol"
```

src/serialization.jl
```diff
@@ -27,7 +88,11 @@ function restoreoptsum!(
     if length(setdiff(allowed_missing, keys(dict))) > 1 # 1 because :lowerbd
         @warn "optsum was saved with an older version of MixedModels.jl: consider resaving."
     end
+
     if any(ops.lowerbd .> dict.initial) || any(ops.lowerbd .> dict.final)
+        @debug "" ops.lowerbd
+        @debug "" dict.initial
+        @debug "" dict.final
```
ararslan, line 95: I don't know whether you separated them on purpose but FYI you can combine them into one `@debug`:
```suggestion
        @debug "" ops.lowerbd dict.initial dict.final
```

ararslan (approved):
I don't feel like I can give a super thoughtful review as I'm not familiar with this part of the codebase but here are some comments/suggestions in case they're useful, plus an approval since nothing seemed obviously wrong (as far as I could tell).

src/serialization.jl
ararslan: In case JuliaFormatter made this change, I think you might want to add
```toml
[compat]
JuliaFormatter = "1"
```
in `format/Project.toml`.

src/serialization.jl
palday: `lowerbd` is in a struct we control and it's always `Vector`, so the indices are fine.

There are two cases here:
- the "slow" optimization, where the parameter vector contains both beta and theta, so length(lowerbd) needs to be length(beta) + length(theta). 
- the "fast" optimization, where the parameter vector contains only theta, so length(lowerbd) needs to be length(theta). (In this case, there is an inner optimization step that finds the best beta for a given theta using PIRLS -- this will generally give the same answer or a very close one to doing free optimization over both, but there are a few pathological cases where the two methods can diverge, in which case the extra freedom of the slow optimization will yield more accurate results.)

The initial model creation initializes the lower bound to the length of theta because that is shared with the initialization for the linear mixed model (where only theta is included in optimization and beta is solved for directly).

Issue #794, opened 2024-11-27
Error in `profile(m)` mutates `m` and fails to clean up


yjunechoe:
I've found that sometimes `profile(m)` throws `ERROR: AssertionError: i ∈ iallowed || iszero(A[i, j])`, which mutates the model `m`.

MWE with a semi-synthesized portion of the data from my usecase:

**Setup**

```julia
using CSV
using MixedModels
x = CSV.read(download("https://gist.githubusercontent.com/yjunechoe/af4a8dd3e2b0343fe3818e93dba8c81d/raw/x.csv"), Table);
m = fit(MixedModel, @formula(y ~ x1 * x2 + (1 | g)), x);
m
```
```
Linear mixed model fit by maximum likelihood
 y ~ 1 + x1 + x2 + x1 & x2 + (1 | g)
   logLik   -2 logLik     AIC       AICc        BIC
  -860.9412  1721.8823  1733.8823  1733.9669  1763.3289

Variance components:
            Column    Variance  Std.Dev.
g        (Intercept)  0.0022759 0.0477069
Residual              0.3264997 0.5714015
 Number of obs: 1000; levels of grouping factors: 3

  Fixed-effects parameters:
─────────────────────────────────────────────────────────
                      Coef.  Std. Error       z  Pr(>|z|)
─────────────────────────────────────────────────────────
(Intercept)      -1.3144      0.0726041  -18.10    <1e-72
x1: a2            1.1196      0.0963029   11.63    <1e-30
x2: b2           -0.0839275   0.0876926   -0.96    0.3385
x1: a2 & x2: b2   0.189803    0.105979     1.79    0.0733
─────────────────────────────────────────────────────────
```

**Surprising behavior**

```julia
profile(m, threshold=3);
# ERROR: AssertionError: i ∈ iallowed || iszero(A[i, j])
```

<details>
<summary>Full stacktrace</summary>

```julia
ERROR: AssertionError: i ∈ iallowed || iszero(A[i, j])
Stacktrace:
  [1] _check_locality(A::StaticArraysCore.SMatrix{3, 2, Float64, 6}, corner::Symbol, allowed_nonzeros_per_column::Int64)
    @ BSplineKit.Recombinations C:\Users\jchoe\.julia\packages\BSplineKit\OQJIx\src\Recombinations\matrices.jl:133
  [2] _RecombineMatrix
[...]

PR #795, opened 2024-12-09
add finally block for restoring original model in profiling

palday:
addresses part of #794

- [x] add entry in NEWS.md

### DIFF
--- a/src/profile/profile.jl
+++ b/src/profile/profile.jl
@@ -51,7 +51,7 @@ function profile(m::LinearMixedModel; threshold=4)
         MixedModelProfile(m, Table(val.tbl), val.fwd, val.rev)
     catch ex
         @error "Exception occurred in profiling; aborting..."
-        rethrow(ex)
+        rethrow()
     finally
         objective!(m, final)   # restore the parameter estimates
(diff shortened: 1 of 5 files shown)


src/profile/profile.jl
```diff
@@ -35,23 +35,30 @@ value of ζ exceeds `threshold`.
...
+        copyto!(m.optsum.final, final)
+        m.optsum.fmin = objective!(m, final)
+        for s in filter(s -> startswith(string(s), 'θ'), keys(first(val.tbl)))
+            profileθj!(val, s, tc; threshold)
+        end
+        profileσs!(val, tc)
+        MixedModelProfile(m, Table(val.tbl), val.fwd, val.rev)
+    catch ex
+        @error "Exception occurred in profiling; aborting..."
+        rethrow(ex)
```
ararslan, line 54:
```suggestion
        rethrow()
```
Functionally equivalent but this form is preferred when you don't need to throw a new exception in place of the one that was caught (and in that case, `throw` is preferred over `rethrow`).

ararslan (approved):

PR #796, opened 2024-12-10
attempt to find better starting values for GLMM when defaults aren't positive semidefinite

### LINKED ISSUE #792: Non-positive definite matrix when fitting nested GLMM
I am trying to understand some issues I've encountered when trying to fit some models or a psychology experiment on memory where there is a lot of nested structure. 
For background, the goal is to estimate the effect of position within a list on the probability of recall. Each subject in the experiment participates in multiple sessions with multiple lists. The following code creates an example dataset:

```
using MixedModels, DataFrames, Random
Random.seed!(0);
df = allcombinations(DataFrame, "subject" => 1:100, 
    "session" => 1:6, "list" => 1:25, "serialpos" => 1:12)
df[!, :recalled] = rand(nrow(df))
df[!, :recalled] = df[!, :recalled] .> .5
```

Then to fit the model I use the following code:
```
serialpos_contrasts = Dict(
    :serialpos => DummyCoding()
)

m = fit(MixedModel, 
    form, 
    df, Bernoulli(), 
    contrasts=serialpos_contrasts; 
    progress=false);
```

I fit several models, some of which fit without problems and one that has problems. The following all fit fine:
_(m1)_ `form => @formula(recalled ~ serialpos + (1 | subject) + (1 | subject & session) + (1 | subject & session & list))`
_(m2)_ `recalled ~ serialpos + zerocorr(serialpos | subject)`
_(m3)_ `recalled ~ serialpos + (1 | subject) + (1 | serialpos & subject) + (1 | subject & session)`

However, I run into issues when trying to fit the following model, with the subject effects varying and random intercepts by session:
_(m4)_ `recalled ~ serialpos + zerocorr(serialpos | subject) + (1 | subject & session))`
This results in this error:
```
ERROR: LoadError: PosDefException: matrix is not positive definite; Factorization failed.
```

I realize this has been partially addressed before (e.g. https://github.com/JuliaStats/MixedModels.jl/issues/349#issue-668432291) but I'm not really sure I fully understand why _(m4)_ has an error -- _(m1)_ should have more parameters and _(m3)_ should be nearly equivalent to _(m4)_ (although I realize the design matrices are slightly different). 
I thought perhaps the correlation structure might be an issue so I tried fitting two models with fulldummy which seems more straight forward.

_(m5)_ `recalled ~ serialpos + zerocorr(fulldummy(serialpos) | subject)`
_(m6)_ `recalled ~ serialpos + zerocorr(fulldummy(serialpos) | subject) + (1 | subject & session)`

_(m5)_ fit fine and _(m6)_ ran into the positive definite error.
 
Is there any good intuition for what is going on here?

palday:
closes #792 

- [x] add entry in NEWS.md

### DIFF
--- a/.github/workflows/current.yml
+++ b/.github/workflows/current.yml
@@ -38,7 +38,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - uses: julia-actions/julia-processcoverage@v1
-      - uses: codecov/codecov-action@v4
+      - uses: codecov/codecov-action@v5
         with:
           flags: current
           files: lcov.info
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -36,7 +36,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - uses: julia-actions/julia-processcoverage@v1
-      - uses: codecov/codecov-action@v4
+      - uses: codecov/codecov-action@v5
         with:
           flags: minimum
           files: lcov.info
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         julia-version: [min]
         julia-arch: [x64]
-        os: [ubuntu-22.04, macos-12, windows-2019]
+        os: [ubuntu-22.04, macos-13, windows-2019]
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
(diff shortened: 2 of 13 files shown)


dmbates (commented):
On reading the section of `src/generalizedlinearmixedmodel.jl` that is modified it seems that a lot of work is being done to handle the case of `constresponse`.  From the comments it seems that the reason for handling `constresponse` is for convenience in starting simulations.  Is that correct?  Would it be better to just error on `constresponse` with a message to come up with more realistic starting values?

dmbates:
I see that I misread the code.  You only try the fix-up if `constresponse` is `false`.  Never mind.

dmbates (approved):

PR #801, opened 2025-01-22
populate optsum in prfit!

palday:
- [x] add entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+MixedModels v4.29.1 Release Notes
+==============================
+- Populate `optsum` in `prfit!` call. [#800]
+
 MixedModels v4.29.0 Release Notes
 ==============================
 - Testbed for experimental support for using PRIMA as an optimization backend introduced via the experimental `prfit!` function. [#799]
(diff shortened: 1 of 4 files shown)


dmbates:
The documentation says that the `optimizer` field contains a dummy value to indicate PRIMA but it seems not to have been changed from :LN_BOBYQA

Did you intend to modify it?

It also looks as if optsum.finitial is not reset.

dmbates:
BTW, thanks for the very rapid response on this.  I was in the library downstairs then came up to our condo and by the time I got here you had a complete PR created.

palday:
> The documentation says that the `optimizer` field contains a dummy value to indicate PRIMA but it seems not to have been changed from :LN_BOBYQA
> 
> Did you intend to modify it?
> 
> It also looks as if optsum.finitial is not reset.


Good catch on both fronts -- I think I've addressed them now.

dmbates:
Two small problems, which I am happy to look at if you wish.

The progress bar information is overwritten by the display of the fitted model.  Was there an extra println() in the `fit!` function to prevent that overwriting?

The optsum.fitlog doesn't include the evaluation at the initial parameter estimates in prfit! but it does in fit!

Just let me know if you would prefer that I look into these changes.

palday:
@dmbates I think I got those fixed now!

dmbates (approved):
Thanks for all the work on this.

PR #802, opened 2025-01-25
refactor optimization to make it easy to add new backends

### LINKED ISSUE #619: Remove repeated initial entry in fitlog
if you set `thin = 1` it is
```julia
julia> ds = MixedModels.dataset(:dyestuff)
Arrow.Table with 30 rows, 2 columns, and schema:
 :batch  String
 :yield  Int16

julia> contrasts = Dict{Symbol, AbstractContrasts}(:batch => Grouping())
Dict{Symbol, AbstractContrasts} with 1 entry:
  :batch => Grouping()

julia> thin = 1
1

julia> dsm01 = let
           form = @formula yield ~ 1 + (1|batch)
           fit(MixedModel, form, ds; contrasts, thin)
       end
 yield ~ 1 + (1 | batch)
   logLik   -2 logLik     AIC       AICc        BIC    
  -163.6635   327.3271   333.3271   334.2501   337.5307

Variance components:
            Column    Variance Std.Dev.
batch    (Intercept)  1388.3332 37.2603
Residual              2451.2501 49.5101

  Fixed-effects parameters:
(Intercept)  1527.5     17.6946  86.33    <1e-99

julia> first(dsm01.optsum.fitlog, 3)
3-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.0], 327.76702162461663)
 ([1.0], 327.76702162461663)
 ([1.75], 331.03619322245146)
```

_Originally posted by @dmbates in https://github.com/JuliaStats/MixedModels.jl/issues/615#issuecomment-1133076508_

Note that this doesn't occur when calling `refit!` as currently done in the tests.

### LINKED ISSUE #689: Refactor NLopt boilerplate into an easy function call
I'm wondering if we can re-factor (not in this PR) so that all the NLopt boilerplate is handled by a single call. That would be also make it easier to one day support other optimization backends, say the MathOptInterface or the like.

_Originally posted by @palday in https://github.com/JuliaStats/MixedModels.jl/pull/639#discussion_r1143527299_

### LINKED ISSUE #742: Switch to PRIMA.bobyqa for optimization?
See the discussion in https://github.com/lme4/lme4/issues/744 regarding `PRIMA.bobyqa` in linear mixed models.  It may be even more useful to switch to `PRIMA.bobyqa` for GLMM optimization because that version of `bobyqa` allows for a `scale` argument which can be very helpful when simultaneously optimizing the fixed-effects and variance-component parameters.

If we make this change should we retain the ability to use the NLopt optimizers or allow only the PRIMA optimizers?  It would be good to retain NLopt and hence compatibility with earlier versions but it could get messy with the optsum structure and with tests - not to mention the potential for some really ugly and confusing code.

palday:
closes #619, closes #689, closes #742

- [x] add entry in NEWS.md
- [x] refactor LMM
- [x] refactor GLMM
- [x] refactor `prfit!` to take advantage of this

### DIFF
--- a/ext/MixedModelsPRIMAExt.jl
+++ b/ext/MixedModelsPRIMAExt.jl
@@ -1,7 +1,9 @@
 module MixedModelsPRIMAExt
 
 using MixedModels
-using MixedModels: ProgressMeter, ProgressUnknown, objective!
+using MixedModels: Statistics
+using MixedModels: ProgressMeter, ProgressUnknown, objective!, _objective!
+using LinearAlgebra: PosDefException
 using PRIMA: PRIMA
 
 function MixedModels.prfit!(m::LinearMixedModel;
@@ -63,14 +65,13 @@ function MixedModels.optimize!(m::LinearMixedModel, ::PRIMABackend;
     optsum.fmin = info.fx
     optsum.returnvalue = Symbol(info.status)
     _check_prima_return(info)
-
     return optsum.final, optsum.fmin
 end
 
-function optimize!(m::GeneralizedLinearMixedModel, ::PRIMABackend;
-                  progress::Bool=true, thin::Int=tyepmax(Int),
-                  fast::Bool=false, verbose::Bool=false, nAGQ=1,
-                  kwargs...)
+function MixedModels.optimize!(m::GeneralizedLinearMixedModel, ::PRIMABackend;
+                               progress::Bool=true, thin::Int=tyepmax(Int),
+                               fast::Bool=false, verbose::Bool=false, nAGQ=1,
+                               kwargs...)
     optsum = m.optsum
     fitlog = optsum.fitlog
     prog = ProgressUnknown(; desc="Minimizing", showspeed=true)
@@ -98,9 +99,24 @@ function optimize!(m::GeneralizedLinearMixedModel, ::PRIMABackend;
     empty!(fitlog)
     push!(fitlog, (copy(optsum.initial), optsum.finitial))
     maxfun = optsum.maxfeval > 0 ? optsum.maxfeval : 500 * length(optsum.initial)
+    scale = if fast
+        nothing
+    else
+        # scale by the standard deviation of the columns of the fixef model matrix
+        # when including the fixef in the nonlinear opt
+        sc = [map(std, eachcol(modelmatrix(m))); fill(1, length(m.θ))]
+        for (i, x) in enumerate(sc)
+            # for nearly constant things, e.g. intercept, we don't want to scale to zero...
+            # also, since we're scaling the _parameters_ and not the data,
+            # we need to invert the scale
+            sc[i] = ifelse(iszero(x), one(x), inv(x))
+        end
+        sc
+    end
     info = prima_optimizer!(Val(optsum.optimizer), obj, optsum.final;
                             xl=optsum.lowerbd, maxfun,
-                            optsum.rhoend, optsum.rhobeg)
+                            optsum.rhoend, optsum.rhobeg,
+                            scale)
     ProgressMeter.finish!(prog)
 
     optsum.feval = info.nf
@@ -108,7 +124,7 @@ function optimize!(m::GeneralizedLinearMixedModel, ::PRIMABackend;
     optsum.returnvalue = Symbol(info.status)
     _check_prima_return(info)
 
-    return xmin, fmin
+    return optsum.final, optsum.fmin
 end
 
 function _check_prima_return(info::PRIMA.Info)
--- a/ext/MixedModelsPRIMAExt.jl
+++ b/ext/MixedModelsPRIMAExt.jl
@@ -15,8 +15,8 @@ function MixedModels.prfit!(m::LinearMixedModel;
 end
 
 prima_optimizer!(::Val{:bobyqa}, args...; kwargs...) = PRIMA.bobyqa!(args...; kwargs...)
-prima_optimizer!(::Val{:cobyla}, args...; kwargs...) = PRIMA.cobyla(args...; kwargs...)
-prima_optimizer!(::Val{:lincoa}, args...; kwargs...) = PRIMA.lincoa(args...; kwargs...)
+prima_optimizer!(::Val{:cobyla}, args...; kwargs...) = PRIMA.cobyla!(args...; kwargs...)
+prima_optimizer!(::Val{:lincoa}, args...; kwargs...) = PRIMA.lincoa!(args...; kwargs...)
 
 push!(MixedModels.OPTIMIZATION_BACKENDS, :prima)
 
@@ -61,6 +61,7 @@ function MixedModels.optimize!(m::LinearMixedModel, ::PRIMABackend; progress::Bo
     optsum.feval = info.nf
     optsum.fmin = info.fx
     optsum.returnvalue = Symbol(info.status)
+    _check_prima_return(info)
 
     return optsum.final, optsum.fmin
 end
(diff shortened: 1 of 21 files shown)


dmbates (approved):
Wow!  This is amazing.  I am very glad to see these changes.

Thanks for disabling the `thin` named argument.  That wasn't one of my better ideas.

I don't think I will be able to do a line-by-line check on the changes (still having trouble with my vision) but I would be happy to approve the changes and you can merge the PR whenever you wish.

Is the intent to replace `MixedModels.prfit!(m::LinearMixedModel,...)` by `MixedModels.optimize!(m::LinearMixedModel, Val(:prima); ...)`?  It isn't clear to me yet how I would use the PRIMA backend without calling `prfit!` or instantiating a `LinearMixedModel` object and manually changing `m.optsum.backend` and `m.optsum.optimizer`.

I had expected that
```
using PRIMA, MixedModels
MixedModels.OPTIMIZATION_BACKENDS
```
would return a vector of two Symbols but it is only one Symbol.  Does the `:prima` backend need to be called before it is added?  I see in `ext/MixedModelsPRIMAExt.jl` that a value is `push!`ed onto `MixedModels.OPTIMIZATION_BACKENDS` but it doesn't seem to be then when I examine that object.

When I invoke `prfit!` on a linear mixed model I sometimes get a warning from the Fortran code in libprima like
```julia
julia> MixedModels.prfit!(m1)

Warning: BOBYQA: X0 is revised so that the distance between X0 and the inactive bounds is at least RHOBEG; set HONOUR_X0 to .TRUE. if you prefer to keep X0 unchanged.
```

Perhaps the `rhobeg` value should be set to a number less than 1?

src/generalizedlinearmixedmodel.jl
```diff
@@ -540,6 +517,30 @@ function StatsAPI.loglikelihood(m::GeneralizedLinearMixedModel{T}) where {T}
...
+function objective!(m::GeneralizedLinearMixedModel; fast=false, kwargs...)
+    return x -> _objective!(m, x, Val(fast); kwargs...)
+end
+
+function objective!(m::GeneralizedLinearMixedModel{T}, x; fast=false, kwargs...) where {T}
+    return _objective!(m, x, Val(fast); kwargs...)
+end
+
+# normally, it doesn't make sense to move a simple branch to dispatch
+# HOWEVER, this wins up getting called in optimization a lot and
```
dmbates, line 530: wins -> winds ?

palday:
> Wow! This is amazing. I am very glad to see these changes.
> 
> Thanks for disabling the `thin` named argument. That wasn't one of my better ideas.
> 
> I don't think I will be able to do a line-by-line check on the changes (still having trouble with my vision) but I would be happy to approve the changes and you can merge the PR whenever you wish.
> 
> Is the intent to replace `MixedModels.prfit!(m::LinearMixedModel,...)` by `MixedModels.optimize!(m::LinearMixedModel, Val(:prima); ...)`? It isn't clear to me yet how I would use the PRIMA backend without calling `prfit!` or instantiating a `LinearMixedModel` object and manually changing `m.optsum.backend` and `m.optsum.optimizer`.

There is a "secret" kwarg that you can use:

```julia
julia> using MixedModels, PRIMA
Precompiling MixedModelsPRIMAExt...
  1 dependency successfully precompiled in 8 seconds. 137 already precompiled.

julia> fit(MixedModel, @formula(reaction ~ 1 + days + (1+days|subj)), MixedModels.dataset(:sleepstudy); optimizer=:bobyqa, backend=:prima)
 reaction ~ 1 + days + (1 + days | subj)
   logLik   -2 logLik     AIC       AICc        BIC    
  -875.9697  1751.9393  1763.9393  1764.4249  1783.0971

Variance components:
            Column    Variance Std.Dev.   Corr.
subj     (Intercept)  565.51525 23.78056
         days          32.68219  5.71683 +0.08
Residual              654.94105 25.59182
 Number of obs: 180; levels of grouping factors: 18

  Fixed-effects parameters:
(Intercept)  251.405      6.63228  37.91    <1e-99
days          10.4673     1.50224   6.97    <1e-11

julia> ans.optsum
Initial parameter vector: [1.0, 0.0, 1.0]
Initial objective value:  1784.642296192471

Backend:                  prima
Optimizer:                bobyqa
Lower bounds:             [0.0, -Inf, 0.0]
rhobeg:                   1.0
rhoend:                   1.0e-6
maxfeval:                 -1

Function evaluations:     69
xtol_zero_abs:            0.001
ftol_zero_abs:            1.0e-5
Final parameter vector:   [0.9292253632482519, 0.018165717899275697, 0.22264540632838237]
Final objective value:    1751.9393444632112
Return code:              SMALL_TR_RADIUS
```

I haven't advertised them yet because I've found a few cases where something goes wrong with PRIMA. For example, on Intel machines, the VSCode REPL + PRIMA causes a `StackOverflowError` even for a very minimal optimization problem. Running in a REPL not tied to VSCode works just fine though. 

> 
> I had expected that
> 
> ```
> using PRIMA, MixedModels
> MixedModels.OPTIMIZATION_BACKENDS
> ```
> 
> would return a vector of two Symbols but it is only one Symbol. Does the `:prima` backend need to be called before it is added? I see in `ext/MixedModelsPRIMAExt.jl` that a value is `push!`ed onto `MixedModels.OPTIMIZATION_BACKENDS` but it doesn't seem to be then when I examine that object.

Huh, this is surprising to me -- I'll take a look.

>
> When I invoke `prfit!` on a linear mixed model I sometimes get a warning from the Fortran code in libprima like
> 
> ```julia
> julia> MixedModels.prfit!(m1)
> 
> Warning: BOBYQA: X0 is revised so that the distance between X0 and the inactive bounds is at least RHOBEG; set HONOUR_X0 to .TRUE. if you prefer to keep X0 unchanged.
> ```
> 
> Perhaps the `rhobeg` value should be set to a number less than 1?

PRIMA's default `rhobeg` is 1 and I don't yet understand well enough what rho represents to have a feeling for a good starting value.

src/generalizedlinearmixedmodel.jl
palday: yes!

```suggestion
# HOWEVER, this winds up getting called in optimization a lot and
```

ext/MixedModelsPRIMAExt.jl
```diff
@@ -1,27 +1,41 @@
...
+using MixedModels
+using MixedModels: Statistics
+using MixedModels: ProgressMeter, ProgressUnknown, objective!, _objective!
+using LinearAlgebra: PosDefException
 using PRIMA: PRIMA
 
+function __init__()
+    push!(MixedModels.OPTIMIZATION_BACKENDS, :prima)
+    return nothing
+end
```
palday, line 12: @dmbates I had to wrap this in an `__init__` so that it happens at load time and not precompile time.

dmbates: Makes sense.

dmbates:
I had worked out that I should add `backend=:prima` in the call to `fit(MixedModel, ...)` but I didn't also add `optimizer=:bobyqa`.  The error message just told me there was a missing method.

palday:
I'm going to merge later this afternoon -- I think us using this is the best way to discover issues with this design. The documented, public facing components aren't changing, so even further internal reorganization isn't technically breaking.

dmbates:
Sounds good to merge this.  I'll just turn off notifications while I am watching football today 😄

kliegl:
Thank you very much!

PR #805, opened 2025-02-24
added function aliases that dont contain unicode characters for convenience

ajinkya-k:
this (draft) PR adds `settheta!` and `profilesigma` as aliases for `setθ!` and `profileσ` respectively. This is useful when displaying code in LaTeX. One place where this is really needed is arxiv preprints, as arxiv does not (currently) accept xelatex or lualatex, and making unicode symbols work in pdflatex adds a lot of burden. 

Tagging @dmbates as he might have suggestions/comments.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+MixedModels v4.31.0 Release Notes
+==============================
+- Added aliases `settheta!` and `profilesigma` for the functions `setθ!` and `profileσ` respectively
+
 MixedModels v4.30.0 Release Notes
 ==============================
 - Refactor calls to backend optimizer to make it easier to add and use different optimization backends.
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MixedModels"
 uuid = "ff71e718-51f3-5ec2-a782-8ffcbfa3c316"
 author = ["Phillip Alday <me@phillipalday.com>", "Douglas Bates <dmbates@gmail.com>"]
-version = "4.30.1"
+version = "4.31.0"
 
 [deps]
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
--- a/src/MixedModels.jl
+++ b/src/MixedModels.jl
@@ -212,8 +212,8 @@ include("prima.jl")
 
 
 # aliases with non-unicode function names
-settheta! = setθ!
-profilesigma = profileσ
+const settheta! = setθ!
+const profilesigma = profileσ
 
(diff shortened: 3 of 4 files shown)


src/MixedModels.jl
```diff
@@ -208,6 +210,11 @@ include("profile/profile.jl")
 include("nlopt.jl")
 include("prima.jl")
 
+
+# aliases with non-unicode function names
+settheta! = setθ!
+profilesigma = profileσ
```
palday, line 216:
```suggestion
const settheta! = setθ!
const profilesigma = profileσ
```

Setting these `const` helps with all sorts of little performance things.

ajinkya-k:
Added `const`. Sorry for the uninformative commit message, I was trying out the GitHub "commit changes" website interface and accidentally hit commit before modifying the message. Do you want me to change it?

ajinkya-k:
Thanks. I'll add the tests and a NEWS.md entry in some time

ajinkya-k:
okay i bumped up the version number to `4.30.2` and added a `NEWS.md` entry, but there wasnt any 4.30.1 entry in there FYI. Feel free to make/suggest changes to the news or project toml

NEWS.md
```diff
@@ -1,3 +1,7 @@
+MixedModels v4.30.2 Release Notes
```
palday, line 1:
```suggestion
MixedModels v4.31.0 Release Notes
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "MixedModels"
 uuid = "ff71e718-51f3-5ec2-a782-8ffcbfa3c316"
 author = ["Phillip Alday <me@phillipalday.com>", "Douglas Bates <dmbates@gmail.com>"]
-version = "4.30.1"
+version = "4.30.2"
```
palday, line 4:
```suggestion
version = "4.31.0"
```

this is technically a new feature, so under SemVer rules, it gets a minor version bump. Patch bumps are for bug fixes.

(Also this is why there are often no NEWS entries for patch releases -- often they're addressing tiny corner case bugs, so we don't feel the need to explain much there. That said, if we ever discuss a critical, inaccurate results bug, then I would include that in NEWS.)

palday (commented):
one more little thing, then I'll be ready to get it merged, so feel free to take it out of draft!

ajinkya-k:
fixed the version number and (force) pushed! I wasnt sure what version number to put in so I just put something in hoping its right and if not get a suggestion

palday (approved):

PR #808, opened 2025-03-06
Using VitePress for documentation

ajinkya-k:
This is a speculative draft PR. I think `VitePress` based docs look much better than the default julia-docs style. While mostly an aesthetic choice, one particular issue with the current docs style is the default fonts chosen by `Documenter.jl` render very poorly for me when using an external monitor with a Mac. So I recreated the documentation for this package using `DocumenterVitepress.jl` and it's deployed for preview at: https://ajinkya-k.github.io/MixedModels.jl/ . I cant setup CI so I manually cobbled together the rendered files on the `gh-pages` branch.  There are still quite a few kinks to work out but I think it looks great! See for example the Model constructors page for a well rendered page: https://ajinkya-k.github.io/MixedModels.jl/constructors

Let me know what you all think!


ajinkya-k:
I think this doc is now good to go!

ajinkya-k:
is it possible to deploy this only for the dev version to elicit feedback, or deploy an experimental version without taking down the current documentation?

ajinkya-k:
@palday was that merge commit just because of a conflict? It looks to me like the final content of the merge commit is same as the previous commit.

ajinkya-k:
@palday what else needs to be done for this to be merged?

ajinkya-k:
I updated the preview doc, take a look: https://ajinkya-k.github.io/MixedModels.jl/api

docs/make.jl
```diff
@@ -2,23 +2,31 @@ using Documenter
...
+        ],
+        "api.md"
     ],
+    format = DocumenterVitepress.MarkdownVitepress(
+        repo = "https://github.com/JuliaStats/MixedModels.jl",
+    ),
+    clean = true
 )
 
 deploydocs(;
```
asinghvi17, line 29: This needs to be DocumenterVitepress.deploydocs in DV v0.2

PR #810, opened 2025-03-16
added `lmm` and `glmm` as convenience wrappers for `fit` 

### LINKED ISSUE #809: Add `lmm` (or something similar) as a convenience fitting function
Maybe this is a bad idea, but I think there should be an interface function say `lmm(frm::Formula, ...)` similar to `lm` and `glm` as a thin wrapper around `fit` because it's more concise (and cleaner) than writing `fit(LinearMixerModel,....)`

Has this already been considered? I'm happy to implement this thin wrapper if it would be useful.

ajinkya-k:
closes #809 

- [x] add entry in NEWS.md

- [x] figure out new minor version no 
- [x] actually bump minor version


ajinkya-k:
can someone help me with this warning: 
```julia-repl
┌ MixedModels
│  WARNING: Method definition lmm(StatsModels.FormulaTerm{L, R} where R where L, Any) in module MixedModels at /home/runner/work/MixedModels.jl/MixedModels.jl/src/linearmixedmodel.jl:230 overwritten at /home/runner/work/MixedModels.jl/MixedModels.jl/src/linearmixedmodel.jl:231.
│  ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
└  
```

but the same doesn't seem to happen with `StatsAPI.fit(LinearMixedModel, ) see lines 194-218 of `src/linearmixedmodel.jl`

palday:
MixedModels.jl had something like this in the past, but we dropped that interface. Maybe @dmbates can comment more on the why.

To actually add it, you can simply do:

```julia
lmm(args...; kwargs...) = fit(LinearMixedModel, args...; kwargs...)
glmm(args...; kwargs...) = fit(GeneralizedLinearMixedModel, args...; kwargs...)
```

ajinkya-k:
Okay i will add the tests. I wanted some comments on what to write in the docstring

ajinkya-k:
@dmbates Added a test for `REML = true`. It only makes sure the returned model `isa LinearMixedModel` . would it be worth it to test maybe the fixed effects or deviance too?

src/linearmixedmodel.jl
```diff
@@ -213,10 +213,21 @@ function StatsAPI.fit(::Type{LinearMixedModel},
...
 
+"""
+    lmm(f::FormulaTerm, tbl; kwargs...)
+    lmm(
+        f::FormulaTerm, tbl;
+        wts=[], contrasts=Dict{Symbol,Any}(), σ=nothing, amalgamate=true,
+        kwargs...)
+
+    Aliases for the fit(LinearMixedModel, ...) functions
+"""
```
palday, line 228:
```suggestion
"""
    lmm(args...; kwargs...)

Convenience wrapper for `fit(LinearMixedModel, args...; kwargs...)`.

See [`fit`](@ref) for more information.
"""
```

Then I would add an example somewhere in the narrative docs, maybe in the introductory sections.

ajinkya-k:
image

The lines with the yellow side bar are not being rendered in the api docs at all. Am I doing something wrong with the formatting?

src/linearmixedmodel.jl
ajinkya-k: thanks!

palday:
> image
> The lines with the yellow side bar are not being rendered in the api docs at all. Am I doing something wrong with the formatting?

you need to explicitly add them in `docs/src/api.md`.

palday:
> [ ] figure out new version no

You're adding a feature in a post 1.0 release without breaking anything, so it's a minor (major.minor.patch) version bump according to SemVer.

ajinkya-k:
> > [ ] figure out new version no
> 
> You're adding a feature in a post 1.0 release without breaking anything, so it's a minor (major.minor.patch) version bump according to SemVer.

right, I meant figure out the new minor version no.

palday:
> @palday I did add `lmm` in `docs/src/api.md` like this: image

you haven't committed and pushed that change yet 😉

ajinkya-k:
@palday I meant on local, lol. I was rendering the docs locally and running `LiveServer`. I just did a clean build of the documentation and it worked

ajinkya-k:
Okay! I think this is good to go!

ajinkya-k:
The reference 
```
[`fit`](@ref)
```
gives an error because there is no doc-string for `fit`. That error went away if I use `fit!` instead but `fit!` does not document some keywords such as `wts`. Should i add a docstring to `fit`?

palday:
> The reference
> 
> ```
> [`fit`](@ref)
> ```
> 
> gives an error because there is no doc-string for `fit`. That error went away if I use `fit!` instead but `fit!` does not document some keywords such as `wts`. Should i add a docstring to `fit`?

I'll push a change shortly that adds a secondary link to the type constructor and that should cover everything.

ajinkya-k:
I don't think we need to display the output in the Quick Start section but I couldn't hide it so I left it as is. Is it possible to hide output without using a ; in an example block?

palday:
If you don't want to display output, do you even want the code executed? If you're not executing the code, then a normal triple-backtick codeblock would work.

That said, I don't think there's harm in showing the output.

ajinkya-k:
Let's keep the output

palday (approved):

PR #814, opened 2025-03-18
LikelihoodRatioTest extends StatsAPI.HypothesisTest

### LINKED ISSUE #683: Have `LikelihoodRatioTest` extend `StatsAPI. HypothesisTest`
xref https://github.com/JuliaStats/StatsAPI.jl/releases/tag/v1.6.0

palday:
Closes #683

- [x] add entry in NEWS.md

### DIFF
--- a/src/likelihoodratiotest.jl
+++ b/src/likelihoodratiotest.jl
@@ -27,11 +27,11 @@ function Base.propertynames(lrt::LikelihoodRatioTest, private::Bool=false)
 end
 
 """
-    StatsAPI.pvalue(lrt::LikelihoodRatioTest)
+    pvalue(lrt::LikelihoodRatioTest)
 
-Extract the p-value associated with the a likelihood ratio test.
+Extract the p-value associated with a likelihood ratio test.
 
-For `LikelihoodRatioTest`s containing more one model comparison, i.e. more than two models,
+For `LikelihoodRatioTest`s containing more than one model comparison, i.e. more than two models,
 this throws an error because it is unclear which p-value is desired.
 
 To get p-values for multiple tests, use `lrt.pvalues`.
@@ -46,7 +46,7 @@ function StatsAPI.pvalue(lrt::LikelihoodRatioTest)
         )
     end
 
-    return only(lrt.pvalues)
+    return only(pvalues)
 end
 
 function Base.getproperty(lrt::LikelihoodRatioTest, s::Symbol)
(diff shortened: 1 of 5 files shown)


src/likelihoodratiotest.jl
```diff
@@ -26,6 +26,29 @@ function Base.propertynames(lrt::LikelihoodRatioTest, private::Bool=false)
     return (:deviance, :formulas, :models, :pvalues, :tests)
 end
 
+"""
+    StatsAPI.pvalue(lrt::LikelihoodRatioTest)
```
ararslan, line 30:
```suggestion
    pvalue(lrt::LikelihoodRatioTest)
```
The qualification in the docstring seems to suggest that it's how the user should expect to call the function, but StatsAPI is not intended to be directly user-facing. Since we're also re-exporting `pvalue`, the StatsAPI bit is irrelevant anyway.

src/likelihoodratiotest.jl
```diff
@@ -26,6 +26,29 @@ function Base.propertynames(lrt::LikelihoodRatioTest, private::Bool=false)
...
+
+Extract the p-value associated with the a likelihood ratio test.
```
ararslan, line 32:
```suggestion
Extract the p-value associated with a likelihood ratio test.
```

src/likelihoodratiotest.jl
```diff
@@ -26,6 +26,29 @@ function Base.propertynames(lrt::LikelihoodRatioTest, private::Bool=false)
...
+
+For `LikelihoodRatioTest`s containing more one model comparison, i.e. more than two models,
```
ararslan, line 34:
```suggestion
For `LikelihoodRatioTest`s containing more than one model comparison, i.e. more than two models,
```

src/likelihoodratiotest.jl
```diff
@@ -26,6 +26,29 @@ function Base.propertynames(lrt::LikelihoodRatioTest, private::Bool=false)
...
+    pvalues = lrt.pvalues
+    if length(pvalues) > 1
+        throw(
+            ArgumentError(
+                "Cannot extract **only one** p-value from a multiple test result."
+            ),
+        )
+    end
+
+    return only(lrt.pvalues)
```
ararslan, line 49:
```suggestion
    return only(pvalues)
```
You defined a local variable for this already, might as well reuse it here

PR #821, opened 2025-03-30
Allow TriangularRFP for diagonal blocks of L

### LINKED ISSUE #813: Potentual issue with `rankUpdate!` dosctring
According to the `rankUpdate` docstring: 
https://github.com/JuliaStats/MixedModels.jl/blob/7b3719136f15256428e7db948a34fb2e4fcf5aca/src/linalg/rankUpdate.jl#L6

the update is `C <-  α*A'A + β*C` but while debugging I had the following: C is of size `(6,6)` and A is of size `(6,24)` which means the update is `C <-  α*AA' + β*C` (note its the reverse order for A and its transpose). 

Image

Am I understanding something incorrectly or should the docstring be updated?

dmbates:
- Allow for blocks on the diagonal of L to be `TriangularRFP` from https://github.com/JuliaLinearAlgebra/RectangularFullPacked.jl
- In the process of making these changes I found that I needed to be more explicit about when a diagonal block is being treated as `Hermitian` and stored in the lower triangle (`copyscaleinflate!` and `rankUpdate!`) and when it is being lower triangular (after `cholUnblocked!` and in the `rdiv!` call).
- I had previously been playing somewhat fast and loose with these distinctions and just passing around `Matrix` objects for the dense case but that can trip you up.
- Initial timings showed a considerable speed penalty for `TriangularRFP`.  This is more of a slowdown in the `rankUpdate!` method, which involves many calls to `setindex!`, rather than the `cholUnblocked!` method.
- By rewriting the `rankUpdate!` method to use linear indexing on the trapezoidal part of the RFP representation we were able to reduce the speed penalty.
- A further enhancement is to sort the levels of the grouping factor associated with the RFP diagonal block by decreasing order of incidence so that more of the updates occur in the trapezoidal part.  This is not guaranteed to be optimal but it is a good heuristic.
- There is a new named argument, `RFPthreshold`, in the `LinearMixedModel` constructor (but not yet in `MixedModel`).
- Right now this has a `[ci skip]` designation because I haven't updated all the tests to the more stringent requirements on `Hermitian` and `Triangular` types.
- Tests should now pass but coverage needs to be addressed.
- Fixes #813


dmbates:
Some timings on my laptop
```julia
(jl_6i3bqi) pkg> st
Status `/private/var/folders/d9/t94kd5m50rl1r3bchncn62sm0000gn/T/jl_6i3bqi/Project.toml`
  [ff71e718] MixedModels v4.34.0

julia> dat = MixedModels.dataset(:insteval)
Arrow.Table with 73421 rows, 7 columns, and schema:
 :s        String
 :d        String
 :dept     String
 :studage  String
 :lectage  String
 :service  String
 :y        Int8

julia> f = @formula(y ~ 1 + service + (1|s) + (1|d) + (1|dept));

julia> m1 = fit(MixedModel, f, dat; progress=false);

julia> @be objective(updateL!($m1))
Benchmark: 20 samples with 1 evaluation
 min    5.143 ms (38 allocs: 1.094 KiB)
 median 5.172 ms (38 allocs: 1.094 KiB)
 mean   5.176 ms (38 allocs: 1.094 KiB)
 max    5.270 ms (38 allocs: 1.094 KiB)

# using the db/RFP branch

julia> m1 = fit(MixedModel, f, dat; progress=false);

julia> BlockDescription(m1)
rows:      s             d            dept         fixed     
2972:   Diagonal   
1128:    Sparse      Diag/TrRFP  
  14:    Dense      Sparse/Dense   Diag/Dense  
   3:    Dense         Dense         Dense         Dense     


julia> @be objective(updateL!($m1))
Benchmark: 4 samples with 1 evaluation
        26.432 ms (31 allocs: 768 bytes)
        26.525 ms (31 allocs: 768 bytes)
        26.669 ms (31 allocs: 768 bytes)
        26.937 ms (31 allocs: 768 bytes)

julia> m2 = LinearMixedModel(f, dat; RFPthreshold=2000);

julia> BlockDescription(m2)
rows:      s             d            dept         fixed     
2972:   Diagonal   
1128:    Sparse      Diag/Dense  
  14:    Dense      Sparse/Dense   Diag/Dense  
   3:    Dense         Dense         Dense         Dense     


julia> @be objective(updateL!($m2))
Benchmark: 20 samples with 1 evaluation
 min    5.144 ms (28 allocs: 656 bytes)
 median 5.173 ms (28 allocs: 656 bytes)
 mean   5.173 ms (28 allocs: 656 bytes)
 max    5.256 ms (28 allocs: 656 bytes)
```

palday:
I agree it's worth trying!

ajinkya-k:
Isn’t there an SYRK for rectangular full packed format?

ajinkya-k:
Why can we not just use function syrk! here: https://github.com/JuliaLinearAlgebra/RectangularFullPacked.jl/blame/b159bfabd981c46b8842ff1ad857aa95f6b52cad/src/hermitian.jl#L43

dmbates:
@ajinkya-k I should have been more clear that I was speaking of the case where L[2,1] is sparse.  Using RFP for L[2,2] only makes sense if L[2,2] is very large and, by design, L[2,1] is even larger (because L[1,1] is as large as possible).  If L[2,1] needs to be stored as a dense matrix you have already got a much bigger problem to contend with.

ajinkya-k:
No I think that was clear but I twisted myself into knots trying to read the file changes from the PR

dmbates:
So I have a question about strategy.  The bottleneck when using TriangularRFP for (usually) the [2,2] block of L which implies that the [2,1] block will be a SparseMatrixCSC is the `rankUpdate!` operation.  This is slow because every assignment like `C[i,j] += A[j, k] * A[i, k]` has to go through a complex calculation of mapping the virtual indices `[i,j]` to the Cartesian indices in `C.data`.

When it is the lower triangle of `C` being stored in `C.data` (and `C.transa` is `false`, which is always the case for us) the first half of the columns of C are either in their original positions in `C.data` or shifted down by one position.

I did a calculation of the number of updates in this `nonflipped` configuration versus the number in the `flipped` configuration for the `insteval` model.  It was about 3/4 nonflipped and 1/4 flipped, which is roughly the proportion you would expect based on the sizes of the chunks in C.
```
(980444, 306131)
```

However, if we reorder the rows of `A` by decreasing row sums
```
julia> MixedModels.countinds(L[p, :])
(1221833, 64742)
```
we get about 95% non-flipped and that certainly seems worth it.  So, at what point should the levels of the second grouping factor be reordered?

I think it will be rare to need to use RFP but, if L[2,1] is a `SparseMatrixCSC` there wouldn't be much harm in reordering the rows (i.e. permute the levels of the second grouping factor).

Is this explanation sufficiently intelligible to be able to offer an opinion?

palday:
@dmbates I think that change makes sense. I also wonder if this would slightly speed up things for the non RFP case --- I've certainly seen this type of optimization have big impacts in other code. 

If I'm thinking about this correctly, ordering by row-sums works out to ordering by something like "strength of crossing", right?

We don't make any guarantees about the internal ordering of levels, so that's not an issue.

dmbates:
For the time being I will do the simplified indexing part and not do the reordering of levels part.  However, we can recommend reordering the levels before creating the model.

src/linearmixedmodel.jl
```diff
@@ -384,34 +398,52 @@ function _pushALblock!(A, L, blk)
...
     end
-    return identity.(A), identity.(L)
+    for k in axes(reterms, 1)    # patch up the UniformBlockDiagonal in L for nested factors
+        diagind = kp1choose2(k)
+        Ldi = L[diagind]
+        if isa(Ldi, UniformBlockDiagonal)
+            L[diagind] = LowerTriangular(Ldi)
+        end
+    end
+    return identity.(A), identity.(L)  # does anyone remember what the `identity` is for?`
```
ararslan, line 446: https://github.com/JuliaStats/MixedModels.jl/pull/772#discussion_r1648333272

dmbates:
I wrote the enclosed to do the reordering of levels in a `CategoricalArray` by incidence.  (I'm not a big fan of `CategoricalArrays.jl` but it seemed the easiest way to accomplish this.)
```julia
using CategoricalArrays, DataFrames, MixedModels

"""
    relevel_by_incidence!(v::CategoricalArray; rev=false)

Return `v` with the levels ordered by increasing (decreasing when `rev=true`) incidence.
"""
function relevel_by_incidence!(v::CategoricalArray; rev=false)
    ra = refarray(droplevels!(v))
    counts = zeros(Int, maximum(ra))
    for i in ra
        counts[i] += 1
    end
    levels!(v, levels(v)[sortperm(counts; rev)])
    return v
end

dat = DataFrame(MixedModels.dataset(:insteval))
dat.d = relevel_by_incidence!(categorical(dat.d); rev=true)
m = LinearMixedModel(@formula(y ~ 1 + service + (1|s) + (1|d) + zerocorr(1+service|dept)), dat)
m.A[3].diag
```

palday (approved):

PR #825, opened 2025-05-13
Cook's Distance for linear mixed model

palday:
Thanks for contributing!

- [x] add entry in NEWS.md


test/pls.jl
```diff
@@ -673,6 +673,39 @@ end
...
+                      1.315739e-05, 0.03073671, 0.00899036, 0.01262709, 0.002494427, 0.03239389, 0.01698841,
+                      0.0002320865, 0.0135889, 0.02761053, 0.02916589, 0.04618232, 0.07875934, 0.02248172,
+                      0.1308213, 0.04340534, 0.05379937, 0.0873526, 0.07648689, 0.03333461, 0.01267992,
+                      0.004915966, 0.0003118122, 0.006997041, 0.01519545, 0.162238, 0.01767151, 0.02365221,
+                      0.05187042, 1.31043e-07, 0.002747362, 0.003266733, 0.005808394, 0.03485179, 0.003650455,
+                      0.0003004733, 1.535027e-05, 0.0168071, 1.510735e-05]
+        model = refit!(first(models(:sleepstudy)); progress=false)
+        @test all(zip(lme4_cooks, cooksdistance(model))) do (x, y)
+            return isapprox(x, y; atol=1e-5)
+        end
```
ararslan, line 707: For my own edification, is there a reason why you want to check component-wise approximate equality here rather than approximate equality of the arrays? I don't have a good intuition for when one approach is preferable to the other.

src/linearmixedmodel.jl
```diff
@@ -414,6 +414,14 @@ function createAL(reterms::Vector{<:AbstractReMat{T}}, Xy::FeMat{T}) where {T}
     return identity.(A), identity.(L)
 end
 
+function StatsAPI.cooksdistance(model::LinearMixedModel)
+    r = residuals(model)
+    _, p, _, _ = size(model)
+    mse = dispersion(model, true)
+    hii = leverage(model)
+    return @. (r / (1 - hii))^2 * hii / (mse * p)
```
ararslan, line 422: An interesting future addition could be conditional Cook's distance as described in this paper: https://www.vvsor.nl/wp-content/uploads/2020/06/KM2000063003.pdf

ararslan (approved):
Cool, I think this makes sense to me. Convenient that the formulation is analogous to that of usual linear models.

test/pls.jl
palday: Component-wise is stronger (which isn't always a good thing). I wanted it here because Cook's distance is used to evaluate individual observations, so we want to make sure it aligns at the level of individual observations.

On the other hand, if you were thinking about whether two points in some higher dimensional space are close, then you would care whether the vectors representing them are approximately equal not whether each coordinate is approximately equal. (Although this works out to the same thing in some norms, but not in L2.)

ararslan: Makes perfect sense. Thanks, I appreciate the explanation!

Issue #826, opened 2025-05-26
refit! on a prfit! model starts at a parameter vector of zeros


dmbates:
I haven't tracked this one down yet but here is a transcript showing the problem (which doesn't occur with NLopt).

```julia
[ Info: using OhMyREPL.jl
[ Info: using AppleAccelerate.jl
[ Info: using Revise.jl
(MixedModels) pkg> activate --temp
  Activating new project at `/var/folders/d9/t94kd5m50rl1r3bchncn62sm0000gn/T/jl_WN1QE7`

(jl_WN1QE7) pkg> add PRIMA, MixedModels;
   Resolving package versions...
    Updating `/private/var/folders/d9/t94kd5m50rl1r3bchncn62sm0000gn/T/jl_WN1QE7/Project.toml`
  [ff71e718] + MixedModels v4.35.0
  [0a7d04aa] + PRIMA v0.2.2
    Updating `/private/var/folders/d9/t94kd5m50rl1r3bchncn62sm0000gn/T/jl_WN1QE7/Manifest.toml`
  [79e6a3ab] + Adapt v4.3.0
  [66dad0bd] + AliasTables v1.1.3
  [4c555306] + ArrayLayouts v1.11.1
  [69666777] + Arrow v2.8.0
  [31f734f8] + ArrowTypes v2.3.0
  [093aae92] + BSplineKit v0.19.0
  [aae01518] + BandedMatrices v1.9.4
  [c3b6d118] + BitIntegers v0.3.5
  [fa961155] + CEnum v0.5.0
  [5ba52731] + CodecLz4 v0.4.6
  [6b39b394] + CodecZstd v0.8.6
  [bbf7d656] + CommonSubexpressions v0.3.1
  [f70d9fcc] + CommonWorldInvalidations v1.0.0
  [34da2185] + Compat v4.16.0
  [f0e56b4a] + ConcurrentUtilities v2.5.0
  [9a962f9c] + DataAPI v1.16.0
  [864edb3b] + DataStructures v0.18.22
  [e2d170a0] + DataValueInterfaces v1.0.0
  [85a47980] + Dictionaries v0.4.5
  [163ba53b] + DiffResults v1.1.0
  [b552c78f] + DiffRules v1.15.1
  [31c24e10] + Distributions v0.25.120
  [ffbed154] + DocStringExtensions v0.9.4
  [4e289a0a] + EnumX v1.0.5
  [e2ba6199] + ExprTools v0.1.10
[...]

PR #827, opened 2025-05-26
[WIP] Trying to solve refit! problems with a prfit! model.

dmbates:
- An attempt to address #826 
- Currently causes a test failure in the `cbpp` tests, which surprises me.


dmbates:
It seems that the refit! method did not reset the value of `optsum.final` which is the cause of the warning from the PRIMA Fortran code about the initial value and the trust region radius.  Perhaps the PRIMA code is setting the parameter vector to the lower bound in that case but that seems unlikely.

PR #829, opened 2025-05-29
Fix markdown show on 1.13

palday:
- [ ] add entry in NEWS.md


src/mimeshow.jl
```diff
@@ -148,8 +148,10 @@ function _markdown(m::MixedModel)
         push!(rows, newrow)
     end
 
-    re_without_fe = setdiff(
-        mapfoldl(x -> Set(getproperty(x, :cnames)), ∪, m.reterms), coefnames(m)
+    re_without_fe = setdiff!(
+        # why not use `Set`? because ordering can change between releases, while
+        # unique preserves the original order
+        unique(mapfoldl(x -> getproperty(x, :cnames), vcat, m.reterms)), coefnames(m)
```
ararslan, line 154: Some quick benchmarking suggests that this is faster with fewer allocations:
```suggestion
        unique!(mapfoldl(x -> getproperty(x, :cnames), append!, m.reterms; init=String[])), coefnames(m)
```
Users could get tens, even hundreds of nanoseconds of their lives back that would otherwise be spent printing to Markdown!

test/pivot.jl
```diff
@@ -3,7 +3,10 @@ using LinearAlgebra, StableRNGs, StatsModels, Test
 import MixedModels: statsrank
 
 xtx(X) = Symmetric(X'X, :U)  # creat the symmetric matrix X'X from X
-LinearAlgebra.rank(F::LinearAlgebra.QRPivoted; tol=1e-8) = searchsortedlast(abs.(diag(F.R)), tol, rev=true);
+# this is defined in Julia 1.13
+@static if VERSION < v"1.13.0-DEV.655"
```
ararslan, line 7: FYI `@static` has no effect at the top level

src/blockdescription.jl
```diff
@@ -55,4 +55,4 @@ function Base.show(io::IO, ::MIME"text/plain", b::BlockDescription)
     end
 end
 
-Base.show(io::IO, b::BlockDescription) = show(io, MIME"text/plain"(), b)
+Base.show(io::IO, b::BlockDescription) = show(io, MIME("text/plain"), b)
```
ararslan, line 58: This is the same, isn't it? Is there a reason to prefer one over the other?

ararslan (approved):
Minor, non-blocking comments/suggestions/questions

test/pivot.jl
palday: yeah, I know, but I just instinctively do it when I'm conditioning on `VERSION`

src/blockdescription.jl
palday: It's exactly the same. I just decided that the former looks really weird and I don't like it....

PR #832, opened 2025-06-16
Add output from insteval model

dmbates:
- This branch is mostly to use github CI to fit the insteval model from https://github.com/JuliaMixedModels/Manuscripts/blockedcholesky on a variety of architectures/operating systems/versions of Julia


palday:
@dmbates should we merge this in or close this PR?

dmbates:
The rationale for including it (these fits being used in the BlockedCholesky paper) no longer holds so I think we can just close it.  In the long run we may wish to reexamine the facility to use `PRIMA`, which requires a Fortran compiler to build `PRIMA_dll`.  I'm not sure the benefit is that great.

Issue #833, opened 2025-06-17
Different results with Julia 1.11 on macOS


andreasnoack:
The following example gives a different result with Julia 1.11.5 compared with Julia 1.10.5
```
using DataFrames, CSV, MixedModels

df = CSV.read("data.csv", DataFrame)

_lmm = LinearMixedModel(
    @formula(endpoint ~ 1 + formulation + sequence + period + (1 | id)),
    df;
    contrasts = Dict(:period => DummyCoding()),
)

_lmm.optsum.optimizer = :LN_COBYLA

fit!(_lmm; REML = true)
```

[data.csv](https://github.com/user-attachments/files/20770820/data.csv)

This doesn't seem to happen on Linux so it might be related to minor rounding differences. If I use BOBYQA then I'm getting the better solution. We are using COBYLA because of https://github.com/JuliaStats/MixedModels.jl/issues/705.

### Julia 1.10.5
```julia
Linear mixed model fit by REML
 endpoint ~ 1 + formulation + sequence + period + (1 | id)
 REML criterion at convergence: 1143.3630490413652

Variance components:
            Column   Variance Std.Dev.
id       (Intercept)  0.842408 0.917828
Residual              2.118190 1.455400
 Number of obs: 298; levels of grouping factors: 77

  Fixed-effects parameters:
───────────────────────────────────────────────────────
                     Coef.  Std. Error      z  Pr(>|z|)
───────────────────────────────────────────────────────
(Intercept)     -0.313961     0.254152  -1.24    0.2167
formulation: T   0.177572     0.168966   1.05    0.2933
sequence: TRTR   0.330212     0.269183   1.23    0.2199
period: 2       -1.39972      0.235472  -5.94    <1e-08
period: 3        0.947798     0.241498   3.92    <1e-04
period: 4       -0.0445583    0.236472  -0.19    0.8505
───────────────────────────────────────────────────────

julia> _lmm.optsum
...
Function evaluations:     28
xtol_zero_abs:            0.001
ftol_zero_abs:            1.0e-5
Final parameter vector:   [0.6306362152099609]
Final objective value:    1143.3630490413652
Return code:              FTOL_REACHED

julia> _lmm.optsum.fitlog
28-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.0], 1153.9741825642604)
 ([1.75], 1197.7827103789102)
 ([0.2500000000000001], 1159.431337870577)
 ([0.625], 1143.3663690793285)
[...]

PR #840, opened 2025-07-19
[WIP] Switch to unconstrained optimizers - step 1

dmbates:
- An attempt to make it easier to switch to unconstrained optimizers by first changing all lower bounds to `-Inf`
- There is a separate `rectify!` step after the optimization to flip the signs on columns of \lambda where the diagonal element converges to a negative value. 
- One of the tests for GLMMs is still giving problems with negative diagonal elements on \lambda.  Not sure why
- ~`test/prima.jl` is giving errors so I disabled it.  I'm too tired to work out what is happening there.~


dmbates:
I'll look into the failures on evaluation of the logdet in the model for the Dyestuff2 data tomorrow.  It should be a straightforward evaluation because the model (should) converge to \theta = 0.

andreasnoack:
I would expect item two to be the more fruitful one, i.e. to focus on differences related to evaluating just the objective function across microarchitectures and leave out the effect of the optimization algorithm. I think it would be interesting to see differences in the initial values and in a value where the two start to differ. It doesn't matter which of the microarchitectures created the value, but the same should be used when evaluating both. Again, to leave out the effect of the optimization. I'm speculating if the objective function is ill-conditioned for some input parameters. (Here, I mean ill-conditioned in the general sense that measures an arbitrary function and not just the linear solve conditioning.)

I think it would be beneficial to stick with BOBYQA for now. It means one moving part less, and I don't suspect that disabling the bounds in BOBYQA by setting the lower bound to `-Inf` should be the issue.

dmbates:
@andreasnoack I am okay with using LN_BOBYQA as the default for a while longer but definitely feel we should eliminate the lowerbd stuff if we are not going to use it.  It may be a result of getting older and less mentally flexible but I am finding the current code to be overly complex in places.  Like any large project there is a tendency to add code to patch special cases and justify this by adding more tests.  Eventually a person looking at the code for the first time is left to wonder "Why is there all this code about a lower bound when the lower bound is always set to -Inf?".  If you don't know the history it is nonsense.

dmbates:
On an x86_64 machine I get, with OpenBLAS
```julia
dmbates@plinth:~/.julia/dev/MixedModels$ julia -t8 --sta=no --quiet --proj
julia> using MixedModels

julia> include("./test/modelcache.jl")
models (generic function with 1 method)

julia> m = refit!(only(models(:penicillin)); fitlog=true);

julia> first(m.optsum.fitlog, 8)
8-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.0, 1.0], 364.6267798165887)
 ([2.0, 1.0], 369.17598910041096)
 ([1.0, 2.0], 341.1011743244961)
 ([0.0, 1.0], 441.9062007516885)
 ([1.0, 0.0], 634.3549438457203)
 ([1.444405652744627, 1.5955538627700872], 340.7336990803923)
 ([0.4942375301999674, 1.9072917958163513], 376.64342079733973)
 ([1.8130833383189084, 1.5302193144504623], 343.97029543404193)

julia> last(m.optsum.fitlog, 4)
4-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.5368792590556977, 3.2176454905841227], 332.1883597436415)
 ([1.5375182176811606, 3.219536746361558], 332.18834878933217)
 ([1.537643747886617, 3.2205288361596702], 332.18834933611157)
 ([1.5375133789110618, 3.219886141303852], 332.18834878807843)
```
and with MKL
```julia
dmbates@plinth:~/.julia/dev/MixedModels$ julia -t8 --quiet --proj
julia> using MixedModels

julia> include("./test/modelcache.jl")
models (generic function with 1 method)

julia> m = refit!(only(models(:penicillin)); fitlog=true);

julia> first(m.optsum.fitlog, 8)
8-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.0, 1.0], 364.62677981659255)
 ([2.0, 1.0], 369.1759891004115)
 ([1.0, 2.0], 341.10117432449795)
 ([0.0, 1.0], 441.90620075169636)
 ([1.0, 0.0], 634.3549438457203)
 ([1.4444056527446676, 1.5955538627700978], 340.73369908034823)
 ([0.4942375301998896, 1.9072917958160005], 376.64342079740277)
 ([1.8130833383190654, 1.530219314450679], 343.97029543397673)

julia> last(m.optsum.fitlog, 4)
4-element Vector{Tuple{Vector{Float64}, Float64}}:
 ([1.5375663667436432, 3.2196670002814987], 332.1883486857041)
 ([1.537451252957679, 3.21988892102474], 332.18834899335576)
 ([1.5375634872760895, 3.2196078285002994], 332.18834870172884)
 ([1.5375910290916945, 3.2196608178035246], 332.18834867791)
```

dmbates:
When using MKL on x86_64 the last element of `last(m.L)` changes, hence the penalized sum of squared residuals changes
```julia
julia> abs2(last(last(m.L)))
158.6355555555715
```

We can reproduce the calculation using `BigFloat`.  In this case the results are simplified because the cholesky factor in `m.L[3]` is exactly `5.0 * I(6)`.  
```julia
julia> m.L[5] == m.A[5] ./ 5.0
true
```

So the downdate of `copyto!(last(m.L), last(m.A))` followed by the Cholesky factorization in `BigFloat` can be written
```julia
julia> BigFloat.(last(m.A))
2×2 Matrix{BigFloat}:
  144.0   3308.0
 3308.0  76582.0

julia> m.L[5]
2×6 Matrix{Float64}:
   4.8    4.8    4.8    4.8    4.8   4.8
 120.8  105.4  119.6  109.8  110.2  95.8

julia> L5 = BigFloat.(m.A[5]) ./ 5
2×6 Matrix{BigFloat}:
   4.8    4.8    4.8    4.8    4.8   4.8
 120.8  105.4  119.6  109.8  110.2  95.8

julia> first(L5)
4.800000000000000000000000000000000000000000000000000000000000000000000000000028

julia> L6 = cholesky(Symmetric(BigFloat.(last(m.A)) - L5 * L5', :L))
Cholesky{BigFloat, Matrix{BigFloat}}
L factor:
2×2 LowerTriangular{BigFloat, Matrix{BigFloat}}:
  2.4       ⋅ 
 55.1333  12.5951

julia> last(L6.L)
12.59506076029629781458345910991177392468429196879571875004312834742566832337298

julia> last(last(m.L))
12.595060760296931
```

When using OpenBLAS on x86_64 the results are
```julia
julia> m = refit!(only(models(:penicillin)); fitlog=true);

julia> objective!(m, [0.0, 1.0])
441.9062007516885

julia> last(last(m.L))
12.595060760296588
```
so the OpenBLAS result is slighly closer to the BigFloat value.

Frankly I am a bit surprised that there is a difference in the Cholesky factor of a 2 by 2 symmetric matrix.

andreasnoack:
I don't think the difference in the Cholesky matters too much here. For the same input, I get very similar results for OpenBLAS and MKL. With
```julia
julia> repr(m.A[6] - m.L[5]*m.L[5]')
"[5.7599999999999625 132.31999999999925; 132.31999999999925 3198.3199999999924]"
```
I'm getting
#### OpenBLAS
```julia
julia> cholesky([5.7599999999999625 132.31999999999925; 132.31999999999925 3198.3199999999924]).L[2,2]
12.595060760296588
```

#### MKL
```julia
julia> cholesky([5.7599999999999625 132.31999999999925; 132.31999999999925 3198.3199999999924]).L[2,2]
12.59506076029659
```

which are only one ULP apart. So I think the difference between the two Choleskys is because of the inputs. With OpenBLAS, there is one ULP error in `L[5]`
```julia
julia> print(m.L[5])
[4.800000000000001 4.800000000000001 4.800000000000001 4.800000000000001 4.800000000000001 4.800000000000001; 120.80000000000001 105.4 119.60000000000001 109.80000000000001 110.2 95.80000000000001]
```
so I have
```julia
julia> m.A[5]/5 == m.L[5]
false
```

but `true` with MKL. I think this difference could explain the difference in the Cholesky result.

Just speculating, but maybe the downdate `m.A[6] - m.L[5]*m.L[5]'` could introduce a large cancellation error for some values of `θ`?

dmbates:
That's interesting.  The calculation of `m.L[5]` is performed as an `rdiv!` with `m.L[3]` as `LowerTriangular` so it may be evaluated differently.  It is just a coincidence for this particular model and parameter vector that `m.L[3]` is `5 * I(6)`.

dmbates:
I would like to follow up on this thread but I am having difficulty thinking of what to do next.  Suggestions welcome.

dmbates:
I think I can go ahead and remove the `lowerbd` property from `optsum`.  @palday Do you think I should do that in a separate PR to make it easier to back out if necessary?

palday:
> I'm not sure what is happening with the formatter test. @palday Do you have any ideas?

I pushed a commit with formatting. The formatter only comments if the PR isn't draft -- that way it doesn't create lots of suggestions when you're still pushing draft code. 😄

dmbates:
@palday Thanks for the formatter correction.

palday:
> I am beginning to see the problem with the style enforcer. For some reason JuliaFormatter on my computer changes successive ':' arguments by removing spaces around them (i.e. `foo(k, :, :)` to `foo(k,:,:)`) whereas the version on github wants them back the other way. I'll just forgo running `format` on the files before committing.

I think part of the problem is the difference between v1 and v2 of JuliaFormatter -- I had similar problems at some point when I was running a local version different from the version in the action.

palday:
> In addition to removing the lowerbd field of optsum I was thinking of changing the structure of fitlog to be a matrix table (in the sense of `Tables.table`). The first change would not need to be considered breaking after this but the second would. Should I try to incorporate all of these in "one big beautiful PR" or go piecemeal?

I'm onboard with these changes, but please go piecemeal. 😄

src/profile/thetapr.jl
```diff
@@ -28,12 +28,13 @@ function profileθj!(
...
     optsum = m.optsum
-    (; final, fmin, lowerbd) = optsum
+    (; final, fmin) = optsum
     j = parsej(sym)
     θ = copy(final)
-    lbj = lowerbd[j]
     osj = optsum
     opt = Opt(osj)
+    pmj = m.parmap[j]
+    lbj = pmj[2] == pmj[3] ? zero(T) : T(-Inf)
```
palday, line 37: this pattern has come up enough, I wonder if it makes sense to define a helper function `_lowerbound(parmapj)`

[noblock]

palday (approved):

andreasnoack:
Sorry for the silence here. I was busy with vacation. I just tried this out with the test cases that we have and this branch seems to avoid any of the issues that we'd hit in the past, i.e. both #705 and #833. No regressions appeared either so I'm in favor of moving forward.

Regarding https://github.com/JuliaStats/MixedModels.jl/pull/840#issuecomment-3099236809 then I don't think it has any immediate implications for the transition to unconstrained optimization. However, it might be worth analyzing if the rank downgrade introduces numerical issues under some conditions and if they can be avoided. Such issues would be unrelated to the optimization approach, but they might explain the required changes to the test values here.

dmbates:
I added a `timingtable` function definition to `test/modelcache.jl` but without adding any calls to it.  If you call it separately for the `:insteval` models it produces
```julia
julia> timingtable()
Table with 6 columns and 16 rows:
      modnum  bkend  optmzr     feval  objtiv     time
    ┌─────────────────────────────────────────────────────
 1  │ 1       nlopt  LN_NEWUOA  81     2.37722e5  0.424785
 2  │ 1       nlopt  LN_BOBYQA  107    2.37722e5  0.560039
 3  │ 1       prima  newuoa     92     2.37722e5  0.483105
 4  │ 1       prima  bobyqa     104    2.37722e5  0.546217
 5  │ 2       nlopt  LN_NEWUOA  50     2.37586e5  0.249036
 6  │ 2       nlopt  LN_BOBYQA  46     2.37586e5  0.230194
 7  │ 2       prima  newuoa     47     2.37586e5  0.235359
 8  │ 2       prima  bobyqa     46     2.37586e5  0.230809
 9  │ 3       nlopt  LN_NEWUOA  196    2.37649e5  1.13438
 10 │ 3       nlopt  LN_BOBYQA  170    2.37649e5  0.985796
 11 │ 3       prima  newuoa     203    2.37649e5  1.1752
 12 │ 3       prima  bobyqa     173    2.37649e5  1.00095
 13 │ 4       nlopt  LN_NEWUOA  181    2.37647e5  1.05708
 14 │ 4       nlopt  LN_BOBYQA  490    2.37647e5  2.85961
 15 │ 4       prima  newuoa     211    2.37647e5  1.23718
 16 │ 4       prima  bobyqa     293    2.37647e5  1.71672
```

allowing us to see which are the faster and slower (mostly determined by the number of function evaluations but not entirely - `:prima` seems to have more overhead than does `:nlopt`) optimizers.  I have been thinking of splitting the objective into two columns, the minimum observed objective for each model, which will be the same for all optimizers, and the deviation from the minimum.  That way we can see which optimizer gives the best value.

Shall I add a column for the number of parameters in the model?

palday:
> I have been thinking of splitting the objective into two columns, the minimum observed objective for each model, which will be the same for all optimizers, and the deviation from the minimum. That way we can see which optimizer gives the best value.
> 

How about each optimizer's final objective value and that value's deviation from the best value across optimizers? 

> Shall I add a column for the number of parameters in the model?

Yes.

palday:
> Finally this version is green on the tests except for the nightly version of Julia and that failure is due to compilation failures in other packages. I will hold off on declaring this to be v5.0.0 until further checking and enhancements take place. I think we should remove the `lowerbd` property in `Optsum`, but I am willing to be persuaded otherwise (it's easier to keep a redundant property in the struct than to remove it and find we need to replace it later).

Yep, you can do the `lowerbd` purging in a follow-up PR, but let's merge this one! :rocket:

palday:
@dmbates There are a few small things that I want to do in a follow-up PR before making the breaking release, but I've added an appropriate NEWS entry and marked the version as `5.0.0-DEV`. I'm going to go ahead and merge this PR. :sunglasses:

PR #843, opened 2025-08-03
stopgap fix for Aqua.jl on Julia 1.13 nightly

palday:
until https://github.com/JuliaTesting/Aqua.jl/issues/343 is addressed


dmbates (approved):
Thanks for looking into this.

lgoettgens:
just FYI: https://github.com/JuliaTesting/Aqua.jl/issues/343 is fixed in Aqua v0.8.14 (released just a few minutes ago)

PR #848, opened 2025-08-21
Deprecations, documentation and clean up in preparation for 5.0

### LINKED ISSUE #567: Consistency in progress options
For historical reasons, the bootstrap and `replicate` use `hide_progress=false` as the default, while the fitting functions use `progress=true`. We should change everything to be `progress` to **show** the progress.

### LINKED ISSUE #599: Use RegressionFormulae.jl for defining nesting syntax in fixed effects
And just generally make sure we'll all playing nice together.

### LINKED ISSUE #618: Simplify fitlog logic
If we want to have fitlog contain the initial and final parameter and objective values and have the `.initial`, `.final` fields just point to those values we should expand the schema for fitlog to include the evaluation number (and increment it at the beginning of the `obj` function, not in the middle as is done now).  Either that or abandon the idea of using `thin` and just incorporate a Boolean `logevaluations` or some name like that.  I doubt very much that many users will choose values of `thin` other than the default, `typemax(Int)` or `1`. 

_Originally posted by @dmbates in https://github.com/JuliaStats/MixedModels.jl/pull/615#issuecomment-1133067632_

### LINKED ISSUE #675: Drop multithreading in bootstrap
With improvements/changes in BLAS multithreading, Julia-level threads are competing with the BLAS threads and we don't see any improvement in most situations with `use_threads=true`. In fact, we sometimes see performance degradation. So let's drop multithreading for the bootstrap and simplify that code. We should instead develop an example for the docs showing how to optionally parallelize using Distributed for people with access to clusters / large computer.

- https://github.com/JuliaStats/MixedModels.jl/issues/640
- https://github.com/JuliaStats/MixedModels.jl/issues/674

NB: This doesn't prevent us from adding threading to various custom linear algebra methods, if those prove worthwhile. That said, I think that should be an internal performance optimization and not expose any changes in the external API.

palday:
# Release TODO
- [x] add entry in NEWS.md

# Changes
- [x] deprecated `use_threads` has been removed (closes #675)
- [x] deprecated `hide_progress` has been removed (closes #567)
- [x] `Optsummary.lowerbd` has been removed (#849)
- [x] `restoreoptsum!` tests have been ... restored 
- [x] `fitlog` logic has been simplified (always keep, `final` is just final fitlog value) (#850)
- [x] nesting syntax in fixed effects is removed (#855)
- [x] README has been updated
- [x] Section on optimization in docs has been updated #853
- [x] Choice of backend and optimizer is abstracted away (allowing us to change the default backend more easily) #853
- [x] dataset and datasets has been removed (#854)
- [x] one-argument GLMM predict supports the `type` kwarg. (#856)
- [x] `likelihoodratiotest` has been consolidated with `lrtest` (#858)


NEWS.md
```diff
@@ -2,7 +2,9 @@ MixedModels v5.0.0 Release Notes
 ==============================
 - Optimization is now performed _without constraints_. In a post-fitting step, the Cholesky factor is canonicalized to have non-negative diagonal elements. [#840]
 - The default optimizer has changed to NLopt's implementation of NEWUOA where possible. NLopt's implementation fails on 1-dimensional problems, so in the case of a single, scalar random effect, BOBYQA is used instead. In the future, the default optimizer backend will likely change to PRIMA and NLopt support will be moved to an extension. Blocking this change in backend is an issue with PRIMA.jl when running in VSCode's built-in REPL on Linux. [#840]
-- [BREAKING] Support for constrained optimization has been completely removed, i.e. the field `lowerbd` has been removed from `OptSummary`. [#849]
+- [BREAKING] Support for constrained optimization has been completely removed, i.e. the field `lowerbd` has been removed from `OptSummary`.[#849]
```
dmbates, line 5: It looks as if the only change here is removal of a space between the period and the opening bracket of `[#849]`  Other entries retain that space.  Was this change unintentional?

dmbates (approved):

PR #850, opened 2025-08-21
replace fitlog with a column table, always populate it

### LINKED ISSUE #618: Simplify fitlog logic

palday:
- [x] add entry in NEWS.md

closes #618


dmbates (approved):
As I said in the review of #853 I think my plan of creating a `MatrixTable` for the `fitlog` led to too many changes so I think that this PR and #853 should both be merged.  Afterwards I will add an extractor function to create a table with scalar columns from the current tabular form, in case that is desired.

palday:
> Afterwards I will add an extractor function to create a table with scalar columns from the current tabular form, in case that is desired.

@dmbates don't forget that we have a `Tables.columntable` method that does something similar: 

https://github.com/JuliaStats/MixedModels.jl/blob/bf4b8ebfec4964059ffd39c30b25264730d17c85/src/optsummary.jl#L108-L134

PR #860, opened 2025-08-27
fix for Grouping() with latest CategoricalArrays

palday:
- [x] add entry in NEWS.md


dmbates (approved):
Always happens - as soon as you release a version a new change is required.

PR #865, opened 2025-09-04
display fix for unlikely but possible type inference failure in bootstrap table

palday:
- [x] add entry in NEWS.md

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -359,7 +359,10 @@ function StatsBase.confint(
     par = filter(collect(propertynames(tbl))) do k
         k = string(k)
         # σ is missing in models without a dispersion parameter
-        if k == "σ" && Missing <: eltype(tbl.σ)
+        Tσ = eltype(tbl.σ)
+        # see https://github.com/JuliaStats/MixedModels.jl/pull/861#discussion_r2323094090
+        # for more info on why this logic is so convulated
+        if k == "σ" &&  ((Tσ === Any && any(ismissing, tbl.σ)) ||  Missing <: Tσ)
             return false
         end
         return !startswith(k, 'θ') && k != "obj"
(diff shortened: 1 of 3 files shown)


src/bootstrap.jl
```diff
@@ -359,7 +359,10 @@ function StatsBase.confint(
     par = filter(collect(propertynames(tbl))) do k
         k = string(k)
         # σ is missing in models without a dispersion parameter
-        if k == "σ" && Missing <: eltype(tbl.σ)
+        Tσ = eltype(tbl.σ)
+        # see https://github.com/JuliaStats/MixedModels.jl/pull/861#discussion_r2323094090
+        # for more info on why this logic is so convulated
```
ararslan, line 364:
```suggestion
        # If inference failed when constructing `tbl.σ`, we can't rely on the element
        # type to know whether missing values may be present
```
Avoids needing to open a browser if you're reading the code locally.

src/bootstrap.jl
```diff
@@ -359,7 +359,10 @@ function StatsBase.confint(
...
+        if k == "σ" && ((Tσ === Any && any(ismissing, tbl.σ)) || Missing <: Tσ)
```
ararslan, line 365: If I'm not mistaken, this assumes that `Tσ` will only ever be `Any` if there are missing values present. If `Tσ === Any && !any(ismissing, tbl.σ)` then `Missing <: Tσ` because `Missing <: Any`, but you don't want to return `false` in that case because there are no missing values, which means that the model has a dispersion parameter. Right?

ararslan: If that's right, this should be
```suggestion
        if k == "σ" && ((Tσ === Any && any(ismissing, tbl.σ)) || (Tσ !== Any && Missing <: Tσ))
```

ararslan (approved):

PR #878, opened 2026-02-18
Integrate CliqueTrees.jl

samuelsonric:
This PR adds [CliqueTrees.jl](https://github.com/AlgebraicJulia/CliqueTrees.jl) as a backend for computing Cholesky factorizations. It addresses this issue.

> https://github.com/AlgebraicJulia/CliqueTrees.jl/issues/14

Here is how to use the new functionality.

```julia
using LinearAlgebra
using CliqueTrees
using MixedModels                                                                                       
using MixedModels: dataset, updateL!, _init, _transform!, _logdet                                               

import Metis

M = fit(MixedModel,                                                                                      
    @formula(y ~ 1 + service + (1 | s) + (1 | d)),
    MixedModels.dataset(:insteval);
    progress=false)

updateL!(M)
logdet1 = logdet(M)

F, W = _init(M; alg=METIS())
cholesky!(_transform!(M, F, W))
logdet2 = _logdet(M, F)
```

The generic solver in CliqueTrees.jl struggles a bit with the unusual problem topology, and it is slower than the custom solver in MixedModels.jl. I am going to try addressing this upstream.

```julia-repl
julia> @btime updateL!(M);
  5.006 ms (10 allocations: 240 bytes)

julia> @btime cholesky!(_transform!(M, F, W));
  21.974 ms (38 allocations: 15.82 MiB)
```

PR #880, opened 2026-03-03
Generalize sparseL as sparsemat

### LINKED ISSUE #879: Replace _coord for generating sparseL with SparseArrays.findnz
@palday In looking at the code for `sparseL`, motivated by my comment on a `sparseA` extractor for #878, I noticed that our `_coord` utility to return coordinate representation, (I, J, V), of a block of L duplicates the effect of  `SparseArrays.findnz`.  It seems desirable to use `findnz` instead of defining our own extractor and to add methods for the `BlockedSparseMatrix` and `UniformBlockedDiagonal` types.  However, defining a method for the `LinearAlgebra.Diagonal` type would be type piracy.  Is that a big deal and, if so, how would one avoid type piracy?

dmbates:
- Generalize `sparseL` as `sparsemat` so that a companion function `sparseA` can be constructed.
- Replace `_coord` utility for `sparseL` with `_findnz` which, in most methods, falls through to `SparseArrays.findnz`.  Closes #879
- Create and export a `sparseA` extractor as discussed in #878 

- [x] add entry in NEWS.md


dmbates:
I seem to be in a race condition with the Style-Enforcer.  If I run `JuliaFormatter.format` on the `src` directory it removes white-space to create expressions like `view(A,:,j)` then Style-Enforcer wants to put the white-space back to create `view(A, :, j)`.  It doesn't really matter to me which style is used but it would be helpful if I could run a version of `format` that would produce the style that Style-Enforcer wants.  @palday Is there something I should change on my local setup?

src/linearmixedmodel.jl
```diff
@@ -1133,82 +1134,101 @@ end
...
+ The matrix that is returned is lower-triangular but has not been wrapped in `LowerTriangular`.
+"""
+function sparseL(m::LinearMixedModel; fname::Symbol=first(fnames(m)), full::Bool=false)
+    return dropzeros!(sparsemat(:L, m; fname, full))
+end
+
+"""
+    sparseA(m::LinearMixedModel; fname::Symbol=first(fnames(m)), full::Bool=false)
+
+Same as `sparseL` but returning the sparse lower triangle of the symmetric `A` matrix.
```
palday:
```suggestion
Same as [`sparseL`](@ref) but returning the sparse lower triangle of the symmetric `A` matrix.
```

palday (approved):

Issue #883, opened 2026-04-20
_isnested for two model matrices should first check if x, the matrix to be checked if it is nested in y, has zero columns


dmbates:
This is a reminder to myself to make a PR to check if the matrix to be tested to see if it is nested in another has zero columns.  I was writing about a hypothesis test of $H_0: \beta = 1500$ versus $H_a: \beta\neq 1500$ and fit a model with no fixed-effects to be $H_0$.  The `lrtest` method failed when checking if the columns of the model matrix for the null hypothesis were nested in those of the alternative hypothesis.

PR #886, opened 2026-04-21
isnested of a model w/o fixed effects always true

dmbates:
As described in #883, an error was thrown from `isnested(m2, m1)` if `m2` had no fixed-effects parameters - now returns `true`.

- [x] add entry in NEWS.md


dmbates:
Not sure if this warrants a patch version bump or minor version bump.  I would say a patch version but I defer to @palday on this.

palday (approved):

PR #888, opened 2026-04-23
Revisit db/issue833cleanup

dmbates:
I think this is still relevant.  These changes only affect an example in the issues directory so I think they can be considered a patch release.

- [ ] add entry in NEWS.md


palday (approved):
There's no change to the actual package code, so no need to do a release. I agree though that's useful to keep this bit of history around and updated.

PR #890, opened 2026-05-15
docs formatting updates

palday:
- tweaked a few things on the API page
- swapped from Gadfly to CairoMakie


dmbates (commented):
Good idea to replace Gadfly with CairoMakie and AlgebraOfGraphics

dmbates (approved):

PR #911, opened 2026-08-10
make `dataset` officially public

palday:
closes #909


dmbates (approved):
Thanks for doing this.

